PR #127, opened 2014-11-06
Updated the grunt target to delete node_modules directory afterward

DamianEdwards:
- It shells out to "rmdir" tool as that is safe for long paths
- Doing this to work around the issue with KRE compile failing when long paths are present in the project folder


davidfowl:
Whats the deal with this?

Issue #37, opened 2014-03-06
Build fails if there's no src folder


davidfowl:
We should check for existence before

dougbu:
Fixed via f38d08a408dd113a9b21823927603defab40497d

PR #133, opened 2014-12-05
#37, Support sample- or test-only repositories

dougbu:
- skip a few targets when src directory does not exist

/cc Universe triage team


dougbu:
@davidfowl ping

Eilon:
Yeah it seems to match the other existing conditions in the file. :shipit:

PR #148, opened 2015-01-21
Make build install go directly to packages cache

davidfowl:
/cc @pranavkm @dougbu

### DIFF
--- a/build/_k-local-publish.shade
+++ b/build/_k-local-publish.shade
@@ -1,0 +1,16 @@
+@{/*
+
+k-local-publish
+    Builds project. Downloads and executes k sdk tools.
+
+sourcePackagesDir=''
+    Required. Path to packages to install (skips symbol packages)
+*/}
+
+@{
+   var packages = Directory.EnumerateFiles(sourcePackagesDir, "*.nupkg")
+                           .Where(p => !p.EndsWith(".symbols.nupkg"));
+}
+
+exec program='cmd' commandline='/C kpm packages add ${package}' if='!IsMono' each='var package in packages'
+exec program='kpm' commandline='packages add ' if='IsMono' each='var package in packages'
--- a/build/_k-local-publish.shade
+++ b/build/_k-local-publish.shade
@@ -13,4 +13,4 @@ sourcePackagesDir=''
 }
 
 exec program='cmd' commandline='/C kpm packages add ${package}' if='!IsMono' each='var package in packages'
-exec program='kpm' commandline='packages add ' if='IsMono' each='var package in packages'
+exec program='kpm' commandline='packages add ${package}' if='IsMono' each='var package in packages'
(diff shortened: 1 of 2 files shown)


pranavkm:
The only concern is it's a bit easy to wipe out all your locally built packages now - all you do is wipe our your .nuget folder. You'd have to hunt these packages down in the future when you want to switch back to packages from the feed.

dougbu:
think I have a similar concern to @pranavkm

`kpm packages add` determines the target directory from the project.json file, with default $env:USERPROFILE\.dotnet\packages.  so this update mixes what I've downloaded (potentially just for a single project) with my locally-built overrides.

scenario 1: blow away everything I've downloaded _but_ keep what I've installed locally.  I can do that today because the local installations are kept separate (just have exploded copies in $env:USERPROFILE\.dotnet\packages).  ugly after this change.

scenario 2: temporarily build using only published packages.  today I would I disable my NuGet cache source for that build.  only option is to manually remove (or temporarily move) all "*-t00000000000.nupkg" packages after this change.

suggest we use the second, optional parameter to `kpm packages add`, providing the target directory.  or provide an `kpm packages push-local` command.  in either the .shade file or the `push-local` command, use a well-known default location such as $env:USERPROFILE\.dotnet\local that we can either add to our NuGet.config files or manage on a per-developer basis.

side notes  :frowning: -
- `kpm help packages add` fails
- `kpm packages add -?` oddly describes the target packages directory as the `[source]`

davidfowl:
Few comments:

> kpm packages add determines the target directory from the project.json file, with default $env:USERPROFILE\.dotnet\packages. so this update mixes what I've downloaded (potentially just for a single project) with my locally-built overrides.

It doesn't mix anything. Think of it as a way to populate the local packages folder to make your package "visible". It skips the build install step which I think is pointless since you don't have to kpm restore on the other projects to actually consume this package.

> scenario 1: blow away everything I've downloaded but keep what I've installed locally. I can do that today because the local installations are kept separate (just have exploded copies in $env:USERPROFILE\.dotnet\packages). ugly after this change.

This isn't true at all. When you build install and kpm restore, you're in the same boat so I'm not quite sure what you mean. Are you saying that its easy to remove locally built packages before they are installed? Sure that's true, except as soon as you do kpm restore you're hosed in the same way.

>  scenario 2: temporarily build using only published packages. today I would I disable my NuGet cache source for that build. only option is to manually remove (or temporarily move) all "*-t00000000000.nupkg" packages after this change.

Same as before, if you do a single kpm restore at any point ever, you'll have to clean up the local cache anyways so I don't quite understand the sentiment.

> suggest we use the second, optional parameter to kpm packages add, providing the target directory. or provide an kpm packages push-local command. in either the .shade file or the push-local command, use a well-known default location such as $env:USERPROFILE\.dotnet\local that we can either add to our NuGet.config files or manage on a per-developer basis.

So the general fear seems to be around cleaning up these things when you're ready to not use them anymore (a problem that exists today) and one that I agree sucks. Let's go with your suggestion but I'd rather not have to run kpm restore in order to make these packages visible to the runtime. So we also need a clean way to add extra packages locations (or have a default one) that is separate from `%userprofile%\.dotnet\packages`. Unfortunately, NuGet.config doesn't work for this.

pranavkm:
Spoke to @davidfowl about this offline. We can evolve kpm to take away the pains of cleaning this up.

:shipit:  from me

pranavkm:
This was addressed as part of https://github.com/aspnet/Universe/pull/152

PR #150, opened 2015-01-22
Remove System.Data packages

bricelam:


Eilon:
Retroactive :shipit: on this too of course.

PR #154, opened 2015-01-26
Fix for test hang issue

suhasj:
@pranavkm 
Running `k test` on mono does not return the control to the shell due to an issue with xunit. Passing the `-parallel none` flag seems to fix the issue.Adding this to Korebuild so that we can have musicstore passing on Linux and Mac on CI

### DIFF
--- a/build/_k-test.shade
+++ b/build/_k-test.shade
@@ -40,7 +40,11 @@ projectFile=''
         
         foreach (var framework in targetFrameworks)
         {
-            if (framework.StartsWith("net", StringComparison.OrdinalIgnoreCase)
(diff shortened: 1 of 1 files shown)


Praburaj:
Change looks good to me :shipit:

pranavkm:
:shipit:

Eilon:
:shipit: and please log a Universe bug to re-enable parallel tests in the future too.

suhasj:
@Eilon  sure
thanks everybody for the review

victorhurdugaci:
I'm going to close this since it is so old. Please reopen if you still think it is necessary

PR #156, opened 2015-01-30
Quiet builds a bit more

dougbu:
As the commits say:
- include `bower` and `npm` commands in `--quiet` target
- short-circuit `TestCommand()` calls once commands are installed locally

/cc @davidfowl


Eilon:
Eh, seems :shipit: to me... assuming you've tested it in the various relevant scenarios (installed, not installed, quiet, not quiet, etc.).

PR #157, opened 2015-01-30
Change native compilation to use VC++ 2015 instead of VC++ 2013.

GrabYourPitchforks:


GrabYourPitchforks:
For you @pranavkm!

pranavkm:
Would this work for the native code in XRE? cc @davidfowl

pranavkm:
cc @ChengTian

ChengTian:
@pranavkm , I think you mentioned that we will have this in Beta3 release. Why is this PR targeting `dev` branch?

pranavkm:
Sorry, this was for the rc branch. My bad.

ChengTian:
XRE isn't using the standard goal to compile native code. XRE/makfile.shade has hard-coded path to MSBuild 14: https://github.com/aspnet/XRE/blob/dev/makefile.shade#L52

davidfowl:
We use both now

Issue #144, opened 2015-01-10
restore-bower-components target flaky


Praburaj:
We have started seeing failures in `MusicStore.SPA` project on CI where bower installation fails. For reference I'm copying the bower install logs in success vs failure case. I’m able to see a difference while the bower fails to install. On a working bower installation the command executed is `cmd` with `npm install`. Whereas on failing cases the program executed is `node` throwing this error. 

Working bower install:
```
[09:29:09][restore-bower-components] 'bower' is not recognized as an internal or external command,
[09:29:09][restore-bower-components]    command not found (non-0 exit code)
[09:29:09][restore-bower-components] operable program or batch file.
[09:29:09][restore-bower-components] Testing for command: npm --version
[09:29:10][restore-bower-components] 1.4.14
[09:29:10][restore-bower-components]    command found (0 exit code)
[09:29:10][restore-bower-components] Exec
[09:29:10][restore-bower-components]   program: cmd
[09:29:10][restore-bower-components]   commandline: /C npm install --prefix C:\BuildAgent\work\6feb55fcd711693a\bin\nodejs bower
[09:29:10][restore-bower-components]   workingdir: C:\BuildAgent\work\6feb55fcd711693a
[09:29:30][restore-bower-components] bower@1.3.12 bin\nodejs\node_modules\bower
```

Failing bower install:
```
[13:57:44][restore-bower-components] 'bower' is not recognized as an internal or external command,
[13:57:44][restore-bower-components] operable program or batch file.
[13:57:44][restore-bower-components]    command not found (non-0 exit code)
[13:57:44][restore-bower-components] Testing for command: node --version
[13:57:44][restore-bower-components] v0.10.29
[13:57:44][restore-bower-components]    command found (0 exit code)
[13:57:44][restore-bower-components] Exec
[13:57:44][restore-bower-components]   program: node
[13:57:44][restore-bower-components]   commandline: C:\BuildAgent\work\6feb55fcd711693a\bin\nodejs\node_modules\bower\bin\bower install
[13:57:44][restore-bower-components]   workingdir: C:\BuildAgent\work\6feb55fcd711693a\src\MusicStore.Spa
[13:57:44][restore-bower-components] 
[13:57:44][restore-bower-components] module.js:340
[13:57:44][restore-bower-components]     throw err;
[13:57:44][restore-bower-components]           ^
[...]

PR #158, opened 2015-01-31
Create target directory for `npm` installations

dougbu:
- #144
- problems seem primarily to occur when `node` is globally installed but a
  package (`grunt` or `bower`) is not; `npm install X` fails in this case
- create the `--prefix` directory first to avoid these installation failures

### DIFF
--- a/build/_node-install.shade
+++ b/build/_node-install.shade
@@ -10,16 +10,16 @@ default nodeVer = '0.10.28'
 default npmVer = '1.4.9'
 default nodeExeSha = '628FFD6C3577068C00CEC9F6F897F0EC8F5212D9'
 default nodeInstallDir = '${Path.Combine(binDir, "nodejs")}'
-default nodeGloballyInstalled = '${ !Directory.Exists(nodeInstallDir) && TestCommand("node" , "--version") }'
 
 var nodeExe = 'node.exe'
 var npmZip = 'npm-${npmVer}.zip'
 var nodeDist = 'http://nodejs.org/dist/'
 var nodeUrl = '${nodeDist}v${nodeVer}/${nodeExe}'
 var npmUrl = '${nodeDist}npm/${npmZip}'
-var doInstall = '${!nodeGloballyInstalled && !Directory.Exists(nodeInstallDir)}'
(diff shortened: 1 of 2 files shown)


dougbu:
/cc @Praburaj @davidfowl @DamianEdwards 

this seems important to fix soon. we're using `bower` and `grunt` in more repos...

build/_npm.shade
```diff
@@ -6,5 +6,10 @@ var npmGloballyInstalled = '${TestCommand("npm" , "--version")}'
 var npmCmd = '${ npmGloballyInstalled ? "npm" : Path.Combine(nodeDir, "npm.cmd") }'
 
 node-install once='installNode'
+
+- // Ensure nodeDir exists. npmCommand is likely a package installation targeting that directory.
+exec program='cmd' commandline='/C mkdir ${nodeDir}' if='!IsLinux && !Directory.Exists(nodeDir)'
```
Eilon: Any reason not to just call `Directory.CreateDirectory`?

dougbu: none.  I just mimicked _robocopy-delete.shade rather than, say, _node-install.shade.  in retrospect, _robocopy-delete.shade is the outlier.

I'll switch to use `CreateDirectory()` in next iteration...

dougbu:
@Eilon update.  also changed _node-install.shade to make a more precise `doInstall` decision.

build/_node-install.shade
```diff
@@ -10,17 +10,17 @@ default nodeVer = '0.10.28'
 default npmVer = '1.4.9'
 default nodeExeSha = '628FFD6C3577068C00CEC9F6F897F0EC8F5212D9'
 default nodeInstallDir = '${Path.Combine(binDir, "nodejs")}'
-default nodeGloballyInstalled = '${ !Directory.Exists(nodeInstallDir) && TestCommand("node" , "--version") }'
```
dougbu: FYI this variable was no longer used elsewhere and `node-install`is always run with `once='installNode'`

Eilon:
Looks :shipit: to me but @DamianEdwards should probably give a thumbs up as well.

Praburaj:
great! :shipit:

DamianEdwards:
:shipit:

PR #170, opened 2015-03-07
Rename

troydai:
@NTaylorMullen any of these variables in build scripts are used on CI?


troydai:
@NTaylorMullen @anurse 

The dnvm included is copied from https://github.com/aspnet/dotnetsdk/tree/anurse/super-secret-open-source-magic-stuff. I'll keep it updated as it is being reviewed.

NTaylorMullen:
@troydai several, when this gets in we'll also need to run across all the CI configurations.. :shipit:

troydai:
Yup, can't verify it now, pending on kvm => dnvm and klr changes.

anurse:
looks good to me

troydai:
@NTaylorMullen ok if i get this in?

muratg:
:shipit:

build/_k-restore.shade
```diff
@@ -4,7 +4,7 @@ k-restore
     Restores nuget packages required for k projects. Downloads and executes k sdk tools.
 */}
 
-default restore_options='${E("KPM_restore_options")}'
+default restore_options='${E("NUGET3_restore_options")}'
```
Eilon: Hmm not sure we wanted to do the kpm -> nuget3 rename, did we? Not quite sure where that lands yet... @davidfowl ?

build/_k-restore.shade
```diff
@@ -4,7 +4,7 @@ k-restore
     Restores nuget packages required for k projects. Downloads and executes k sdk tools.
```
Eilon: Lots of other "k" stuff in here (including filenames) - do we have a work item for these?

muratg: These should be changed when we remove k.cmd and friends. There's a work item in the other repo.

build/_k-restore.shade
troydai: The name is entirely used internally. Not consumed by customers. They can be renamed anytime.

PR #173, opened 2015-03-09
`Console-Write` does not exist

dougbu:
- caused `dnvm use default -p` failures


troydai:
@dougbu the actual source of `dnvm.ps1` is here https://github.com/aspnet/dotnetsdk/tree/dev/src. The fix should be done over there and bring back here.

The fix itself looks good. Thanks.

troydai:
Move to dotnetsdk repo and :shipit:

dougbu:
Pushed as aspnet/dotnetsdk@0388cd0

/cc @troydai could you copy this to aspnet/Universe please?

troydai:
Yes, will do.

PR #180, opened 2015-03-11
Bower blocks build on first run

dougbu:
- aspnet/Mvc#2116
- run Bower non-interatively from our scripts
 - does not affect grunt-brower-task; programmatic Bower use never prompts
- see Bower discussion at bower/bower#1162
- FYI `$env:CI = 'TeamCity'` would work similarly but TeamCity does not set that


victorhurdugaci:
Looks good :shipit:

DamianEdwards:
:shipit:

dougbu:
commit f6ae75e

PR #181, opened 2015-03-11
Fix the `init` target

dougbu:
- correct check for whether target file is present
- correct `makefile.shade` filename
- do not copy master and release config files into every repo


troydai:
:shipit:

dougbu:
commit 456a38a

PR #184, opened 2015-03-12
Fix build failures when source is under a directory with spaces in the name

dougbu:
- aspnet/Mvc#2082
- problems specific to local `node` and related installations
 - quote authenticode verification target
 - quote local `node` library paths
 - quote local `npm.cmd` path (with ye olde `/S` double-quoting hack)
 - quote `--prefix` values in `npm install` commands
 - quote `rimraf` directories to delete

### DIFF
--- a/build/_npm.shade
+++ b/build/_npm.shade
@@ -5,12 +5,12 @@ default npmDir = '${currentDir}'
 node-install once='installNode'
 
 var npmFile = '${ Path.Combine(nodeDir, "npm.cmd") }'
-var npmCmd = '${ File.Exists(npmFile) ? npmFile : "npm" }'
+var npmCmd = '${ File.Exists(npmFile) ? ("\"" + npmFile + "\"") : "npm" }'
 
 - // Ensure nodeDir exists. npmCommand is likely a package installation targeting that directory.
 @{
   Directory.CreateDirectory(nodeDir);
 }
 
-exec program='cmd' commandline='/C ${npmCmd} ${npmCommand}' workingdir='${npmDir}' if='!IsLinux'
+exec program='cmd' commandline='/S /C "${npmCmd} ${npmCommand}"' workingdir='${npmDir}' if='!IsLinux'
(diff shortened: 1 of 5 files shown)


dougbu:
fixes here will help MusicStore repos with spaces in the name as well

/cc @troydai and anyone who has hacked `cmd /c` commands

troydai:
Looks good to me. :shipit: 
/cc @lodejard for advices on `cmd /S` usage.

build/_npm.shade
```diff
@@ -5,12 +5,12 @@ default npmDir = '${currentDir}'
 node-install once='installNode'
 
 var npmFile = '${ Path.Combine(nodeDir, "npm.cmd") }'
-var npmCmd = '${ File.Exists(npmFile) ? npmFile : "npm" }'
+var npmCmd = '${ File.Exists(npmFile) ? ("\"" + npmFile + "\"") : "npm" }'
```
muratg: Wouldn't just ```(" + nmpFile + ")``` work?

build/_verify-authenticode.shade
```diff
@@ -20,7 +20,7 @@ default expectedHash = ''
         WorkingDirectory = workingdir,
         FileName = signToolExe,
         RedirectStandardOutput = true,
-        Arguments = "verify /pa /v " + verifyFilePath,
+        Arguments = "verify /pa /v \"" + verifyFilePath + "\"",
```
muratg: Wouldn't using single quotes for the outer strings and not escaping quite inside work?

muratg:
:shipit:

build/_npm.shade
dougbu: no because we're in a C# code block; I'm just wrapping the expression that adds the quotes in parens to make the precedence more clear

build/_verify-authenticode.shade
dougbu: this is pure C# (within `@{ ... }`

dougbu:
commit 90294aa

PR #185, opened 2015-03-12
Skip credentials prompt for CI builds

victorhurdugaci:
The XPlat build hangs because git waits for user input. Added a flag that will not invoke git if credentials are not available.

@NTaylorMullen @troydai @ChengTian @suhasj


ChengTian:
:shipit:

Issue #186, opened 2015-03-17
Rename kpm to dnu


ChengTian:
This should be done after dnu.cmd is added (https://github.com/aspnet/dnx/issues/1428)

PR #187, opened 2015-03-20
Rename kpm to dnu in standard goal scripts

ChengTian:
parent https://github.com/aspnet/Universe/issues/186

@davidfowl @muratg 

Following the gradual adoption strategy. Only rename the kpm inside standard goal scripts, so nothing will be broken.

I see we still have standard goals with names like `_k-restore.shade` and `_kpm-build.shade`. We may want to replace `k` with `dnx` and replace `kpm` with `dnu` later.


muratg:
:shipit:

PR #188, opened 2015-03-25
Remove k command and use dnx instead

ChengTian:
https://github.com/aspnet/dnx/issues/1473

@davidfowl @muratg

### DIFF
--- a/build/_k.shade
+++ b/build/_k.shade
@@ -1,7 +1,7 @@
 @{/*
 
 k 
-    Run klr commands in your project. Executes k cmd.
+    Run dnx commands in your project. Executes dnx cmd.
 
 command=''
 dnvmUse=''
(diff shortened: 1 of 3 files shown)


build/_k.shade
```diff
@@ -1,7 +1,7 @@
 @{/*
```
dougbu: shouldn't this file also be renamed to _dnx.shade?
same for other _k-*.shade files?

ChengTian: No, it will break a lot of things. We want to make sure the changes that are more visible to external users are checked in for beta4 milestone ASAP. For those stuff mainly for internal use, we leave it for now, so that they don't block more important changes.

ChengTian:
@davidfowl , this should be the first one we check in

davidfowl:
:shipit:

PR #189, opened 2015-03-26
UpdateRepos

suhasj:
@pranavkm 

I have updated the list of repos. Let me know if I missed any

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -40,10 +40,15 @@ functions
             "Scaffolding",
             "Security",
             "SignalR-Server",
+            "SignalR-SQLServer",
+            "SignalR-Redis",
             "StaticFiles",
             "WebListener",
             "KestrelHttpServer",
             "WebSockets",
+            "Session",
+            "UserSecrets",
+            "CORS",
             "Entropy",
         };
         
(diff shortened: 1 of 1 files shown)


makefile.shade
```diff
@@ -421,9 +426,10 @@ functions
             var nonDefaultRepos = new[]
             {
               "KRuntime",
-              "MusicStore"
+              "MusicStore",
+              "Coherence"
```
pranavkm: Also needs to include `Coherence-Signed` and `dnvm`

pranavkm:
:shipit:

makefile.shade
suhasj: Coherence-signed doesn't have Nuget.config so I might need to special case it. I'll add dnvm

makefile.shade
```diff
@@ -421,9 +426,10 @@ functions
             var nonDefaultRepos = new[]
             {
               "KRuntime",
```
Praburaj: Should we call this XRE?

Praburaj:
why not for beta4 (to release branch)?

makefile.shade
suhasj: Github routes it XRE anyway but sure I'll change it

suhasj:
we have already branched to beta4 so this is for the next iteration

Praburaj:
Thanks :shipit:

victorhurdugaci:
I'm going to close this since it is so old. Please reopen if you still think it is necessary

PR #190, opened 2015-03-26
Make pushing to NuGet feed more resilient.

pranavkm:


build/_nuget-resilient-publish.shade
```diff
@@ -0,0 +1,42 @@
...
+        var retries = 0;
+        var packagePath = Path.Combine(sourcePackagesDir, package);
+        start:
+        try
+        {
+            NuGetPush (nugetFeed, packagePath, "");
+        }
+        catch
+        {
+            if (++retries == 3)
```
NTaylorMullen: Maybe have retry count configurable?

- Also: There's no issues with publishing the same package to myget right? Just wondering in the case that we fail but a package was published.

pranavkm: > Maybe have retry count configurable?

Ehh, we could do that if we see issues here.

> There's no issues with publishing the same package to myget right? Just wondering in the case that we fail but a package was published.

No.

NTaylorMullen:
:shipit:

PR #192, opened 2015-03-28
Set the file version environment variable

victorhurdugaci:
This change is to set an environment variable for the AssemblyFileVersion attribute. 

The version has 5 characters, the first 3 are the number of months since 2015/1/1 and the last 2 are the day of the month. @joeloff suggested this format since it is used in other places. For example, today's version is 00327.

Please review: @joeloff @pranavkm 

@Eilon, @davidfowl are you okay with the version in this format?

There is a DNX change coming soon that will actually write the attribute. 

Part of the fix for https://github.com/aspnet/dnx/issues/1504

### DIFF
--- a/build/BuildEnv.shade
+++ b/build/BuildEnv.shade
@@ -3,6 +3,22 @@ use namespace="System"
 
 functions
   @{
+    string CreateDayBasedVersionNumber()
+    {
+        var start = new DateTime(2015, 1, 1);
+        var now = DateTime.UtcNow;
+            
+        string version = "0";
+        // If the computer date is set before the start date, then the version is 0
+        if (now >= start)
+        {
+            var monthsSinceStart = (now.Year - start.Year) * 12 + now.Month;
+            version = monthsSinceStart.ToString("000") + now.Day.ToString("00");
+        }
+
+        return version;
+    }
+
     string BuildNumber
     {
         get 
--- a/build/_k-standard-goals.shade
+++ b/build/_k-standard-goals.shade
@@ -20,6 +20,10 @@ default Configuration='${E("Configuration")}'
   {
     E("DNX_AUTHOR", AUTHORS);
   }
+  if (string.IsNullOrEmpty(E("DNX_ASSEMBLY_FILE_VERSION")))
+  {
+    E("DNX_ASSEMBLY_FILE_VERSION", CreateDayBasedVersionNumber());
+  }
   if (string.IsNullOrEmpty(Configuration))
   {
     Configuration = "Debug";


Eilon:
That's a very wacky versioning scheme. Why not just YYMMDD? Or YYYYMMDD?

Also, why is it always 1.0.*.0? That would be very confusing given that our product versions are often not 1.0.*.0.

pranavkm:
File versions are capped at 65k (http://blogs.msdn.com/b/msbuild/archive/2007/01/03/why-are-build-numbers-limited-to-65535.aspx). So most leading year based schemes are broken.

build/BuildEnv.shade
```diff
@@ -3,6 +3,26 @@ use namespace="System"
...
+    string FileVersion 
+    {
+        get
+        {
+            DateTime start = new DateTime(2015, 1, 1);
+            DateTime now = DateTime.UtcNow;
+            
+            string version = "0";
+
+            // If the computer date is set before the start date, then the version is 0
```
pranavkm: This seems pretty unrealistic.

build/BuildEnv.shade
```diff
@@ -3,6 +3,26 @@ use namespace="System"
...
+            if (now >= start)
+            {
+                int monthsSinceStart = (now.Year - start.Year) * 12 + now.Month;
+                version = monthsSinceStart.ToString("000") + now.Day.ToString("00");
+            }
+
+            return "1.0." + version + ".0";
```
pranavkm: This needs to match the `AssemblyVersion`'s major and minor version. So Mvc is `6.0.` something.

pranavkm:
:watch:

joeloff:
I suggested the wacky scheme to give a maximum number of versions beyond the typical six year limitation we have today with the old .NET scheme

build/BuildEnv.shade
joeloff: +1, We'd want something like 6.0.12345 for MVC and 7.0.12345 for EF, etc.

davidfowl: Why the 3rd number out of interest?

victorhurdugaci:
Updated. You should be able to specify the pattern now with a wildcard. If for example MVC want v6, then it can set the pattern to `6.0.*.0` and `*` will be replaced by the number that we autogenerate

build/_k-standard-goals.shade
```diff
@@ -20,6 +20,14 @@ default Configuration='${E("Configuration")}'
   {
     E("DNX_AUTHOR", AUTHORS);
   }
+
+  var assemblyFileVersion = E("DNX_ASSEMBLY_FILE_VERSION");
+  if (string.IsNullOrEmpty(assemblyFileVersion))
+  {
+    assemblyFileVersion = "1.0.*.0";
```
muratg: Does Sake support comments?  Should we document how * is used?

victorhurdugaci:
Changed the logic to only set the number that we increment. DNX does all the rest

Issue #195, opened 2015-03-30
The dnvm in KoreBuild on the release branch points to the wrong feed


victorhurdugaci:
https://github.com/aspnet/Universe/blob/release/build/dnvm.ps1#L79
and
https://github.com/aspnet/Universe/blob/release/build/dnvm.sh#L13

BrennanConroy:
#191

PR #196, opened 2015-03-30
Fix the release feed in the dnvm copy

victorhurdugaci:
Fixes https://github.com/aspnet/Universe/issues/195

cc @pranavkm


victorhurdugaci:
There is already another PR for this...

PR #199, opened 2015-04-01
Skip projects that were not cloned

victorhurdugaci:
The cross plat CI fails because it tries to build projects that were not cloned.

@Eilon I would like to check this in for beta4

@pranavkm please review


ChengTian:
:shipit:

PR #202, opened 2015-04-15
Update dnu restore to use --parallel flag

pranavkm:


build/_k-restore.shade
```diff
@@ -6,5 +6,5 @@ k-restore
 
 default restore_options='${E("NUGET3_restore_options")}'
 
-exec program='cmd' commandline='/C dnu restore${restore_options}' if='!IsMono'
+exec program='cmd' commandline='/C dnu restore --parallel ${restore_options}' if='!IsMono'
```
dougbu: nit: `restore_options` is either empty or starts with a space.  no need for another space before it.

dougbu:
previous use of `--parallel` had two issues only one of which (unreliability) has been fixed.  what about disentangling the output so that it's somewhat meaningful?

realize some warnings and errors are output at the end.  but that output lacks the context running w/o `--parallel` provides.

dougbu:
:watch:

build/_k-restore.shade
pranavkm: Ehh, makes this easier to read. Plus it requires specifying the env variable with a leading space which seems unusual.

pranavkm:
> what about disentangling the output so that it's somewhat meaningful?

That would be a DNX issue, but it shouldn't stop us from using this.

Eilon:
If it works, :shipit: If it doesn't, undo it :smile:

PR #207, opened 2015-04-29
Removed master feed and fixed copying Nuget.config

suhasj:
@pranavkm 

- Removed myget/master feed
- Added check to copy Nuget.config only if present

### DIFF
--- a/build-template/NuGet.master.config
+++ b/build-template/NuGet.master.config
@@ -2,6 +2,5 @@
 <configuration>
   <packageSources>
     <add key="NuGet" value="https://nuget.org/api/v2/" />
-    <add key="AspNetMaster" value="https://www.myget.org/F/aspnetmaster/api/v2" />
   </packageSources>
 </configuration>
--- a/build-template/NuGet.master.config
+++ b/build-template/NuGet.master.config
@@ -2,5 +2,6 @@
 <configuration>
   <packageSources>
     <add key="NuGet" value="https://nuget.org/api/v2/" />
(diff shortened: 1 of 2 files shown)


build-template/NuGet.master.config
```diff
@@ -2,6 +2,5 @@
 <configuration>
   <packageSources>
     <add key="NuGet" value="https://nuget.org/api/v2/" />
-    <add key="AspNetMaster" value="https://www.myget.org/F/aspnetmaster/api/v2" />
```
pranavkm: Oops. We checked with @danroth27 and it sounded like we were pushing all our packages to `nuget.org`. We dorely on this for bringing in packages like `KoreBuild` that don't end up on NuGet.org.

@suhasj we might need to do another update to the master branches with this feed in it.

Eilon: Yeah we definitely don't push *everything* to NuGet.org. Master will always be needed.

suhasj: ah let me sync offline

suhasj:
Updated repos and reverted the changes in this PR

suhasj:
Ping @pranavkm

davidfowl:
What does this do?

suhasj:
super trivial change actually. Couple of repros like Coherence-Signed dont have a Nuget.config and hence this checks if that is the case then we don't replace the Nuget.config before pushing. It throws an error if this not there.

suhasj:
closing this since will be adding more features

PR #209, opened 2015-04-30
Consistent versioning of native binaries

moozzyk:
* Adding an MSBuild target file that generates a header containing file and product versions which will be used in .rc files in native projects across all repos.
  Note that keeping header generation in MSBuild is cleaner than generating it in a .shade file since the entire project is being built within a single build subsystem and therefore you can build the project directly (i.e. just with MSBuild) without having to do it with Sake (if the file did not exist)
* Updating native-compile target in _k-standard-goals.shade to pass version numbers.


build/_k-standard-goals.shade
```diff
@@ -95,10 +95,16 @@ default Configuration='${E("Configuration")}'
         var msbuildPath =  Path.Combine(programFilesX86, "MSBuild", msbuildVersions[i], "Bin", "MSBuild.exe");
         if (File.Exists(msbuildPath))
         {
+          var commonParameters =
+            " /p:Configuration=" + Configuration +
+            " /p:ProductVersion=1.0.0" +
```
moozzyk: Newbie question - where can I get the product version from? Some repos (dnx) define `PRODUCT_VERSION` in the shade file. Some repos define `VERSION` but it does not seem to be used (UserSecrets define `VERSION=0.1` but the dll is `1.0.0....`). At the moment I hardcoded it but there should be only one place when we define this version.

Eilon: Hmm I thought in general the project.json files had the version in it, though they have `x.0.0-*` and then the `*` comes from the build system. Local builds have `*` == `t<timestamp>`, whereas the CI has `*` == `<release quality>` e.g. beta4, rc1, etc. As such, I don't think there's a global way of doing this.

Now, it could very well be that it so happens that all our native projects are all 1.0.0 right now so it's technically correct.

moozzyk: Thanks @Eilon. I will leave this as is for now then.

moozzyk:
Ping :watch:

Eilon:
I guess it's :shipit: ? :smile:

PR #214, opened 2015-05-06
Use aspnetvnext feed as the default feed in dev branch

pranavkm:


build/dnvm.ps1
```diff
@@ -84,7 +84,7 @@ Set-Variable -Option Constant "CommandFriendlyName" ".NET Version Manager"
 Set-Variable -Option Constant "DefaultUserDirectoryName" ".dnx"
 Set-Variable -Option Constant "OldUserDirectoryNames" @(".kre", ".k")
 Set-Variable -Option Constant "RuntimePackageName" "dnx"
-Set-Variable -Option Constant "DefaultFeed" "https://www.nuget.org/api/v2"
+Set-Variable -Option Constant "DefaultFeed" "https://www.myget.org/F/aspnetvnext/api/v2"
```
Eilon: Eh, should we really be doing this? We should use the exact `dnvm` that we ship. We could instead update some env var or build.cmd/.sh to pass in the `--unstable` param or something.

pranavkm: Ok. We'll have to account for this as part of our build.cmd (to have it pass the right `DNX_FEED` \ flag) in release \ master branches. https://github.com/aspnet/Universe/issues/215 to track this

Eilon: Sounds good.

victorhurdugaci: Late comment on this: dnvm shouldn't be hardcoded to 2 feeds. We have at least 3 (NuGet, myget/vnext, myget/release). We need to find a better solution. Maybe dnvm should only have one configurable feed and not hardcode multiple options. Today, this breaks the release build on non CI machines because it brings beta7 dnx when build beta6 release

Eilon: `myget/release` is never used outside of testing/staging.

The only two feeds we care about are stable (NuGet) and unstable (AspNetVNext).

pranavkm: The CI pivots the DNX_FEED based on what branch we're on. The plan was to have this specified in `build.cmd` instead as part of #215.

victorhurdugaci: Sure, that would work too. We need to do that now or let our team know that they need to set the feed. You cannot build release today without that feed set.

Sort of related: DNX_FEED should be an argument, not an env var https://github.com/aspnet/dnvm/issues/354

PR #219, opened 2015-05-18
Replacing `directory delete` with `rimraf` in the `target-dir-clean`

moozzyk:
`directory delete` fails for read-only files


pranavkm:
:shipit:

PR #220, opened 2015-05-20
Add support to optionally generate xunit test xml files on build

gingters:
By setting the environment variable DNX_TEST_XML to any value, Korebuild now will generate test xml files in the test project directory for every tested framework.

When specifying a path in the DNX_TEST_XML_PATH environment variable, Korebuild will tell xunit to place the files in the defined directory.

It is sufficient to seit either of these to have xml files generated.


pranavkm:
@gingters the change looks :ok: but I'm not certain what its intent is. Who's looking at the xml file?

gingters:
@pranavkm The intent is that (at least currently) several CI systems (i.e. appveyor, travis etc.) need the test results as xml files to be able to show that the tests were okay or are failing. Currently the xUnit DNX runner only supports TeamCity, and it is unlikely that support for other CI systems comes anytime soon there.

With this change it is possible to have xml test result files generated at build, optionally in a given directory, so that one can configure his CI system of choice to pick up and evaluate the files.

One other idea had been to include test result file generation directly in the test command in the unit test projects. The problem with that approach was that you cannot specify different parameters (files) per framework, so if you had the tests run on multiple frameworks by Korebuild, (i.e. dnx451 and dnxcore), every run would overwrite the test results from the previous run, and you would lose test information for all others but the last run.

pranavkm:
>  several CI systems (i.e. appveyor, travis etc.) need the test results as xml files to be able to show that the tests were okay or are failing

AFAIK, xunit supports AppVeyor logging - https://github.com/aspnet/xunit/blob/dev/src/xunit.console/Visitors/AppVeyorLogger.cs#L8

Do you know what needs to change to get Travis to pick these files up?

gingters:
@pranavkm This is just the console runner that supports output to appveyor. The DNX runner does not, and they postponed that to 'somewhen': https://github.com/xunit/dnx.xunit/pull/4
So for now, you have to upload the test xml to the appveyor API after testing manually, for example like I currently do here: https://github.com/gingters/SmartDev.ConfigurationMapper
The problem with that is that I currently can't use 'the new way of building dnx stuff' Korebuild for that, but have to script my build and the test runs for each platform myself.

As far as I understood for travis, it's sufficient to put the xml files in the build artifacts folder to have travis grab and parse them. But getting my stuff running on travis is after I have appveyor running with the standard way of doing things.

In fact, this will be helpful with all CI systems (i.e. bamboo) that can't parse xunit test outputs directly but understand and nicely display the standard testresults.xml format.

pranavkm:
If it's generally useful, does it make sense to always produce an xml file (without specifying the environment variable)? We could drop these in the `artifacts\test\{project}\{framework}` directory so it mirrors the `artifacts\build` directory structure and we don't have to mangle the file name.

> The problem with that is that I currently can't use 'the new way of building dnx stuff' Korebuild for that, but have to script my build and the test runs for each platform myself.

Just an FYI - Sake \ KoreBuild is our choice for building dnx repositories locally and on our CI. There's a lot of stuff in there that's specifically designed based on how we structure our repositories and what our requirements were. Feel free to continue using if it works for you, but it certainly isn't the only one we endorse for people building things using DNX.

gingters:
@pranavkm Sorry for the delay. Was a bit distracted with other stuff. Always generating the files would be fine too, yes. I just thought opting in would be an approach that does not dictate anything, but generally having the files around after a build / test run does not hurt and having them in a default directory also gives people the option to simply grab the files without the need to configure anything.

Regarding Sake / KoreBuild, what would you suggest for alternatives for people 'just' creating libs / apps on top of DNX other than that? MSBuild does not really work on OS X / Linux yet.

pranavkm:
> Always generating the files would be fine too, yes.

Great. Feel free to update the PR. We had some discussion about doing something similar with `dnu build` \ `dnu restore` so that it's easier to import these into TeamCity.

> Regarding Sake / KoreBuild, what would you suggest for alternatives for people 'just' creating libs / apps on top of DNX other than that? MSBuild does not really work on OS X / Linux yet.

@davidfowl \ @danroth27 would you guys like to respond to this?

muratg:
@gingters Looks like this PR was open for quite a bit of time.  Will you be making the change @pranavkm suggested?  You'll probably also need to rebase as well.

muratg:
@pranavkm as this is quite an older PR, please feel free to make the update and push if you think this is useful. Otherwise, we should close this.

gingters:
Actually, we can close it. I was of the impression that using korebuild / sake would be the suggested way of doing things, but david fowler and just lastly also damian edwards on the community standup clarified that this is just internal stuff that does not really have relevance to the outside world.

While this would help me, I would doubt that others also use this and would have a benefit of this change. I probably will switch my own build process to something scripted in JS.

muratg:
Thanks @gingters, yes this is only for building ASP.NET 5 components.

PR #221, opened 2015-05-20
Fix `target-dir-clean` target

dougbu:
- not currently cleaning in most cases
- typo in 16bb0df


dougbu:
Current version results in errors like `The file '...\artifacts\build\Microsoft.AspNet.Razor\dnx451\Microsoft.AspNet.Razor.dll' already exists.` in the first project that copies into the artifacts folder. Happens any time you run two builds in a row i.e. don't remove the folder or `git clean`.

moozzyk:
:ship::it:

dougbu:
75c1efe

PR #222, opened 2015-05-21
Replacing rimraf in target-dir-clean with non-node.js based solution

moozzyk:
rimraf needs to pull node.js and is slow.

### DIFF
--- a/build/_k-standard-goals.shade
+++ b/build/_k-standard-goals.shade
@@ -65,7 +65,20 @@ default Configuration='${E("Configuration")}'
   k-restore
 
 #target-dir-clean target='clean'
-  rimraf rimrafDir='${TARGET_DIR}' if='Directory.Exists(TARGET_DIR)'
+  @{
+      if (Directory.Exists(TARGET_DIR))
+      {
+          var directory = new DirectoryInfo(TARGET_DIR);
+          directory.Attributes &= ~FileAttributes.ReadOnly;
+
+          foreach (var info in directory.GetFileSystemInfos("*", SearchOption.AllDirectories))
+          {
+              info.Attributes &= ~FileAttributes.ReadOnly;
+          }
+
+          directory.Delete(true);
+        }
+    }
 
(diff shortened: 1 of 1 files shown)


build/_k-standard-goals.shade
```diff
@@ -65,7 +65,19 @@ default Configuration='${E("Configuration")}'
   k-restore
 
 #target-dir-clean target='clean'
-  rimraf rimrafDir='${TARGET_DIR}' if='Directory.Exists(TARGET_DIR)'
+  @{
+      if(Directory.Exists(TARGET_DIR))
```
dougbu: space after `if`

build/_k-standard-goals.shade
```diff
@@ -65,7 +65,19 @@ default Configuration='${E("Configuration")}'
...
+      {
+          var directory = new DirectoryInfo(TARGET_DIR) { Attributes = FileAttributes.Normal };
+
+          foreach (var info in directory.GetFileSystemInfos("*", SearchOption.AllDirectories))
+          {
+              info.Attributes = FileAttributes.Normal;
```
dougbu: This loses all other attributes set in `info`.  Use `|=` instead and also do same update separately for `directory`.

dougbu:
:watch:

build/_k-standard-goals.shade
moozzyk: Why do we care if we are deleting the files?

dougbu: The system won't let you change some files to `Normal` but would let you remove them. So my `|=` suggestion should have been `&= ~FileAttributes.ReadOnly`. `ReadOnly` is the only bit worth twiddling here.

moozzyk: I see. Will do and update the PR.

moozzyk:
@dougbu updated the PR as per your comments

dougbu:
:shipit:

PR #224, opened 2015-05-22
Add `Quiet` default to enable conditionalizing `Log.Info()` calls

dougbu:
- use in new `_copy.shade` which overrides the Sake default


dougbu:
Anyone have objections? This only affects builds with `--quiet` on the command line.

dougbu:
664d718

PR #226, opened 2015-06-01
Add `restoreDir` parameter in `_k-restore.shade`

dougbu:
- enables use of this command where it's currently overridden
- in turn, this means `--quiet` would be honoured in those `makefile.shade` files

For example can use this hook to shut DNX build up a bit
- `makefile.shade` there calls `dnu restore` directly and ignores `--quiet`

### DIFF
--- a/build/_k-restore.shade
+++ b/build/_k-restore.shade
@@ -1,10 +1,16 @@
 @{/*
 
 k-restore
-    Restores nuget packages required for k projects. Downloads and executes k sdk tools.
+    Restores nuget packages required for DNX projects. Downloads and executes DNX sdk tools.
+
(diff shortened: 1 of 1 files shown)


muratg:
@dougbu Is this still a valid PR?

dougbu:
Out-of-date but definitely one way to minimize the spewage when building the DNX repo.

dougbu:
Do you want me 'ta rebase?

muratg:
Please sync up with @pranavkm to get his suggestion :)  If you do end up rebasing, please make sure to push :)  Otherwise, let's close this one.

dougbu:
22952e1

PR #236, opened 2015-06-13
Build projects in one go

victorhurdugaci:
Add support for `dnu build src/**`. Unless a particular environment variable is set, the build will still work as it does today so this change is not breaking and should not impact anything.

### DIFF
--- a/build/BuildEnv.shade
+++ b/build/BuildEnv.shade
@@ -33,4 +33,12 @@ functions
             return Environment.GetEnvironmentVariable("TEAMCITY_PROJECT_NAME") != null;
         }
     }
+    
+    bool IsBuildV2
+    {
+        get 
+        { 
+            return Environment.GetEnvironmentVariable("KOREBUILD_BUILD_V2") == "1"; 
+        }
+    }
   }
--- a/build/_k-standard-goals.shade
+++ b/build/_k-standard-goals.shade
@@ -30,6 +30,8 @@ default Quiet='${ false }'
     Configuration = "Debug";
     E("Configuration", Configuration);
   }
+  
+  Log.Info("Build v2: " + IsBuildV2);
 }
 
 #restore-npm-modules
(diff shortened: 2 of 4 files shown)


victorhurdugaci:
Please review: @pranavkm @anurse @troydai @davidfowl

build/_k-standard-goals.shade
```diff
@@ -30,6 +30,8 @@ default Quiet='${ false }'
     Configuration = "Debug";
     E("Configuration", Configuration);
   }
+  
+  Console.WriteLine("Build v2: " + IsBuildV2);
```
davidfowl: Log.Info

build/_k-standard-goals.shade
```diff
@@ -88,19 +90,29 @@ default Quiet='${ false }'
...
 #build-compile target='compile' if='Directory.Exists("src")'
   @{
-    var projectFiles = Files.Include("src/**/project.json").ToList();
-    if (ShouldRunInParallel)
+    if (IsBuildV2) 
     {
-        Parallel.ForEach(projectFiles, projectFile => DnuPack(projectFile, BUILD_DIR, Configuration));
+        const string buildPattern = "src/**";
+        
+        Console.WriteLine("Running new build");
```
davidfowl: Log.Info

build/_k-standard-goals.shade
dougbu: Remove

build/_k-standard-goals.shade
dougbu: Remove since information is obvious from logging of the actual `build` and `pack` commands.

build/BuildEnv.shade
```diff
@@ -33,4 +33,12 @@ functions
             return Environment.GetEnvironmentVariable("TEAMCITY_PROJECT_NAME") != null;
         }
     }
-  }
+    
+    bool IsBuildV2
+    {
+        get 
+        { 
+            return Environment.GetEnvironmentVariable("KOREBUILD_BUILD_V2") == "1";
```
dougbu: "V2" isn't helpful and other changes to `dnu` use are controlled by `"NUGET3_build_options"`, `""NUGET3_pack_options"`, and `"NUGET3_restore_options"` environment variables. Since those names are no longer where we're heading, suggest `"DNU_BUILD_ONE_SHOT"` for this and to rename the existing environment variables to start with "DNU". Second bit can come later but you're already touching most of the affected `.shade` files.

Might as well make everything uppercase since case isn't relevant on Windows and uppercase seems more common in KoreBuild.

dougbu:
:watch:

build/_k-standard-goals.shade
victorhurdugaci: I'll keep it. It makes easy to debug logs since I will enable it for some repos only. But I'll change it to `Log.Info`

build/_k-standard-goals.shade
victorhurdugaci: Removed

build/BuildEnv.shade
victorhurdugaci: That's a temporary variable that allows me to enable the new build for only a few repos. I plan to remove it in a week or so once the build is stable. But I don't want to affect the internal build while making these changes

victorhurdugaci:
Updated

Issue #231, opened 2015-06-05
CI builds will fail


dmccaffery:
Both of the CI build scripts [.travis.yml](https://github.com/aspnet/Universe/blob/664d718398ab0d94ce14dc35d13509248d79f9fd/.travis.yml#L4) and [appveyor.yml](https://github.com/aspnet/Universe/blob/664d718398ab0d94ce14dc35d13509248d79f9fd/appveyor.yml#L4) files include the following arguments:

- --quiet
- verify

These arguments are passed to sake as targets within build.sh. The #--quiet target is defined in the k-standard-goals include as part of KoreBuild. The #verify target is defined in the use-standard-goals include as part of Sake itself. Since neither of these includes are sourced in makefile.shade as part of the Universe build script (since it does its own git pull / build for each repository in the asp.net Universe), these CI builds will always fail.

I recommend changing the travis and appveyor commands to use the #verify-all target defined in makefile.shade, which will compile and verify each of the repos associated with the asp.net Universe. This will "prove" that all of the repositories are compiling and successfully passing tests. Otherwise; the CI build scripts should be removed (as they are useless the way they are defined now).

victorhurdugaci:
We are no longer going to build this repo on travis/appveyor: https://github.com/aspnet/Universe/pull/268

PR #237, opened 2015-06-13
Pass correct targets for travis and appveryor

victorhurdugaci:
The internal CI passes `pack` as argument. External CIs should do the same. 
`--quiet verify` doesn't make sense for Universe.

Fixes https://github.com/aspnet/Universe/issues/231

Please review @pranavkm @anurse

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: csharp
 sudo: false
 script:
-  - ./build.sh --quiet verify
+  - ./build.sh pack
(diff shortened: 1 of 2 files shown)


victorhurdugaci:
Ugh this still fails because sake invokes the NuGet in the wrong way. @lodejard do you take PRs for sake?

lodejard:
how does that require a sake change? the built-in shade files and targets and can be overridden in korebuild, no?

dmccaffery:
@lodejard short answer, yes, but the build script for Universe doesn't use k-standard-goals or k-standard-lifecycle, so its using Sake in the raw, currently.

victorhurdugaci:
Postponing for now

PR #238, opened 2015-06-14
build: Enables FreeBSD support

jasonwilliams200OK:
Changed hardcoded bash shebang to env to support multiple directory
structures (required to build on FreeBSD).

copy @josteink, @akoeplinger, @janhenke, @ajensenwaud, @saper

Related to: https://github.com/aspnet/dnvm/pull/295 & https://github.com/aspnet/dnx/pull/2066.


build/dnvm.sh
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
```
dougbu: Correct changes to this file have already been made. However the two `build.sh` changes remain relevant and haven't happened in this repo yet.

build-template/build.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 if test `uname` = Darwin; then
     cachedir=~/Library/Caches/KBuild
```
saper: I'd change `~` to `${HOME}` and we can just use `#! /bin/sh` here

saper: You could also add support for `/usr/bin/fetch` which is built-in on FreeBSD.

build/dnvm.sh
jasonwilliams200OK: @dougbu, @saper I have addressed your comments.

build-template/build.sh
```diff
@@ -1,7 +1,7 @@
-#!/bin/bash
+#!/bin/sh
 
 if test `uname` = Darwin; then
-    cachedir=~/Library/Caches/KBuild
+    cachedir=$HOME/Library/Caches/KBuild
```
dougbu: I understand this enables the script to use `/bin/sh`. But why make this one inconsistent with most others in the system? Don't think we have any plans to switch from `bash` in general. (May have a few `sh` scripts for test coverage or historical reasons.)

jasonwilliams200OK: I thought this is the only thing holding it back from being (portable) sh. But you are right, bash is much polished standard implementation, although on some systems, bash is not available by default http://stackoverflow.com/a/5725402
#! ${Rebased}

jasonwilliams200OK:
@dougbu, please PTAL. I have fixed the nits and rebased.

dougbu:
32f8c69

jasonwilliams200OK:
Thanks, I am compiling the list of repos, which require modification. I will open an issue in aspnet/Home. Is it ok?

dougbu:
An issue sounds good. Tracking in aspnet/Home or aspnet/Universe should be fine.

PR #239, opened 2015-06-15
Fix argument parsing for fast build

victorhurdugaci:
@troydai @pranavkm

### DIFF
--- a/build/_kpm-build.shade
+++ b/build/_kpm-build.shade
@@ -16,12 +16,22 @@ default build_options='${E("NUGET3_build_options")}'
 @{
     if (IsBuildV2)
     {
-        var projectFolders = projectFile.Split((char)';')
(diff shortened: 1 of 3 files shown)


build/_kpm-build.shade
```diff
@@ -16,12 +16,24 @@ default build_options='${E("NUGET3_build_options")}'
...
+                projectsToPack.Add(Path.GetDirectoryName(arg));
+            }
+            else
+            {
+                var projectFolders = Files.Include(arg + "/project.json").Select(proj => Path.GetDirectoryName(proj));
+                projectsToPack.AddRange(projectFolders);
+            }
+        }
+        
+        projectsToPack = projectsToPack.ToList();
```
troydai: Is this needed?

victorhurdugaci: Yup. It is used multiple times

Issue #247, opened 2015-07-11
Change how pin-version target works


pranavkm:
Right now, the pin-version target requires specifying the version of CoreCLR and Roslyn packages and uses this information to perform a find and replace in `project.json`s. This process doesn't inspire too much confidence since it doesn't discriminate between arbitrary lines in the file.

A better approach would be to pin the versions based on aspnetmaster feed \ Coherence drop. It removes the risk of accidentally replacing incorrect values and being repeatable.

troydai:
Looks good to me. Did we try it in a dry run?

troydai:
:shipit:

pranavkm:
Yeah, tried it on the current release branch (with the `git push` commented out) and the results looked good.

PR #249, opened 2015-07-20
Change pin-version to use the coherence build results

pranavkm:
Fies #247


pranavkm:
cc @troydai

makefile.shade
```diff
@@ -321,11 +276,11 @@ var buildTarget = "compile"
...
                     GitCommand(repo, string.Format("add -f {0}", gitFilePath));
                 }
-            } 
+            }
 
-            GitCommand(repo, "commit -am \"Updating json files to pin versions and build.cmd to pin KoreBuild and DNX\""); 
+            GitCommand(repo, "commit -am \"Updating json files to pin versions and build.cmd to pin KoreBuild and DNX\"");
 
-            GitCommand(repo, string.Format("push origin {0}:{0} -f", branch));
+            GitCommand(repo, "push origin master:master -f");
```
troydai: Fixed to `master` branch? Not allowing an option?

tools/PinVersion/Program.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
+using System.IO;
+using System.Linq;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using NuGet;
+
+namespace ConsoleApp1
```
troydai: Better namespace.

tools/PinVersion/Program.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
```
troydai: License header.

troydai:
Did we try it in a dry run?

makefile.shade
pranavkm: Yeah - I'm not super certain what the utility of the flag is. Seems like one more place for things to go bad (since we force push to the branch) plus we don't actually pin anything apart from the master branch (as yet).

troydai:
If it is not too much work, adding a feature to allow `-WhatIf` option would be helpful.

pranavkm:
For the `-whatif`, it sort of seems easier to not perform the actual git push and just print out the diff from there. Simply stating `Foo 1.0` was pinned to `Foo 1.0.-beta-13` doesn't tell us a great deal about what wasn't updated, where it was updated etc.

PR #258, opened 2015-08-07
Build tests and samples during regular build

victorhurdugaci:
cc @pranavkm @troydai @BrennanConroy

### DIFF
--- a/build/_k-standard-goals.shade
+++ b/build/_k-standard-goals.shade
@@ -92,30 +92,54 @@ default DNX_TOOLS_FEED = 'https://www.myget.org/F/dnxtools/api/v2'
 
 #ci-deep-clean .deep-clean target='clean' if='IsTeamCity'
 
-#build-compile target='compile' if='Directory.Exists("src")'
+#build-compile target='compile' if='!IsBuildV2 && Directory.Exists("src")'
   @{
-    if (IsBuildV2) 
+    var projectFiles = Files.Include("src/**/project.json").ToList();
+    if (ShouldRunInParallel)
     {
-        const string buildPattern = "src/**";
-        DnuPack(buildPattern, BUILD_DIR, Configuration);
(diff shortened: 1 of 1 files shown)


build/_k-standard-goals.shade
```diff
@@ -96,7 +96,7 @@ default DNX_TOOLS_FEED = 'https://www.myget.org/F/dnxtools/api/v2'
   @{
     if (IsBuildV2) 
     {
-        const string buildPattern = "src/**";
+        const string buildPattern = "src/**;test/**;samples/**";
```
pranavkm: The way we do publishing \ Coherence builds now is we grab everything in the artifacts directory. Now that we'll have tests and samples copied to artifacts (https://github.com/aspnet/Universe/pull/258/files#diff-4405a98bc7f6daec30b9a687f626cbd4R115) is there a way to filter them out of being published to the feed? Should we change the structure of packages so `src` -> `artifacts/build`, `test` -> `artifacts/test`, `samples` -> `artifacts/samples`?

pranavkm:
:watch:

victorhurdugaci:
Updated

BrennanConroy:
:sheep: it

pranavkm:
:shipit:

build/_k-standard-goals.shade
```diff
@@ -118,6 +110,38 @@ default DNX_TOOLS_FEED = 'https://www.myget.org/F/dnxtools/api/v2'
     }
   }
 
+#build-compile target='compile' if='IsBuildV2'
```
bricelam: I know it doesn't really matter, but should this run as part of `test-compile`instead?

victorhurdugaci: It doesn't matter :)

bricelam: I just say it because I often run `build compile` to get nupkgs as quickly as possible.

PR #259, opened 2015-08-11
Pin build scripts

kichalla:
This is to enable pinning build scripts like `build.cmd` to the exact version that we want.

@pranavkm 

Related PR:
https://github.com/aspnet/Diagnostics/pull/165

### DIFF
--- a/tools/PinVersion/Program.cs
+++ b/tools/PinVersion/Program.cs
@@ -90,16 +90,16 @@ namespace PinVersion
             }
 
             var buildCmdFile = buildCmdFiles[0];
-            var buildCmdFileContent = File.ReadAllText(buildCmdFile);
-            buildCmdFileContent = buildCmdFileContent.Replace(
-                "SET BUILDCMD_KOREBUILD_VERSION=\"\"",
-                $"SET BUILDCMD_KOREBUILD_VERSION={korebuildVersion}");
-            buildCmdFileContent = buildCmdFileContent.Replace(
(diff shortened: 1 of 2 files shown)


tools/PinVersion/Program.cs
```diff
@@ -12,62 +13,138 @@ namespace PinVersion
...
+                    });
+                });
+
+                app.Command("buildscripts", command =>
                 {
-                    latestPackage = packageRepository.FindPackage(dependency.Name);
-                    if (latestPackage != null)
+                    command.Description = "Pins build scripts";
+                    var repositoryPath = command.Argument("[Repository Path]", "Path to the cloned repository");
+                    var koreBuildVersion = command.Argument("[Korebuild Version]", "Korebuild package version");
```
pranavkm: 1. Could we infer the version for these packages (from the Coherence drop share) rather than specifying them?

2. Does it make sense to have these in the makefile.shade for Universe (which calls this method) rather than change this? The primary reason the pining above was written as an executable and not a sake script was because it needed to reference NuGet.Core \ Json.Net. This piece of code could possibly be written in sake.

kichalla: Regarding 1. :+1:

kichalla: Updated. @pranavkm

pranavkm:
:shipit:

kichalla:
commit 39408739432b4b969e6e9748d7fe0155de53f7d8

PR #278, opened 2015-09-16
Upgrade dnvm to beta8

natemcmaster:
Current version of dnvm is fairly old.

### DIFF
--- a/build/dnvm.ps1
+++ b/build/dnvm.ps1
@@ -67,8 +67,8 @@ function _WriteOut {
 
 ### Constants
 $ProductVersion="1.0.0"
-$BuildVersion="beta6-10394"
-$Authors=".NET Foundation"
+$BuildVersion="beta8-15516"
+$Authors="Microsoft Open Technologies, Inc."
 
 # If the Version hasn't been replaced...
 # We can't compare directly with the build version token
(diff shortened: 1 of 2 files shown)


build/dnvm.ps1
```diff
@@ -177,11 +184,24 @@ if($CmdPathFile) {
...
+    }
+
+    $env:DNX_GLOBAL_HOME="$GlobalHome"
+    $env:DNX_HOME="$env:DNX_HOME;$env:DNX_GLOBAL_HOME"
+    $UnencodedHomes = "$UnencodedHomes;$GlobalHome"
+}
+
 $UnencodedHomes = $UnencodedHomes.Split(";")
 $RuntimeHomes = $UnencodedHomes | ForEach-Object { [Environment]::ExpandEnvironmentVariables($_) }
 $RuntimeDirs = $RuntimeHomes | ForEach-Object { Join-Path $_ "runtimes" }
```
dougbu: Until we fix aspnet/Home#899 and / or aspnet/dnvm#422, we should not take this version of `dnvm.ps1`. The problem is that all variables in `$UnencodedHomes` are already expanded when `[Environment]::ExpandEnvironmentVariables()` is called above i.e. they're referenced before they are definitely defined -- `$env:DNX_HOME` and `$env:DNX_GLOBAL_HOME` in particular. That leads to failures because `$_` is empty here.

dougbu:
:watch:

natemcmaster:
See #288

PR #280, opened 2015-09-16
Update build-template

natemcmaster:


pranavkm:
Could you also update NuGet.master.config while you're at it?

:shipit: once that's done.

natemcmaster:
Updated.

PR #286, opened 2015-09-22
Tweaks for Universe Tests

pranavkm:
* Use --quiet for git pull and git clone so git fetch messages do not show up in CI error logs.
* Use --parallel for default verify target.


pranavkm:
cc @victorhurdugaci \ @dougbu

victorhurdugaci:
Looks good :shipit:

makefile.shade
```diff
@@ -82,7 +82,7 @@ var buildTarget = "compile"
 
 #change-default-build-target
     @{
-        buildTarget = "verify";
+        buildTarget = "--parallel verify";
```
dougbu: Need a way to disable `--parallel` if enabling it in this fashion. This also isn't a great default due to the spew of messages coming from the different threads -- makes finding a problem more difficult. (Could change the `Parallel.ForEach()` blocks to spew with `msbuild`-like thread ids but that's probably too much work for this private build system.)

dougbu:
:shipit: after adding `--noparallel` or similar.

makefile.shade
pranavkm: Will remove this for now. Just doing the `--quiet` to reduce CI spew.

Issue #272, opened 2015-09-07
Support building with Core CLR (on Linux)


dougbu:
Points to cover:
- [x] Use `IsLinux` and not `IsMono` when appropriate in `*.shade` files.
- [x] Remove `-runtime CLR -arch x86` from `dnvm use default` commands in `build.cmd` and `build.sh`. (Should run with whatever the user has chosen as `default`, even after aspnet/dnvm#405 is fixed.)
- [ ] Change `_k-test.shade` to skip tests for a different runtime than the current `default` on Linux. (Could also be more explicit on Windows and use `default -runtime CLR` as well as current `default -runtime CoreCLR`.)
- [x] Stretch: Extend `_k-test.shade` / `_k.shade` to support multiple runtimes on Linux. This requires either aspnet/dnvm#265 or aspnet/dnvm#405 to be fixed everywhere. Tests on Linux currently run using only the `default` runtime and the scripts assume that `default` is full CLR.
- [x] Optional: Work around aspnet/dnx#2566, allowing `build.sh` and `build.cmd` to build only for Core CLR. (`export "NUGET3_build_options= --quiet --framework dnxcore50"` and similar hacks mostly work.)

BrennanConroy:
Bullet #2 dnvm issue has been fixed

BrennanConroy:
aspnet/dnvm#265 and aspnet/dnvm#405 both fixed

dougbu:
Thanks @BrennanConroy !  Have both fixes been pushed into Universe / KoreBuild?

BrennanConroy:
Oh... I think we manually update that repo once in a while?

Sent from my HTC

----- Reply message -----

muratg:
@dougbu is this still happening in beta 8? do we need to move it out?

dougbu:
Yes, move it out. We've now got a `dnvm.sh` that supports this work but I haven't spent the necessary time in the `.shade` (files). My plan is to do the work early in RC1.

dougbu:
2f76bcb

PR #294, opened 2015-10-06
Support building with Core CLR (on Linux)

dougbu:
- hits all of the checkboxes in #272

Specify test runtime explicitly and support testing w/ DNX Core on Linux
- turn on DNX Core on Linux testing with `--test-dnxcore` target
  - or `KOREBUILD_TEST_DNXCORE` environment variable

Work around aspnet/dnx#2566
- builds fail with DNX Core on Linux otherwise

Do not attempt to install nodejs if not on Windows

Use user's `default` DNVM when `%SKIP_DNX_INSTALL%` defined on Windows
- enables builds with DNX Core
- should be the default for new repos; older repos can opt in

Use consistent case for all `--quiet` environment variables
  - also name these variables consistently; start w/ `KOREBUILD_`
- environment variables are case-sensitive on Linux and mixed case was annoying
- also get rid of leading space in these environment variable values

Check `IsLinux` and not `IsMono`
- `IsMono` is always `true` on Linux since Sake always runs in Mono there
  - but `IsLinux` is the right question

nits:
- remove tabs from a few files
- remove compilation warnings in local `makefile.shade`

### DIFF
--- a/build/_node-install.shade
+++ b/build/_node-install.shade
@@ -21,7 +21,9 @@ var npmInstallZipPath = '${Path.Combine(nodeDir, npmZip)}'
 var nodeInstalled = '${ File.Exists(nodeInstallExePath) }'
 
 default nodeGloballyInstalled = '${ !nodeInstalled && TestCommand("node", "--version") }'
-var doInstall = '${ !(nodeGloballyInstalled || nodeInstalled) }'
+
+-// Command simply fails on Linux if nodejs is not available.
+var doInstall = '${ !IsLinux && !(nodeGloballyInstalled || nodeInstalled) }'
 
 @{
   if (doInstall) {
--- a/build-template/build.cmd
+++ b/build-template/build.cmd
@@ -19,20 +19,20 @@ copy %CACHED_NUGET% .nuget\nuget.exe > nul
 :restore
 IF EXIST packages\KoreBuild goto run
 IF %BUILDCMD_KOREBUILD_VERSION%=="" (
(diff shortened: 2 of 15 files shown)


BrennanConroy:
Is there any plan to add IsUnix to Sake or KoreBuild for OSX?

build/_node-install.shade
```diff
@@ -21,7 +21,9 @@ var npmInstallZipPath = '${Path.Combine(nodeDir, npmZip)}'
 var nodeInstalled = '${ File.Exists(nodeInstallExePath) }'
 
 default nodeGloballyInstalled = '${ !nodeInstalled && TestCommand("node", "--version") }'
-var doInstall = '${ !(nodeGloballyInstalled || nodeInstalled) }'
+
+-// Command simply fails on Linux if nodejs is not available.
```
BrennanConroy: Because it tries to run an exe, we should try to change it to work on Linux

dougbu: I don't feel strongly about this because the installation process is distribution-specific and may require `sudo`.

pranavkm:
:shipit:

dougbu:
> Is there any plan to add IsUnix to Sake or KoreBuild for OSX?

Covered already. Sake's `Environment.shade` file implements `IsLinux` and it returns `true` for `PlatformId.Unix`, `PlatformId.MacOS`, and `(PlatformId)128` (forget what that means).

BrennanConroy:
Oh cool!

dougbu:
2f76bcb

Issue #290, opened 2015-09-28
Switch to coreclr for restore (at least on xplat)


davidfowl:
It works better than mono.

/cc @dougbu @pranavkm

dougbu:
Self-assigning and moving into this milestone. Travis is just too annoying.

dougbu:
Reopening since 4e87b0c was not a fix for this bug.

dougbu:
Part one of another run at this bug: aspnet/DNX@a2ed65a

dougbu:
Part two: 419d3a3

dougbu:
Travis runs on Ubuntu 14.04 and performs `dnu restore` using .NET Core in all of our repos now.

PR #295, opened 2015-10-09
Always restore w/ Core CLR and build w/ CLR or Mono

dougbu:
- #290
- remove workaround for aspnet/dnx#2566; not needed if building with CLR or Mono
- install Core CLR DNX in `build.sh`
  - must be applied in all repos using Travis before this PR goes in
- add `Dnu()` and `Dnx()` macros
  - remove `K()` macro from `_k-test.shade`; use `Dnx()` instead
- incorporate per-repo changes not yet applied to Universe templates
  - use newer `NuGet.exe`
  - Universe itself does not need the Korebuild package or `dnvm`

Changes partially revert 618a73b e.g. always start build with CLR or Mono DNX

nits:
- reduced DNX switches and `dnvm` calls in `build.cmd`
- base NuGet skip decision on _both_ KoreBuild and Sake's installation status
- make `Quiet` variable work more reliably; small changes often led to compilation problems
  - requires a small change in DNX repo at the same time
- remove some trailing whitespace and tabs
- add a bit more description to `_k.shade`


dougbu:
Have pushed branches for repos that need to react, in particular to ensure DNX Core is available for `dnu restore` in Travis (and Linux in general) builds. Will open a couple of example PRs but can look at same-named branches as you wish.

build/_k-restore.shade
```diff
@@ -12,5 +12,4 @@ default restoreDir = '${ currentDir }'
 
 default restore_options=' ${ E("KOREBUILD_DNU_RESTORE_OPTIONS") }'
 
-exec program='cmd' commandline='/C dnu restore --parallel${ restore_options }' workingdir='${ restoreDir }' if='!IsLinux'
-exec program='dnu' commandline='restore${ restore_options }' workingdir='${ restoreDir }' if='IsLinux'
+dnu command='restore${ restore_options }' workingDir='${ restoreDir }' dnvmUse='default -runtime coreclr'
```
pranavkm: Could we add `--parallel` here?

pranavkm:
:shipit:

PR #299, opened 2015-10-14
Enabling setting version/alias of coreclr runtime tests should be run on

moozzyk:


moozzyk:
This is to fix the dnx repo where CoreClr tests are not run using the runtime that was just built but using the default runtime. Since we always upgrade to the latest avaiable runtime the `default` runtime is never the one we just built.

/cc @davidfowl @pranavkm - I am open to suggestions if you know how to fix it easier or better - this was the simplest thing I could come up with.

https://github.com/aspnet/dnx/pull/2977

build/_k-test.shade
```diff
@@ -3,6 +3,7 @@ use import="Environment"
 
 default NO_PARALLEL_TEST_PROJECTS='${E("NO_PARALLEL_TEST_PROJECTS")}'
 default KOREBUILD_TEST_DNXCORE='${E("KOREBUILD_TEST_DNXCORE")}'
+default CORECLR_RUNTIME_VERSION='${E("CORECLR_RUNTIME_VERSION")}'
```
dougbu: Suggest this environment variable name is too specific for KoreBuild. Name should mention testing and align with `KOREBUILD_TEST_DNXCORE`. How about `KOREBUILD_TEST_DNXCORE_VERSION`?

Also seems strange to support this override only for DNX Core testing. Should CLR / Mono testing call `Dnx()` with a `dnvmUse` parameter when e.g. `KOREBUILD_TEST_VERSION` is non-empty?

dougbu:
:watch:

build/_k-test.shade
moozzyk: I guess the fundamental flaw here is that we run tests for dnx451 differently than dnxcore50. I actually thought about adding the `dnvmUse` parameter but what the default value would be for dnx451? Wouldn't using the `default` runtime is a change in the behavior that affects all the repos? 
Interestingly dnx repo is adding the path to the runtime before running the test which would work if we ran CoreClr tests the same way we run Desktop Clr tests.

dougbu: We run the tests identically except `_k-test.shade` assumes `build.cmd` and `build.sh` ensure `$Path` contains the full CLR `default` runtime. If you want it a special case for DNX testing, add it for both runtime flavors. Main complication is that it'll be `{alias} -runtime clr` on Windows and `{alias} -runtime mono` everywhere else.

Also, I wasn't suggesting a change to the `default` alias. Add an alias for the just-built, just-installed runtime. Or add two -- one for DNX Core and one for Mono or CLR.

moozzyk: This is not true. For dnx451 we run just `cmd /C dnx test` while for dnxcore50 we run ` cmd /C "C:\Source\dnx\packages\KoreBuild\build\dnvm" run default -runtime coreclr test`. So we run tests differently. On top of that - because we use dnvm in the second case the build will pass even if tests failed because the original exit code is __lost__.

dougbu: > So we run tests differently

If there's a real difference between `dnx test` and `dnvm run default test` then a `dnvm` issue has regressed recently.

> the original exit code is **lost**

@BrennanConroy fixed aspnet/dnvm#428 a while ago. I haven't seen any evidence of a regression there. But  reopen that issue if it's a problem again.

moozzyk: I actually think it is not in dnvm. Sake runs:
`cmd /C "C:\Source\dnx\packages\KoreBuild\build\dnvm" run default -runtime coreclr test` 
and the exitcode is lost. 
When I run:
`& "C:\Source\dnx\packages\KoreBuild\build\dnvm" run default -runtime coreclr test`
the exit code is preserved.
I also confirmed that dnvm in the KoreBuild have the fixes for  aspnet/dnvm#428

moozzyk: I guess I will commit this is as to unblock dnx. The issue with the exit code needs to be tackled separately. Longer term I think we may not need this code at all.

moozzyk: (I will change the name of the variable to `KOREBUILD_TEST_DNXCORE_VERSION` as you proposed)

dougbu: @moozzyk you're not unblocking DNX as this PR stands. Not sufficient to test CLR with the old DNX and Core with the new DNX.

Further if `cmd` has the problems you mentioned, failures when testing w/ the new DNX won't break the build.

moozzyk: CLR is tested correctly since it sets %PATH% before invoking this method. Since for Clr we do just `dnx test` it runs using the correct runtime. We do the same for CoreClr but this method happily overwrites the setting running against `default`. 
Another option would be to do something similar in this method - get the path of the current runtime and change it to coreclr if it is not already coreclr and run without dnvm which would solve both issues.

dougbu: Why not update `$Path` for the DNX Core testing and have this file support executing only DNX Core tests the same way it executes only Mono tests on Linux? That is update `$Path` before calling `DnxTest()` or running `k-test` and have that macro / command use `dnx test` if the project targets `dnxcore50`.

PR #302, opened 2015-10-16
Call correct build.sh when building repos.

cesarbs:
Mono has inconsistent behavior in Process.Start (which is what Exec calls). Consider this directory structure

base
base/build.sh
base/sub
base/sub/build.sh

If $PWD is base and we call Process.Start with ProcessStartInfo { FileName = "build.sh", WorkingDirectory = "sub" }, the expected behavior (i.e. what .NET does on Windows) is that base\sub\build.sh would get started with a working directory of base/sub. However, base/build.sh gets called instead. For whatever reason, the same does not happen when trying to start a binary executable.

Because of this behavior, the universe build is broken on Linux. After change d679c85aa386f00afde6ce7062f286dedb76c7f9, the universe build.sh script does not get KoreBuild anymore and all builds fail because universe/build.sh is being called instead of universe/<repo>/build.sh.

The solution is to explicitly pass Exec the path to each repo's build.sh script.

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -360,7 +360,7 @@ var buildTarget = "compile"
 
                 if (IsMono)
                 {
-                    Exec("build.sh", buildTarget, repo);
+                    Exec(Path.Combine(repo, "build.sh"), buildTarget, repo);
                 }
                 else
                 {
@@ -419,7 +419,7 @@ var buildTarget = "compile"
         {
             if (IsMono)
             {
-                Exec("build.sh", "install", repo);
+                Exec(Path.Combine(repo, "build.sh"), "install", repo);
             }
             else
             {


cesarbs:
@dougbu @davidfowl

dougbu:
Note the same issue would also lead to failures when building repos (subdirectories) such as aspnet/MusicStore that require x64 DNX packages.

makefile.shade
```diff
@@ -360,7 +360,7 @@ var buildTarget = "compile"
 
                 if (IsMono)
                 {
-                    Exec("build.sh", buildTarget, repo);
+                    Exec(Path.Combine(repo, "build.sh"), buildTarget, repo);
```
dougbu: Minor but worthwhile for consistency: Do the same fixup w/ `build.cmd` in the `else` branch here and below.

dougbu:
:shipit:

makefile.shade
cesarbs: I'm not sure that's a good idea - I'd rather not touch it since it behaves as expected there.

PR #311, opened 2015-10-23
fix Linux RID on Travis

anurse:
Going to wait until aspnet/dnx#3048 is merged so I know that it actually works ;). Once this is in, we can remove this from DNX's makefile

/cc @davidfowl @pranavkm

### DIFF
--- a/build/_k-standard-goals.shade
+++ b/build/_k-standard-goals.shade
@@ -40,6 +40,14 @@ default NUGET_FEED = 'https://api.nuget.org/v3/index.json'
   }
 
   Log.Info("Build v2: " + IsBuildV2);
+
+  // It looks like it isn't possible to do this in .travis.yml because we only want
+  // to do it on Linux and I couldn't find an easy way to do that kind of selective
+  // environment variable set.
+  if(E('TRAVIS') == "true" && CanBuildForLinux)
+  {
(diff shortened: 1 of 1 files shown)


build/_k-standard-goals.shade
```diff
@@ -40,6 +40,14 @@ default NUGET_FEED = 'https://api.nuget.org/v3/index.json'
   }
 
   Log.Info("Build v2: " + IsBuildV2);
+
+  // It looks like it isn't possible to do this in .travis.yml because we only want
+  // to do it on Linux and I couldn't find an easy way to do that kind of selective
+  // environment variable set.
+  if(E('TRAVIS') == "true" && CanBuildForLinux)
```
pranavkm: `if (` formatting. Also, does this need to work for Mac - does `CanBuildForLinux` work in that scenario?

pranavkm:
:shipit:

build/_k-standard-goals.shade
anurse: It might... I'll see what the RID story looks like on OSX

muratg:
This good to check-in?

davidfowl:
CanBuildForLinux doesn't exist...

anurse:
This is super old, lets just kill it

PR #312, opened 2015-10-23
Fix bug in README

fearthecowboy:
It appears that unless one is part of the Universe project, you can't clone it via ssh. (or anonymously)
Changed to https://

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ This repo is to build the whole ASP.NET 5 stack.
 
 ## Getting started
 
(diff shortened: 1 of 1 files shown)


pranavkm:
It's a public repository - cloning via ssh should work if you have a public key set up with GitHub. What error are you seeing?

fearthecowboy:
Ah, I wasn't running pageant at the time. Interesting. Still, you can't clone it anonymously :)

PR #325, opened 2015-11-29
Adding directory existence check for nuget-install

JunTaoLuo:
Removing the placeholder project in https://github.com/aspnet/ServerTests/pull/14 broke our CI. This is because nothing was built and the nuget-install step tries to publish from the non-existent `artifacts/build` directory. I'm thinking that the easiest fix is to just check for the existence of the build directory for nuget-install, thoughts? Is there a way for me to verify this fixes the issue before merging the PR?

@pranavkm @victorhurdugaci @rynowak

### DIFF
--- a/build/_k-standard-goals.shade
+++ b/build/_k-standard-goals.shade
@@ -202,7 +202,7 @@ default NUGET_FEED = 'https://api.nuget.org/v3/index.json'
     }
   }
 
-#nuget-install target='install' description='Install NuGet packages to local repo'
+#nuget-install target='install' if='Directory.Exists("src")' description='Install NuGet packages to local repo'
   kpm-publish sourcePackagesDir='${BUILD_DIR}' targetPackagesDir='${E("PACKAGES_PUBLISH_DIR")}'
   nuget-resilient-publish sourcePackagesDir='${BUILD_DIR}' nugetFeed='${E("NUGET_PUBLISH_FEED")}' if='!string.IsNullOrEmpty(E("NUGET_PUBLISH_FEED"))'
 


victorhurdugaci:
I don't think this is the right fix. If the output folder doesn't exist, we shouldn't silently ignore that.

The right fix should be an opt-in model that each repo can individually and explicitly enable. One solution would be to override the `nuget-install` target with a blank target in the server tests repo.

dougbu:
@victorhurdugaci I'm curious: Why not silently ignore non-existent output folder? That's what the `.shade` files do when `src` folder doesn't exist for some reason. Other than a couple of repos which completely override `_k-standard-goals.shade`, we don't have the complications you're suggesting elsewhere.

victorhurdugaci:
Apologies, I didn't explain why I think it's not the right fix. If a repo has projects in the `src` folder, we expect on a successful build to also have `artifacts\build`. `nuget-install` should:

1. Fail if the `src` folder is there but `artifacts\build` is not (a prerequisite failed or was not met)
2. Do nothing if there's no project to build in the first place (no prereq required)

By checking for `artifacts\build`, instead of failing in case `#1` when a prereq is not met, the build script will just silently ignore the missing folder. Thus, you remove check `#1`.

Maybe you could check for the `src` folder instead of `BUILD_DIR`?

JunTaoLuo:
Updated. Thanks @victorhurdugaci. I do think the check should be made in `_k-standard-goals.shade` since we shouldn't assume all repos have `src` and require explicit opting out. It's also consistent with other goals where we check for existence for `src`, `test`, or `samples` before proceeding.

dougbu:
:shipit: (nice: even shorter than the original output folder check :smiley_cat:)

Issue #326, opened 2015-11-30
Environment flags used during build should be checked for =="1" instead of existence


JunTaoLuo:
Some build flags are set by existing https://github.com/aspnet/Universe/blob/dev/build/_k-standard-goals.shade#L361. This could be confusing since the setting the value to "0" still turns it on. We should check whether the value is "1" instead like https://github.com/aspnet/Universe/blob/dev/build/_k-standard-goals.shade#L366.

cc @victorhurdugaci

dougbu:
Disagree since `true`, `True`, and so on would inexplicably not do the needful. What is the use case for setting an environment variable to a value that means the same thing as not setting it at all?

JunTaoLuo:
Isn't it better to have consistency across different settings? We can even whitelist `1`, `true`, `True` as recognized settings that turns flags on but it is rather confusing that `0` would turn on the settings. In fact I already suspect that a setting in our build is accidentally turned on because of this. 

Also, right now some settings are only turned on if the value is set to `1` and some are checked for existence. We should at least make it consistent across the board.

Eilon:
I think we often do explicitly check for `1`, `true`, and `True`, so this sounds fine. Checking existence sounds super error-prone to me.

dougbu:
@JunTaoLuo @Eilon this is completely internal: We're talking about environment variables that are rarely used off the CI machines. So, what errors are we concerned about? Why do more than check for existence -- make that consistent?

Eilon:
To prevent bugs in the build system. Internal or not, it's something we all use every day...

cesarbs:
My take, as I said in https://github.com/aspnet/Universe/pull/330, is to check that something is *not* `0` or `false` and treat all else as true. However I do concede it might be overkill and just checking for `== 1` could be all we need.

muratg:
Per @Eilon's decision, we decided to just check for 1 and true (lowercase as far as I'm concerned) and ignore everything else.

muratg:
@victorhurdugaci Is this something we should do in dorkbuild?

Eilon:
Closing because we are not planning to make changes to this or it is no longer needed.

PR #330, opened 2015-12-08
Skip tests on Mono if KOREBUILD_TEST_SKIPMONO is set.

cesarbs:
This change will enable us to run tests on CoreCLR on the CI.

cc @muratg @Eilon @victorhurdugaci

### DIFF
--- a/KoreBuild-dotnet/build/_dotnet-test.shade
+++ b/KoreBuild-dotnet/build/_dotnet-test.shade
@@ -3,6 +3,7 @@ use import="Environment"
 
 default NO_PARALLEL_TEST_PROJECTS='${E("NO_PARALLEL_TEST_PROJECTS")}'
 default KOREBUILD_TEST_DNXCORE='${E("KOREBUILD_TEST_DNXCORE")}'
+default KOREBUILD_TEST_SKIPMONO='${E("KOREBUILD_TEST_SKIPMONO")}'
 
 @{/*
 
@@ -59,9 +60,16 @@ projectFile=''
                     testArgs = " -parallel none";
                 }
 
-                // Dotnet("test" + testArgs, projectFolder);
+                if (!IsLinux ||
+                    (!string.Equals(KOREBUILD_TEST_SKIPMONO, "1") &&
+                     !string.Equals(KOREBUILD_TEST_SKIPMONO, "true")))
+                {
+                    // Dotnet("test" + testArgs, projectFolder);
+                }
             }
-            else if (!IsLinux || !string.IsNullOrEmpty(KOREBUILD_TEST_DNXCORE))
+            else if (!IsLinux ||
+                     string.Equals(KOREBUILD_TEST_DNXCORE, "1") ||
+                     string.Equals(KOREBUILD_TEST_DNXCORE, "true"))
             {
                 // Dotnet("test" + testArgs, projectFolder, "default -runtime coreclr");
             }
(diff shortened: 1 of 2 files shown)


victorhurdugaci:
Why not use the `FrameworkSkipConditionAttribute`?

build/_k-test.shade
```diff
@@ -58,7 +59,10 @@ projectFile=''
                     testArgs = " -parallel none";
                 }
 
-                Dnx("test" + testArgs, projectFolder);
+                if (!IsLinux || string.IsNullOrEmpty(KOREBUILD_TEST_SKIPMONO))
```
muratg: Sorry, I'm not sure if I understand the logic inside `if`.  When do we execute this command and why the difference between Linux and OS X?

cesarbs:
@victorhurdugaci Because I'd have to put it on every test in every repo. We want to be able to run tests on CoreCLR **only** across the board. The idea is to have two CIs per non-Windows system, one where we run tests on Mono and another where we run tests on CoreCLR.

build/_k-test.shade
cesarbs: That's a poorly named variable. It's set to true on OS X.

victorhurdugaci: Check if `1` of `true`, not that that if it's empty or not. Otherwise, setting it to "false" or "foo" will skip the tests too

victorhurdugaci:
Can you wait until I merge the dotnet support (1h or so) and then make the same change in `_dotnet-test` too?

cesarbs:
Sure.

build/_k-test.shade
cesarbs: I'll check for specific negative values, `0` and `false`. All else is `true`.

victorhurdugaci: Why?

cesarbs:
Ping.

KoreBuild-dotnet/build/_dotnet-test.shade
```diff
@@ -59,9 +60,18 @@ projectFile=''
...
+                    string.Equals(KOREBUILD_TEST_SKIPMONO, "false"))
+                {
+                    // Dotnet("test" + testArgs, projectFolder);
+                }
             }
-            else if (!IsLinux || !string.IsNullOrEmpty(KOREBUILD_TEST_DNXCORE))
+            else if (!IsLinux ||
+                     (!string.IsNullOrEmpty(KOREBUILD_TEST_DNXCORE) &&
+                      !string.Equals(KOREBUILD_TEST_DNXCORE, "0") &&
+                      !string.Equals(KOREBUILD_TEST_DNXCORE, "false")))
```
dougbu: Conditions are rather odd. Where else do we consider "anything but `0` and `false`" as true? Use `string.IsNullOrEmpty()` to be consistent with existing `KOREBUILD_TEST_DNXCORE` checks. Or switch both checks to treat only `1` and `true` as true, doing part of #326.

dougbu:
:watch:

KoreBuild-dotnet/build/_dotnet-test.shade
cesarbs: I just did this in response to @victorhurdugaci saying not to just use string.IsNullOrEmpty...

cesarbs: Also, I don't think it's so odd. Handling `0` and `false` as false and all else as true is somewhat like what some programming languages do.

build/_k-test.shade
Eilon: This is getting crazy. Just check for `1` and `true` (lowercase as far as I'm concerned) and ignore everything else.

cesarbs: :+1:

Eilon: BTW the reasoning for this is: options are generally **off** by default, so we want only very particular settings to turn an option **on**. All other values do nothing.

cesarbs:
Checking for `1` or `true` per @Eilon's order :grin:

dougbu:
:shipit:

PR #331, opened 2015-12-10
move more to nupkg

victorhurdugaci:
- change `packages` dir to be `.build`
- download Sake inside KoreBuild
- move as much logic as possible in KoreBuild (now build.cmd/sh only invoke KoreBuild)

please review @pranavkm @anurse @BrennanConroy


build-template-dotnet/build.sh~
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
```
victorhurdugaci: This file shouldn't be here. I'll remove it

KoreBuild-dotnet/build/KoreBuild.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+sakeFolder=$koreBuildFolder/build/Sake
+
+if test ! -d $sakeFolder; then
+    mono $nugetPath install Sake -ExcludeVersion -o $koreBuildFolder/build -nocache -pre
```
BrennanConroy: why are the args different for this one than the .cmd

anurse:
@BrennanConroy 's comment seems relevant. Once resolved, :shipit:

KoreBuild-dotnet/build/KoreBuild.cmd
```diff
@@ -0,0 +1,31 @@
...
+	ECHO Error: KOREBUILD_FOLDER is not set.
+	EXIT /B 1
+)
+IF "%NUGET_PATH%"=="" (
+	ECHO Error: NUGET_PATH is not set.
+	EXIT /B 1
+)
+
+IF NOT EXIST Sake  (
+    "%NUGET_PATH%" install Sake -ExcludeVersion -Source https://www.nuget.org/api/v2/ -o %~dp0
```
pranavkm: Use the v3 endpoint.

KoreBuild-dotnet/build/KoreBuild.cmd
```diff
@@ -0,0 +1,31 @@
+@ECHO off
+SETLOCAL
+
+IF "%KOREBUILD_FOLDER%"=="" (
+	ECHO Error: KOREBUILD_FOLDER is not set.
```
pranavkm: Change the indent size to match build.cmd (2 spaces)

KoreBuild-dotnet/build/KoreBuild.cmd
```diff
@@ -0,0 +1,31 @@
+@ECHO off
```
pranavkm: Don't fully understand - why do we need this file? Why not just do all of this work in build.cmd (it'll have to pull down Sake, but that doesn't seem super problematic?)

pranavkm:
:watch:

KoreBuild-dotnet/build/KoreBuild.cmd
victorhurdugaci: - Sake is only required by KoreBuild in it's current form. So KoreBuild should get it. 
- We also want to install dotnet with the dotnet-install.cmd file which I think is a temporary solution. If there will be a dnvm-like tool for dotnet, I don't want to update all the build.cmd files in each repo

KoreBuild-dotnet/build/KoreBuild.sh
victorhurdugaci: Fixed

KoreBuild-dotnet/build/KoreBuild.cmd
victorhurdugaci: Done

KoreBuild-dotnet/build/KoreBuild.cmd
victorhurdugaci: build.cmd has 4 spaces too

victorhurdugaci:
Updated

victorhurdugaci:
Testing it now on linux

victorhurdugaci:
Works on linux too

KoreBuild-dotnet/build/KoreBuild.cmd
pranavkm: > Sake is only required by KoreBuild in it's current form

Are there plans to switch from Sake as a dependency? cc @Eilon

pranavkm: For the second one, wouldn't it be changing the build script inside Sake? I'd much rather have us use Sake \ C# for tweaking rather than have to manage cmd and sh files.

victorhurdugaci: We can look into replacing the cmd/sh files with sake in a future iteration

victorhurdugaci:
Closing and will reopen against release

PR #335, opened 2015-12-14
Support for dnx test (coreclr only)

victorhurdugaci:
Run tests with the new runner on CoreCLR. Run with dnx on desktop.

Please review: @BrennanConroy @anurse 

FYI @piotrpMSFT


victorhurdugaci:
Merging because I need the package. Feel free to leave comments and I'll address them

PR #336, opened 2015-12-14
fix line endings in DorkBuild

anurse:
The line-endings for the DorkBuild scripts were being set to CRLF on the CI when producing the DorkBuild package, which caused issues when using DorkBuild on non-Windows. This should ensure that the next DorkBuild build has the right line endings. It's a safe change because we use it on all of our other repos (and `.sh` **should** always use `\n` anyway)

/cc @victorhurdugaci

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,9 +4,9 @@
 * text=auto
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


anurse:
Apparently this is fixed in release.

PR #337, opened 2015-12-17
Remove dnx and make CLR tests use the xunit runner

victorhurdugaci:
- Remove dnx
- Use `xunit.runner.console` to run tests on CLR
- Cleanup a bit the files
- Remove one unnecessary file

cc @anurse @pranavkm


KoreBuild-dotnet/build/_xunit-test.shade
```diff
@@ -0,0 +1,52 @@
...
+            configs.Keys.Where(k => k.StartsWith("dnx", StringComparison.OrdinalIgnoreCase));
+            
+        var runnerExe = Path.Combine(KoreBuildFolderPath, "build", "xunit.runner.console", "tools", "xunit.console.exe");
+
+        foreach (var framework in targetFrameworks)
+        {
+            if (!framework.StartsWith("dnxcore"))
+            {
+                 var publishFolder = Path.Combine(projectFolder, ".testPublish", framework);
+                 DotnetPublish(projectFile, publishFolder, framework);
```
anurse: Actually, you can just `dotnet-compile`. Since we use the standard Desktop CLR host, `dotnet-compile` has to copy all dependencies locally anyway.

anurse: You can probably just do the normal compilation too (i.e. without `.testPublish`) since it's just a standard compile.

victorhurdugaci: Will not work because these are not executables

anurse: :+1:

Issue #338, opened 2015-12-23
KoreBuild-dotnet should support concurrent builds


victorhurdugaci:
Because it builds the entire compilation graph everytime, there's a good chance it will hit locked files. It makes our parallel build fail... Dotnet will add support for it as part of: https://github.com/dotnet/cli/issues/383

This issue is to track that we still need to add support for this. For now I'll disable it.

muratg:
So when the dotnet-compile support goes in, do we just need to re-enable concurrent build, or will we need to change code?

victorhurdugaci:
Not sure yet. Depends on how they implement the support for concurrent builds

muratg:
Cool.  Putting in RC2 for now, since the CLI change seems to be scheduled for RC2 as well.  If it goes in too late, we'll move this one to 1.0.0

victorhurdugaci:
This is happening automatically with the latest CLI. There's no work needed on our side.

PR #339, opened 2015-12-23
Remove the --parallel flag in KoreBuild-dotnet

victorhurdugaci:
Because dotnet dumps file on disk and removes them, the parallel build doesn't work anymore since files are locked. Thus, I'm removing the `--parallel` flag

FYI @pranavkm @NTaylorMullen


victorhurdugaci:
Related https://github.com/aspnet/Universe/issues/338

NTaylorMullen:
:shipit:

PR #348, opened 2016-01-12
Add test args to dotnet-test-xunit

natemcmaster:
cc @victorhurdugaci


victorhurdugaci:
:shipit:

victorhurdugaci:
But this code is not executed today :) we still use dnx for coreclr

natemcmaster:
Ok. Well hopefully it prevent future bugs.

PR #352, opened 2016-01-13
Fix #350.

### LINKED ISSUE #350: KoreBuild.cmd is setting DOTNET_HOME wrong
Need to set it without `\bin`.

cesarbs:
#350

cc @victorhurdugaci


victorhurdugaci:
Looks good. Make sure that repos can be still built with this and then :shipit:

PR #354, opened 2016-01-13
Honor KOREBUILD_TEST_SKIPMONO (#353).

cesarbs:
Nothing needs to be done for `KOREBUILD_TEST_DNXCORE` since the new build doesn't skip tests on CoreCLR on non-Windows platforms.

#353

cc @victorhurdugaci


victorhurdugaci:
Looks good :shipit:

bricelam:
@natemcmaster Can we use this to get rid of our custom shell script?

natemcmaster:
Probably. KoreBuild still requires mono be installed though. As long as we are still using dnx to test, let's keep our custom shell script. It runs *much* faster and has the same test coverage.

PR #361, opened 2016-01-28
Build tests and samples

victorhurdugaci:
Turns out we're not building tests and samples anymore

cc @pranavkm @Tratcher

### DIFF
--- a/KoreBuild-dotnet/build/_k-standard-goals.shade
+++ b/KoreBuild-dotnet/build/_k-standard-goals.shade
@@ -102,6 +102,18 @@ default NUGET_FEED = 'https://api.nuget.org/v3/index.json'
         File.Copy(nupkg, Path.Combine(BUILD_DIR, Path.GetFileName(nupkg)), true);
     }
   }
+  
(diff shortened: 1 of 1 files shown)


pranavkm:
:shipit:

Tratcher:
Would this cause a normal `build.cmd` run to compile the tests twice? Once for `build-test` and once for `test`?

pranavkm:
In theory, once we have incremental builds in CLI, it should no-op the second build.

Tratcher:
But in the meantime this makes everything even slower locally? That might be a problem.

pranavkm:
They have incremental builds working in the latest drop. We're waiting on https://github.com/dotnet/cli/pull/1067#discussion_r51060978 to go through before we switch to it. That should be soon. We could hold off on this check in until that happens.

victorhurdugaci:
@JunTaoLuo this might impact your configurable folder change

victorhurdugaci:
Here's the deal. I want to push this because the CI doesn't build samples, so I've added a switch that allows you to build "the old way" as a workaround for building locally in case you feel the build is too slow.

PR #364, opened 2016-02-02
Add resx generation back

pakrym:
@victorhurdugaci


victorhurdugaci:
What's the difference between our resx generator and the one in CLI?

pakrym:
Cli does not have resx to cs generator only .resx -> .resource, .resource -> .dll

pakrym:
:up: :date:

PR #366, opened 2016-02-03
Set DOTNET_HOME correctly.

cesarbs:
cc @victorhurdugaci


victorhurdugaci:
Ha, good catch :) :shipit:

PR #368, opened 2016-02-04
update dotnet-install script and use beta channel

anurse:
/cc @victorhurdugaci @davidfowl 

This is using the install script from https://github.com/dotnet/cli/pull/1228 optimistically :). If that gets changed before merging I'll update it here.

Both scripts default the `CHANNEL` value to `beta`. But it can be overridden by an environment variable (`CHANNEL` ;))

### DIFF
--- a/KoreBuild-dotnet/build/KoreBuild.cmd
+++ b/KoreBuild-dotnet/build/KoreBuild.cmd
@@ -9,6 +9,9 @@ IF "%NUGET_PATH%"=="" (
 	ECHO Error: NUGET_PATH is not set.
 	EXIT /B 1
 )
+IF "%KOREBUILD_DOTNET_CHANNEL%"=="" (
+    SET KOREBUILD_DOTNET_CHANNEL=beta
+)
 
 IF NOT EXIST Sake  (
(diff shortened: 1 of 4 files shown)


KoreBuild-dotnet/build/KoreBuild.cmd
```diff
@@ -9,6 +9,9 @@ IF "%NUGET_PATH%"=="" (
 	ECHO Error: NUGET_PATH is not set.
 	EXIT /B 1
 )
+IF "%CHANNEL%"=="" (
```
victorhurdugaci: Can you make this `KOREBUILD_DOTNET_CHANNEL` or something similar?

anurse: Yeah, I can. Though since the scripts read the environment variable, it won't change the fact that `CHANNEL` will work :)

victorhurdugaci: That's okay. But our build will at least use something more meaningful

anurse:
ping?

victorhurdugaci:
pong

victorhurdugaci:
:shipit:

PR #372, opened 2016-02-05
Publish full .net tests into artifacts folder

natemcmaster:
Publishing inline w/ source in the .testPublish folder causes some problems like https://github.com/aspnet/EntityFramework/pull/4477.

Also will make it easier to reuse CI artifacts


KoreBuild-dotnet/build/_xunit-test.shade
```diff
@@ -3,6 +3,7 @@ use import="Environment"
 
 default NO_PARALLEL_TEST_PROJECTS='${E("NO_PARALLEL_TEST_PROJECTS")}'
 default KOREBUILD_TEST_SKIPMONO='${E("KOREBUILD_TEST_SKIPMONO")}'
+default TEST_ARTIFACT_DIR='${Path.Combine(Directory.GetCurrentDirectory(), "artifacts", "tests")}'
```
pranavkm: This would affect any test that attempts to determine paths relative to the test publish directory. For instance https://github.com/aspnet/Localization/blob/dev/test/Microsoft.AspNetCore.Localization.FunctionalTests/LocalizationTest.cs#L16.

Could you ready follow up PRs to make sure your changes work with this?

natemcmaster: Ah good catch. I'll scan to see if I can find any similar cases.

natemcmaster:
As @pranavkm pointed out, this may affect tests that use relative file paths. Let's find all those first.

Btw, I'm leaving on vacation in an hour and will be out for the next 10 days so I won't be working on that until I get back...unless someone else wants to finish that work for me.

pranavkm:
cc @kichalla - might affect the CLI conversion.

KoreBuild-dotnet/build/_xunit-test.shade
victorhurdugaci: Why not publish in bin or obj instead?

natemcmaster: Good idea. I'll change to publish to `obj/` so we don't need to update filepaths in testing and then just copy the files to the artifacts directory. This is what we do for all other artifacts anyways.

natemcmaster:
I updated the PR. Now this PR has just 2 changes.
1. The test publish puts output into `obj/testPublish-$framework/` instead of `.testPublish/$framework/`
2. Copy the test publish artifacts to `artifacts/tests/$projectName/$framework`

natemcmaster:
Superseded by #390

PR #373, opened 2016-02-06
Increase OSX file descriptor limit during restore

natemcmaster:
Workaround https://github.com/dotnet/cli/issues/809


victorhurdugaci:
I think we have a fix for this. It might not be needed. @moozzyk ?

moozzyk:
Yes. NuGet folks committed the fix. I am not sure if cli moved to this version of NuGet yet though.

natemcmaster:
FWIW dotnet 1.0.0-beta-001491 appears to have the fix. no more osx restore crashes :)

PR #378, opened 2016-02-10
Handle --version in dotnet-install.sh.

cesarbs:
cc @pranavkm @victorhurdugaci


victorhurdugaci:
It looks okay... Does the KoreBuild.sh need to be updated too?

cesarbs:
No,  it is calling dotnet-install.sh passing --version, the problem is that it is not handled.

victorhurdugaci:
Okay, :shipit: it then

cesarbs:
I actually had to change `Korebuild.sh` to pin the dotnet version to 1244, since 1249 is not available for *nix yet.

Issue #381, opened 2016-02-11
Disable app domains for tests using custom attributes


victorhurdugaci:
cc @JunTaoLuo

natemcmaster:
Is this due to the disappearing tests on full .NET https://github.com/aspnet/Testing/issues/203? Disabling app domains doesn't always fix the problem (as in EF tests). The fix for EF is to change all tests from `dnx451` => `net451`

JunTaoLuo:
Hmm I'll try the approach mentioned above.

pranavkm:
:+1: to switching to net451. If you feel up to it, feel free to bulk update tests in all our repos to switch to `net451`.

JunTaoLuo:
All repos have been checked. Some will retain dnx451 due to dependencies on test sites (but they either do not require custom test discovery or have no issues with it) and EF's update is in PR https://github.com/aspnet/EntityFramework/pull/4509.

PR #386, opened 2016-02-16
Enable custom test discovery by disabling app domains

JunTaoLuo:
#381 
Verified:
- custom test discovery in Performance Benchmarks test will work
- xunit.runner.json will not be required 
- Security repo tests all pass

cc @victorhurdugaci


KoreBuild-dotnet/build/_xunit-test.shade
```diff
@@ -51,6 +51,9 @@ projectFile=''
                         testArgs = " -parallel none";
                     }
 
+                    // Removing app domains to allow custom test discovery
+                    testArgs += " -noappdomain";
```
pranavkm: This affects binding redirects (they would no longer work). What's the reason for this change?

JunTaoLuo: I have been trying to come up with a scenario with binding redirect that will be broken by this change and so far it seems to work. I have also checked a few repos, such as Security, to see if this change will break them.

natemcmaster: Last I checked, this causes most 99 of 109 tests to fail in ["Microsoft.EntityFrameworkCore.Commands.FunctionalTests"](https://github.com/aspnet/EntityFramework/tree/dev/test/Microsoft.EntityFrameworkCore.Commands.FunctionalTests)

pranavkm: Mvc has plenty of these. Try referencing a test which references two different versions of Newtonsoft.Json for e.g.

JunTaoLuo:
Will proceed with the alternative approach of `dnx451` => `net451`.

PR #387, opened 2016-02-17
Execute tests for net451 and/or dnx451

natemcmaster:
Also publishes to obj/ instead of .testpublish/

cc @JunTaoLuo @pranavkm

### DIFF
--- a/KoreBuild-dotnet/build/_xunit-test.shade
+++ b/KoreBuild-dotnet/build/_xunit-test.shade
@@ -34,37 +34,47 @@ projectFile=''
             var testArgs = noParallelTestProjects.Contains(projectName) ? " -parallel none" : "";
 
             var configs = project.ValueAsJsonObject("frameworks");
-            IEnumerable<string> targetFrameworks = configs == null?
-                new string[0]:
-                configs.Keys.Where(k => k.StartsWith("dnx", StringComparison.OrdinalIgnoreCase));
+            var targetFrameworks = configs == null
+                ? new string[0]
+                : configs.Keys
+                    .Where(k => k.StartsWith("dnx4", StringComparison.OrdinalIgnoreCase)
+                        || k.StartsWith("net4", StringComparison.OrdinalIgnoreCase));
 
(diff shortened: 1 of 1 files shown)


KoreBuild-dotnet/build/_xunit-test.shade
```diff
@@ -34,37 +34,47 @@ projectFile=''
...
             var runnerFolder = Path.GetFullPath(Path.Combine(KoreBuildFolderPath, "build", "xunit.runner.console", "tools"));
             var xunitCoreFolder = Path.GetFullPath(Path.Combine(KoreBuildFolderPath, "build", "xunit.core", "build", "_desktop"));
             
             foreach (var framework in targetFrameworks)
             {
-                if (!framework.StartsWith("dnxcore"))
+                if (IsLinux)
+                {
+                    // Work around issue with testing in parallel on Mono.
+                    testArgs = " -parallel none";
```
BrennanConroy: Since you're changing this code, I don't think -parallel is a switch anymore

natemcmaster: `-parallel` is still an option on xunit.console.exe. Did you mean we don't need to always specify this on Linux?

BrennanConroy: Hmm, maybe I'm thinking running `dotnet test -parallel none` doesn't work

JunTaoLuo: `dotnet test` is used for tests targeting `dnxcore50` and is used in `_dotnet-test.shade`. Tests targeting dnx451 or net451 will be run using the xunit runner which requires the switch.

JunTaoLuo:
:shipit: as we will need it before updating test TFMs. Will need to update some repos that depend on the relative paths of the tests such as https://github.com/aspnet/ServerTests/blob/dev/test/ServerComparison.FunctionalTests/Helpers.cs#L13

natemcmaster:
The relative paths shouldn’t need updating. I’m just moving the test publish folder sideways, not up or down.

JunTaoLuo:
You're right, they don't have to be updated due to *this* change. However they still need to be updated since they have not been correct since we moved away from dnx test which launched the tests in the project directory.

PR #390, opened 2016-02-19
Build tests and publish into artifacts folder

natemcmaster:
Change the `build-test` target to compile and publish tests to `artifacts/tests/$framework/$projectName`


KoreBuild-dotnet/build/_k-standard-goals.shade
```diff
@@ -111,7 +112,14 @@ default NUGET_FEED = 'https://api.nuget.org/v3/index.json'
       .Where(f => Path.GetFileName(f).Equals("project.json", StringComparison.OrdinalIgnoreCase))
       .ToList();
 
-    projectFiles.ForEach(projectFile => DotnetBuild(projectFile, Configuration));
+    projectFiles.ForEach(projectFile =>
+      {
+        var projectName = Path.GetFileName(Path.GetDirectoryName(projectFile));
+        foreach (var framework in GetFrameworks(projectFile))
+        {
+          DotnetPublish(projectFile, Path.Combine(TEST_BUILD_DIR, DefaultRid, framework, projectName), framework, Configuration);
```
pranavkm: Would dropping the rid work? I think `dotnet publish -o some-dir` should publish things to the specified diretory.

natemcmaster: Yeah, we don't *need* rid in the output path *per se*, but I added it to future-proof our artifacts for when we may build test artifacts for multiple RIDs.

pranavkm: Could we do it in when we have to do this? Would be nice not to have to deal with the regex parsing etc for now.

natemcmaster: Yeah, that's probably a more reasonable approach. I'll rip that out. Is that the only concern holding back this change?

KoreBuild-dotnet/build/_k-standard-goals.shade
```diff
@@ -111,7 +112,14 @@ default NUGET_FEED = 'https://api.nuget.org/v3/index.json'
       .Where(f => Path.GetFileName(f).Equals("project.json", StringComparison.OrdinalIgnoreCase))
       .ToList();
 
-    projectFiles.ForEach(projectFile => DotnetBuild(projectFile, Configuration));
+    projectFiles.ForEach(projectFile =>
```
pranavkm: Is the publish a perf thing? (We're already publishing as part of the xunit-test target).

natemcmaster:
The idea was to make it easy to run tests on multiple platforms w/o needing to recompile, but this has become irrelevant now due to other awesome CI stuff coming soon.

Issue #389, opened 2016-02-19
Pass --configuration parameter to dotnet-test


tushargupta51:
We are using korebuild to build our library: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet. We set Configuration property to "Release" in makefile.shade. 

This is what is happening step by step:

1. Korebuild.cmd first runs `dotnet pack` on all the projects under `src` directory, this generates nugets under bin\Release which is good
2. Then, it runs `dotnet build` on all the projects under `test` directory which generates assemblies under bin\Release which is also fine
3. Then, it runs `dotnet publish` on all the test projects for `dnx451` framework and outputs the results under obj\testPublish-dnx451 directory. This command is run with `--configuration Debug` option and xunit.console.runner runs the test against dnx451 framework
4. xunit dnx runner runs tests against dnxcore50 framework using `dotnet test` command. This is giving error: `Could not resolve coreclr path`. Reading on dotnet\cli, this error appears when project is not built before running tests. Dotnet thinks so because `dotnet test` command is run with `--configuration Debug` option by default and is trying to look for assemblies under bin\debug\dnxcore50 which do not exist. They exist under bin\Release\dnxcore50 as a result of step 2. 

Problem is step 4 and that is happening because we don't pass the --configuration parameter in `_dotnet-test.shade`. When I run `dotnet test --configuration Release` inside the test directory, it runs the tests fine. Can we please update the _dotnet-test.shade to take that parameter, the dotnet tooling expects that argument already: https://github.com/dotnet/cli/blob/rel/1.0.0/src/dotnet/commands/dotnet-test/Program.cs#L62.

This is failing our build. Is there any workaround? 

Also, dotnet-publish command should pass the configuration parameter here:
https://github.com/aspnet/Universe/blob/de449ee24985fff84389a526069aa80bc519db8b/KoreBuild-dotnet/build/_k-standard-goals.shade#L351

Eilon:
@pranavkm @victorhurdugaci any ideas on this?

victorhurdugaci:
I can take a look at this tomorrow

tushargupta51:
@victorhurdugaci Awesome! Thanks :)

PR #392, opened 2016-02-22
Pass the correct configuration to the test runners

victorhurdugaci:
Fixes https://github.com/aspnet/Universe/issues/389

cc @pranavkm @tushargupta51


pranavkm:
:shipit:

PR #397, opened 2016-02-23
Don't invoke k-clean as part of default lifecycle.

### LINKED ISSUE #382: Consider not doing invoking k-clean as part of default build target
We invoke `k-clean` as part of the default build lifecycle in k-standard-goals - https://github.com/aspnet/Universe/blob/dev/KoreBuild-dotnet/build/_k-standard-goals.shade#L90-L91. On the CLI, this affects incremental compilation (we throw away binaries that could be used for incremental compilation). Consequently running build multiple times is much slower than what it could be.

One suggestion would be to not invoke clean when `.\build` is invoked. A developer would have to call `build rebuild verify` \ `build build-clean verify` to get what we have now.

pranavkm:
Fixes #382


pranavkm:
This would require you to `build build-clean verify` if you want to do a clean build. cc @Eilon

Eilon:
Please send FYI to the whole team about this.

Eilon:
:shipit:

victorhurdugaci:
Please make this change in https://github.com/aspnet/KoreBuild too

PR #398, opened 2016-02-23
Reduce the number of directories we look for projects to build in.

pranavkm:
cc @victorhurdugaci \ @moozzyk \ @natemcmaster 

Not super sure if this is something that is correct to do. A cursory scan of our projects suggests we don't have src, test or samples nested beyond the first level. In Mvc, we have test websites that are nested. However they are built as dependencies to our functional tests so we can avoid rebuilding them separately.


moozzyk:
:ship::it:

victorhurdugaci:
Looks good :shipit:

natemcmaster:
:shipit:

victorhurdugaci:
Please make this change in https://github.com/aspnet/KoreBuild too

PR #399, opened 2016-02-23
Removing -pre as it causes test failures due to version mismatch

moozzyk:
between what the build runs vs what project references

Also removing for Sake since it is not needed


moozzyk:
/cc @victorhurdugaci @cesarbs @pranavkm 

Korebuild.cmd does not use `-pre`.

PR #401, opened 2016-02-23
Enable passing in 'parallel' parameter to dotnet test

kichalla:
@natemcmaster @victorhurdugaci @pranavkm

### DIFF
--- a/KoreBuild-dotnet/build/_dotnet-test.shade
+++ b/KoreBuild-dotnet/build/_dotnet-test.shade
@@ -18,7 +18,16 @@ default configuration = 'Debug'
 
 @{
     var projectFolder = Path.GetDirectoryName(projectFile);
-    var testArgs = "--configuration " + configuration;
-    Dotnet("test " + testArgs, projectFolder);
+    var projectName = Path.GetFileName(projectFolder);
+
+    var noParallelTestProjects = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+    if (!string.IsNullOrEmpty(NO_PARALLEL_TEST_PROJECTS))
+    {
+        noParallelTestProjects.UnionWith(NO_PARALLEL_TEST_PROJECTS.Split((char)','));
+    }
+
+    var testArgs = " --configuration " + configuration;
+    testArgs += noParallelTestProjects.Contains(projectName) || IsLinux ? " -parallel none" : "";
+    Dotnet("test" + testArgs, projectFolder);
 }
 


KoreBuild-dotnet/build/_dotnet-test.shade
```diff
@@ -18,7 +18,16 @@ default configuration = 'Debug'
 
 @{
     var projectFolder = Path.GetDirectoryName(projectFile);
```
kichalla: Following similar approach as in dnx test shade file

victorhurdugaci: Does it support parallel?

natemcmaster: Yes. Somehow the rumor started that dotnet test didn't, but this was just a bug. See https://github.com/dotnet/coreclr.xunit/pull/7

natemcmaster:
:shipit:

KoreBuild-dotnet/build/_dotnet-test.shade
victorhurdugaci: interesting. but if you run `dotnet test --help` you don't get the runner arguments

KoreBuild-dotnet/build/_dotnet-test.shade
```diff
@@ -18,7 +18,16 @@ default configuration = 'Debug'
...
-    var testArgs = "--configuration " + configuration;
-    Dotnet("test " + testArgs, projectFolder);
+    var projectName = Path.GetFileName(projectFolder);
+
+    var noParallelTestProjects = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+    if (!string.IsNullOrEmpty(NO_PARALLEL_TEST_PROJECTS))
```
dougbu: Will probably need to restore the `NO_PARALLEL_TEST_PROJECTS` setting in at least the MVC repo.

KoreBuild-dotnet/build/_dotnet-test.shade
kichalla: My guess is that `dotnet test` is a general purpose kick starter for a test runner, so you could have something like NUnit which might not support parallelism in which case this argument does not hold true.

natemcmaster: I believe dotnet test flows any unrecognized arguments to the testrunner. e.g. `--configuration Debug` is handled by `dotnet-test.exe` but `-parallel none` is passed to `dotnet-test-xunit.exe`. Order matters though: cref https://github.com/dotnet/cli/issues/1456

kichalla: > but if you run dotnet test --help you don't get the runner arguments

@victorhurdugaci Can you file a bug for this in dotnet cli?

victorhurdugaci: https://github.com/dotnet/cli/issues/1549

citizenmatt:
Does this mean that test runners are not parallel by default? I.e. if I just run `dotnet test` for an xunit test project it will run single threaded?

natemcmaster:
To see what is default, run `dotnet test -diagnostics`. It will show you the parallel setting it is using for that run.  
```
$ dotnet test -diagnostics
xUnit.net DNX Runner (64-bit osx.10.11-x64)
  Discovering: Microsoft.Data.Sqlite.Tests (method display = ClassAndMethod)
  Discovered:  Microsoft.Data.Sqlite.Tests (running 225 test cases)
  Starting:    Microsoft.Data.Sqlite.Tests (parallel test collections = on, max threads = 4)
```

victorhurdugaci:
@citizenmatt that depends on the runner implementation. We don't control it.

PR #404, opened 2016-02-24
Use Dotnet to run tests for DNXCore50

### LINKED ISSUE #400: dotnet test skipped for dnxcore50 tests
This started happening because of a recent change in _k-standard-goals.shade, specifically here:

https://github.com/aspnet/Universe/blob/dev/KoreBuild-dotnet/build/_k-standard-goals.shade#L196

The assumption here is that dotnet-test-xunit is listed in every project.json that has dnxcore50 tests. But that is not true. For example. Microsoft.IdentityModel.Logging.Tests (https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/KDev/test/Microsoft.IdentityModel.Logging.Tests/project.json) depends on Microsoft.IdentityModel.Tokens.Tests. MS.IM.Tokens.Tests has "dotnet-test-xunit" listed as a dependency but MS.IM.Logging does not since that gets pulled by MS.IM.Tokens.Tests automatically.

I think we should look for "dnxcore50" target and if no commands are specified (which means not running on dnx), we should run "dotnet-test" command.

pranavkm:
Fixes #400


pranavkm:
cc @victorhurdugaci \ @moozzyk \ @kichalla 

Assuming we've converted all our dnxcore50 tests to use `dotnet test`, this would be the right thing to do - always use dotnet test.

victorhurdugaci:
Told you we should have kept the old logic that was abstracted in each runner :p Now we ended up with this complication of conditions

:shipit:

PR #405, opened 2016-02-24
Downgrade CLI version because of https://github.com/NuGet/NuGet.Clien…

BrennanConroy:
…t/pull/289

@pranavkm @dougbu


davidfowl:
This is such a big downgrade...

BrennanConroy:
Would you like to find a version in between that doesn't have any bugs and works for all our repos?

davidfowl:
I rather force the fix to go in than regress a ton of CLI functionality.

dougbu:
Is a fix ready? What exactly will we lose, given Travis is on the floor with this CLI version or any other?

KoreBuild-dotnet/build/KoreBuild.cmd
```diff
@@ -41,6 +41,8 @@ CALL %~dp0dotnet-install.cmd -Channel %KOREBUILD_DOTNET_CHANNEL% -Version %KOREB
 
 ECHO Adding %DOTNET_LOCAL_INSTALL_FOLDER_BIN% to PATH
 SET PATH=%DOTNET_LOCAL_INSTALL_FOLDER_BIN%;%PATH%
+ECHO Setting DOTNET_HOME to %DOTNET_LOCAL_INSTALL_FOLDER%
+SET DOTNET_HOME=%DOTNET_LOCAL_INSTALL_FOLDER%
```
dougbu: Thought we only just removed this setting (perhaps because it was never read)?

davidfowl:
Travis is already on the floor, what I don't want is to go back that many versions without understanding the impact. What fixes are you regressing by this downgrade?

kichalla:
I suppose this is a fix for the `end of central directory not found`? Do we need to downgrade though? Yesterday @pranavkm found the issue to be with MyGet actually and if you point your feeds to our build shares this problem went way. This is what fixed MusicStoreTest yesterday.

Issue #347, opened 2016-01-07
Refactor build.cmd/sh to not include blocks


victorhurdugaci:
Do this after the dotnet migration

victorhurdugaci:
More details in this PR: https://github.com/aspnet/HtmlAbstractions/pull/10#discussion_r49028200

dougbu:
Poaching this because I was looking at unexpected differences between repos. This problem has been fixed in a few but not the majority.

dougbu:
edd1f7a and loads of commits in individual repos to match

PR #406, opened 2016-02-24
Update template's `build.cmd` to match best of our repo's

dougbu:
- #347
- `%KOREBUILD_VERSION%` doesn't work without this fix

Will apply accross the repos once this is approved.


victorhurdugaci:
:shipit:

dougbu:
edd1f7a

PR #408, opened 2016-02-26
Remove mac limit for open files

JunTaoLuo:
cc @victorhurdugaci


JunTaoLuo:
got a :shipit: offline so will merge immediately. Also will update KoreBuild repo.

PR #411, opened 2016-03-05
Batch build on the CI

pranavkm:


pranavkm:
cc @victorhurdugaci \ @BrennanConroy

makefile.shade
```diff
@@ -70,7 +71,67 @@ var buildTarget = "compile"
 
 #verify-all .pull .change-default-build-target-to-verify .build-all
 
-#build-and-install-all .change-default-build-target-for-coherence-build .build-all
+#ci-build
+    @{
+        var ciVolatileShare = Environment.GetEnvironmentVariable("CI_VOLATILE_SHARE");
+        var nugetExe = Environment.GetEnvironmentVariable("PUSH_NUGET_EXE");
```
victorhurdugaci: Do we still use this? Can't use just use the nuget under `.build`?

makefile.shade
```diff
@@ -70,7 +71,67 @@ var buildTarget = "compile"
...
+        var universeArtifacts = Path.Combine("artifacts", "build");
+        Directory.CreateDirectory(universeArtifacts);
+        buildTarget = Environment.GetEnvironmentVariable("KOREBUILD_BUILD_TARGETS") ?? "--quiet compile nuget-install";
+
+        foreach (var batch in BatchReporties())
+        {
+            Parallel.ForEach(batch, repoInfo =>
```
victorhurdugaci: Reporting will suck here... Can we also drop a file/repo with that repo's output so a human can read it?

makefile.shade
```diff
@@ -780,4 +841,138 @@ functions
...
+
+                        info.Dependencies.Add(dependency);
+                    }
+                }
+            }
+
+            return repositoryLookup.GroupBy(r => r.Order).OrderBy(r => r.Key);
+        }
+
+        static IEnumerable<string> GetDependencies(string projectJsonPath, HashSet<string> dependencies)
```
victorhurdugaci: You might want to include `tools` too in the search because that dependency appears only in that section

makefile.shade
```diff
@@ -780,4 +841,138 @@ functions
...
+                {
+                    var otherDir = Path.Combine(repo, otherDirs[i]);
+                    if (Directory.Exists(otherDir))
+                    {
+                        foreach (var directory in Directory.EnumerateDirectories(otherDir))
+                        {
+                            var projectJson = Path.Combine(directory, "project.json");
+                            if (File.Exists(projectJson))
+                            {
+                                GetDependencies(projectJson, info.DependencyNames);
```
victorhurdugaci: This one returns a collection of strings... What happened to the return value?

makefile.shade
```diff
@@ -780,4 +841,138 @@ functions
...
+
+                if (Directory.Exists(srcDir))
+                {
+                    foreach (var directory in Directory.EnumerateDirectories(srcDir))
+                    {
+                        info.RepositoryNames.Add(Path.GetFileName(directory));
+                        var projectJson = Path.Combine(directory, "project.json");
+                        if (File.Exists(projectJson))
+                        {
+                            GetDependencies(projectJson, info.DependencyNames);
```
victorhurdugaci: What happened to the return value?

makefile.shade
```diff
@@ -780,4 +841,138 @@ functions
...
+
+            foreach (var info in repositoryLookup)
+            {
+                foreach (var item in info.DependencyNames)
+                {
+                    var dependency = repositoryLookup.Find(r => r.RepositoryNames.Contains(item));
+                    if (dependency != null)
+                    {
+                        // Testing has odd cyclic dependencies
+                        if (string.Equals(info.Name, "Testing", StringComparison.OrdinalIgnoreCase) &&
```
victorhurdugaci: Don't hardcode these here. Write some code to detect cycles

makefile.shade
```diff
@@ -780,4 +841,138 @@ functions
...
+
+            public HashSet<string> DependencyNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+            public HashSet<RepositoryInfo> Dependencies = new HashSet<RepositoryInfo>();
+
+            public int Order
+            {
+                get
+                {
+                    if (string.Equals("Testing", Name, StringComparison.OrdinalIgnoreCase))
```
victorhurdugaci: Don't hardcode this

makefile.shade
```diff
@@ -780,4 +841,138 @@ functions
...
+            }
+
+            return dependencies;
+        }
+
+        private class RepositoryInfo
+        {
+            public string Name { get; set; }
+
+            public HashSet<string> RepositoryNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
victorhurdugaci: {get;set;}

makefile.shade
```diff
@@ -780,4 +841,138 @@ functions
 
             return reposToBuild.ToList();
         }
+
+        static IEnumerable<IGrouping<int, RepositoryInfo>> BatchReporties()
```
victorhurdugaci: Rename to `CreateBuildGraph` or something similar

makefile.shade
```diff
@@ -70,7 +71,67 @@ var buildTarget = "compile"
...
-#build-and-install-all .change-default-build-target-for-coherence-build .build-all
+#ci-build
+    @{
+        var ciVolatileShare = Environment.GetEnvironmentVariable("CI_VOLATILE_SHARE");
+        var nugetExe = Environment.GetEnvironmentVariable("PUSH_NUGET_EXE");
+        var universeArtifacts = Path.Combine("artifacts", "build");
+        Directory.CreateDirectory(universeArtifacts);
+        buildTarget = Environment.GetEnvironmentVariable("KOREBUILD_BUILD_TARGETS") ?? "--quiet compile nuget-install";
+
+        foreach (var batch in BatchReporties())
```
victorhurdugaci: How's the perf when doing all the building in parallel? Will it kill the machine?

makefile.shade
```diff
@@ -780,4 +841,138 @@ functions
...
+            {
+                get
+                {
+                    if (string.Equals("Testing", Name, StringComparison.OrdinalIgnoreCase))
+                    {
+                        // Testing has a cyclic dependency with Common, PlatformAbstractions and Logging.
+                        return 2;
+                    }
+
+                    return 1 + (Dependencies.Count == 0 ? 0 : Dependencies.Max(d => d.Order));
```
victorhurdugaci: This is confusing. Instead:

```
if (Dependencies.Count == 0) 
{
    return 1;
}
else
{
    return 1 + Dependencies.Max(d => d.Order);
}

File: makefile.shade [hunk 12]
```diff
@@ -70,7 +71,67 @@ var buildTarget = "compile"
...
+
+        foreach (var batch in BatchReporties())
+        {
+            Parallel.ForEach(batch, repoInfo =>
+            {
+                var repo = repoInfo.Name;
+                var blockName = string.Format("Building {0}", repo);
+                if (IsTeamCity)
+                {
+                    Log.Info(string.Format("##teamcity[blockOpened name='{0}']", FormatForTeamCity(blockName)));
```
victorhurdugaci: Didn't @natemcmaster fix this already to report correctly?

makefile.shade
```diff
@@ -70,7 +71,67 @@ var buildTarget = "compile"
...
+        var nugetExe = Environment.GetEnvironmentVariable("PUSH_NUGET_EXE");
+        var universeArtifacts = Path.Combine("artifacts", "build");
+        Directory.CreateDirectory(universeArtifacts);
+        buildTarget = Environment.GetEnvironmentVariable("KOREBUILD_BUILD_TARGETS") ?? "--quiet compile nuget-install";
+
+        foreach (var batch in BatchReporties())
+        {
+            Parallel.ForEach(batch, repoInfo =>
+            {
+                var repo = repoInfo.Name;
```
victorhurdugaci: Can you please print the build tree/groups before you start building? It would provide valuable information to know what and when it will be built

makefile.shade
natemcmaster: It automatically reports targets, but I believe @pranavkm  is trying to explicitly open a new block.

Btw @pranavkm, you can reuse some of that code. (this is c# 6, so you may have to adjust for mono). See https://github.com/aspnet/KoreBuild/blob/dev/build/shade/_use-ci-loggers.shade#L4

```c#
(Log as IBlockLogger)?.StartBlock(blockName);
(Log as IBlockLogger)?.EndBlock(blockName);
```

victorhurdugaci:
:watch:

makefile.shade
pranavkm: Is there a exe in that directory - I thought that was removed?

makefile.shade
pranavkm: Sure.

makefile.shade
pranavkm: Is there a sample where we use this?

makefile.shade
pranavkm: We know every repo depends on Testing and Testing depends on Common, PlatformAbstractions and Logging - how do you propose we resolve it?

makefile.shade
pranavkm: Ehh.

makefile.shade
pranavkm: No idea but it's the build machine, so presumably capping the CPU would be a good thing.

makefile.shade
pranavkm: Sure.

makefile.shade
victorhurdugaci: We still have it because `dotnet` doesn't have `add`

makefile.shade
victorhurdugaci: [Here](https://github.com/aspnet/MusicStore/blob/dev/src/MusicStore/project.json#L20)

makefile.shade
pranavkm: Fixed.

pranavkm:
Updated.

makefile.shade
```diff
@@ -780,4 +841,163 @@ functions
...
+                        {
+                            continue;
+                        }
+
+                        info.Dependencies.Add(dependency);
+                    }
+                }
+            }
+
+            return repositoryLookup.GroupBy(r => r.Order, r => r.Name).OrderBy(r => r.Key);
```
victorhurdugaci: ToList()/ToArray(), otherwise you'll reevaluate the output every time and I noticed that you do it at least twice (once to print the groups and once to build)

makefile.shade
```diff
@@ -780,4 +841,163 @@ functions
...
+            if (frameworkNodes != null)
+            {
+                foreach (var framework in frameworkNodes.Keys)
+                {
+                    dependenciesNode = frameworkNodes.ValueAsJsonObject(framework).ValueAsJsonObject("dependencies");
+                    AddKeys(dependencies, dependenciesNode);
+                }
+            }
+
+            AddKeys(dependencies, project.ValueAsJsonObject("commands"));
```
victorhurdugaci: You forgot to include "tools"

makefile.shade
```diff
@@ -780,4 +841,163 @@ functions
...
+            if (source != null)
+            {
+                foreach (var key in source.Keys)
+                {
+                    target.Add(key);
+                }
+            }
+        }
+
+        string CreateBuildWithFlowId(string repo)
```
victorhurdugaci: This seems complicated. Why not `Process.Start` and set that environment variable?

pranavkm: `exec` shade from Sake has code to redirect standard out \ standard err to Log. I'd rather not duplicate it here.

makefile.shade
pranavkm: Yup - presumably I don't need `commands` here (since they're a DNX-ism)

pranavkm:
Updated

victorhurdugaci:
:shipit:

PR #413, opened 2016-03-10
* Use shallow clone when cloning on the CI

pranavkm:
* Avoid downloading the .build directory

cc @victorhurdugaci

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -24,7 +24,7 @@ functions
 
         // Doesn't build on Mono since their contracts don't match
         string[] excludeReposOnMono = new[] { "Helios" };
-        IEnumerable<string> repositories = GetRepositoriesToBuild();
+        string[] repositories = GetRepositoriesToBuild().ToArray();
 
         static bool useHttps = UseHttps(BASE_DIR);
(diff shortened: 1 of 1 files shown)


makefile.shade
```diff
@@ -98,7 +110,15 @@ var buildTarget = "compile"
                     blockLogger.StartBlock(blockName);
                 }
 
-                Log.Info(blockName);
+                if (IsLinux)
+                {
+                    Exec("copy", "-ar .build " + Path.Combine(repo, ".build"), "");
```
dougbu: `No command 'copy' found, did you mean: ...` where none of the suggestions is the correct `cp`.

victorhurdugaci: I think sake already has a copy shade file

davidfowl: Yea, why isn;t this just using copy

pranavkm: It uses File.Copy which is surprisingly slow when you have a bunch of files.

pranavkm:
Updated

makefile.shade
dougbu: @pranavkm if you don't want to fix that in this PR, please file a follow-up to update `shade\_copy.shade` in Korebuild. We don't use the Sake script but do use `File.Copy()`.

Issue #304, opened 2015-10-19
Enable CI CoreCLR test runs on Linux and OS X 


muratg:

divega:
@cesarbs please let us know when you are doing this. We want to make sure EF tests are run an work great on Mac.

cc @natemcmaster

natemcmaster:
All EF tests are passing on OSX CI now. This required manually setting locale on our CI agents.

For future reference, due to https://github.com/dotnet/coreclr/issues/2531 I had to manually set environment variables to ensure the thread culture is "en-US", not "en-US-POSIX".

Env variables I added:

Var Name | Value
---------|---------
LANG |	 en_US.UTF-8
LC_COLLATE|	 en_US.UTF-8
LC_CTYPE	| en_US.UTF-8
LC_MESSAGES	| en_US.UTF-8
LC_MONETARY|	 en_US.UTF-8
LC_NUMERIC	| en_US.UTF-8
LC_TIME	| en_US.UTF-8

muratg:
:clap: :100:

cesarbs:
@natemcmaster Did you set them in-proc or globally on the agent?

natemcmaster:
Globally on our Mac agents in `<TeamCity Agent Home>/conf/buildagent.properties`

cesarbs:
This is finally working :smile:

natemcmaster:
Looks like our new agent `aspnetci-a20` is missing these variables.

PR #418, opened 2016-03-15
Cache Coherence locally in ci-test.

cesarbs:
Part of #304.

Still needs a bit more testing.

cc @victorhurdugaci @pranavkm

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -12,6 +12,7 @@ use namespace='System.Text'
 use namespace='System.Text.RegularExpressions'
 use namespace='System.Threading.Tasks'
 use import="BuildEnv"
+use import="Files"
 use import="Json"
 
 functions
@@ -86,7 +87,7 @@ var buildTarget = "compile"
 
 #verify-all .pull .change-default-build-target-to-verify .build-all
 
-#ci-test .pull .sync-commits .remove-src-folders .change-default-build-target-to-verify .build-all
+#ci-test .pull .sync-commits .remove-src-folders .cache-coherence .change-default-build-target-to-verify .build-all
 
 #ci-build
     @{
@@ -200,6 +201,40 @@ var buildTarget = "compile"
         foreach (var repo in GetRepositoriesToBuild())
         {
             RemoveSrcFolder(repo);
+        }
+    }
+
+#cache-coherence
+    @{
+        var coherenceCacheDir = Path.Combine(BASE_DIR, Environment.GetEnvironmentVariable("COHERENCE_CACHE_DIR") ?? ".coherence-cache");
+        var dropsShare = Environment.GetEnvironmentVariable("ASPNETCI_DROPS_SHARE");
+
+        if (!string.IsNullOrWhiteSpace(dropsShare))
+        {
+            var coherenceShare = Path.Combine(dropsShare, "Coherence", BUILD_BRANCH);
+            var latestBuild = Directory.EnumerateDirectories(coherenceShare)
+                .Select(d => Path.GetFileName(d))
+                .Where(d => { int _; return int.TryParse(d, out _); })
+                .OrderByDescending(d => d)
+                .First();
+            var targetDir = Path.Combine(coherenceCacheDir, latestBuild);
+            
+            Log.Info("Latest Coherence build is " + latestBuild     + ".");
+
+            if (!Directory.Exists(targetDir))
+            {
+                var latestBuildShare = Path.Combine(coherenceShare, latestBuild);
+
+                Log.Info("Caching Coherence build " + latestBuild + " at " + targetDir + ".");
+                CopyFolder(latestBuildShare, targetDir, overwrite: true);
+            }
+            else
+            {
+                Log.Info("Coherence build " + latestBuild + " already cached at " + targetDir + ".");
+            }
+            
+            Environment.SetEnvironmentVariable("NUGET_VOLATILE_FEED_ASPNETVNEXT", Path.Combine(targetDir, "build"));
+            Environment.SetEnvironmentVariable("NUGET_VOLATILE_FEED_EXTERNAL", Path.Combine(dropsShare, "latest-packages", "external", BUILD_BRANCH));
         }
     }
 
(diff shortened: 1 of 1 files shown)


makefile.shade
```diff
@@ -203,6 +204,32 @@ var buildTarget = "compile"
...
+    @{
+        var coherenceCacheDir = Path.Combine(BASE_DIR, Environment.GetEnvironmentVariable("COHERENCE_CACHE_DIR") ?? ".coherence-cache");
+        
+        if (!Directory.Exists(coherenceCacheDir))
+        {
+            var dropsShare = Environment.GetEnvironmentVariable("ASPNETCI_DROPS_SHARE");
+
+            if (!string.IsNullOrWhiteSpace(dropsShare))
+            {
+                var coherenceShare = Path.Combine(dropsShare, "Coherence", BUILD_BRANCH);
```
victorhurdugaci: I'm wondering if this can fail if you get a partially pushed coherence... Maybe coherence should drop a file as the final step and we should look only at the folders that have that file

makefile.shade
```diff
@@ -203,6 +204,32 @@ var buildTarget = "compile"
         }
     }
 
+#cache-coherence
+    @{
+        var coherenceCacheDir = Path.Combine(BASE_DIR, Environment.GetEnvironmentVariable("COHERENCE_CACHE_DIR") ?? ".coherence-cache");
+        
+        if (!Directory.Exists(coherenceCacheDir))
```
victorhurdugaci: should you replace if it exists?

makefile.shade
```diff
@@ -203,6 +204,32 @@ var buildTarget = "compile"
...
+            {
+                var coherenceShare = Path.Combine(dropsShare, "Coherence", BUILD_BRANCH);
+                var latestBuild = Directory.EnumerateDirectories(coherenceShare)
+                    .Select(d => Path.GetFileName(d))
+                    .Where(d => { int _; return int.TryParse(d, out _); })
+                    .OrderByDescending(d => d)
+                    .First();
+                var latestBuildShare = Path.Combine(coherenceShare, latestBuild);
+
+                CopyFolder(latestBuildShare, coherenceCacheDir, overwrite: true);
```
victorhurdugaci: Write what the latest build share is used

victorhurdugaci:
I think it's okay. :shipit:

makefile.shade
cesarbs: I could revert this. I had it like this to test faster locally. What do you think?

cesarbs:
Went with a slightly different approach: I'm keeping the Coherence build number in the cached directory. If I see there's already a cache for that build number, I don't try to copy it over again.

makefile.shade
```diff
@@ -203,6 +204,40 @@ var buildTarget = "compile"
...
+
+                Log.Info("Caching Coherence build " + latestBuild + " at " + targetDir + ".");
+                CopyFolder(latestBuildShare, targetDir, overwrite: true);
+            }
+            else
+            {
+                Log.Info("Coherence build " + latestBuild + " already cached at " + targetDir + ".");
+            }
+            
+            Environment.SetEnvironmentVariable("NUGET_VOLATILE_FEED_ASPNETVNEXT", Path.Combine(targetDir, "build"));
```
pranavkm: Caching the nupkgs as-is is not super nice for perf especially given the directory has close to 600 nupkgs. You should probably expand it.

makefile.shade
```diff
@@ -203,6 +204,40 @@ var buildTarget = "compile"
...
+            Environment.SetEnvironmentVariable("NUGET_VOLATILE_FEED_EXTERNAL", Path.Combine(dropsShare, "latest-packages", "external", BUILD_BRANCH));
```
pranavkm: Why not set this up in the CI? It's easier to remove this in case we need to not use volatile?

makefile.shade
```diff
@@ -203,6 +204,40 @@ var buildTarget = "compile"
         }
     }
 
+#cache-coherence
+    @{
+        var coherenceCacheDir = Path.Combine(BASE_DIR, Environment.GetEnvironmentVariable("COHERENCE_CACHE_DIR") ?? ".coherence-cache");
```
victorhurdugaci: I think the build dir should not be relative to the `BASE_DIR` if the `COHERENCE_CACHE_DIR` is set. On the CI we'll probably set it to a global location so that builds share it and they don't have to redownload it.

makefile.shade
victorhurdugaci: There should be a way to overwrite this with an env var but we should move away from configurations that don't have defaults in code because it makes it hard to reproduce builds locally.

Btw, where are this variable and the one above used?

makefile.shade
```diff
@@ -203,6 +204,40 @@ var buildTarget = "compile"
...
+        {
+            var coherenceShare = Path.Combine(dropsShare, "Coherence", BUILD_BRANCH);
+            var latestBuild = Directory.EnumerateDirectories(coherenceShare)
+                .Select(d => Path.GetFileName(d))
+                .Where(d => { int _; return int.TryParse(d, out _); })
+                .OrderByDescending(d => d)
+                .First();
+            var targetDir = Path.Combine(coherenceCacheDir, latestBuild);
+            
+            Log.Info("Latest Coherence build is " + latestBuild     + ".");
```
victorhurdugaci: space space space space space

makefile.shade
```diff
@@ -203,6 +204,40 @@ var buildTarget = "compile"
         }
     }
 
+#cache-coherence
+    @{
+        var coherenceCacheDir = Path.Combine(BASE_DIR, Environment.GetEnvironmentVariable("COHERENCE_CACHE_DIR") ?? ".coherence-cache");
+        var dropsShare = Environment.GetEnvironmentVariable("ASPNETCI_DROPS_SHARE");
```
victorhurdugaci: There should be a default for this in code. It's hard to run it locally otherwise because you need to search through code to figure out what vars to set

makefile.shade
cesarbs: @pranavkm We don't want those variables set unless caching Coherence was successful.

@victorhurdugaci http://aspnetci/viewType.html?buildTypeId=UniverseCoherence As we've discussed offline, we should come up with a clearer way to go about this.

makefile.shade
cesarbs: lol brain damage

Issue #423, opened 2016-03-18
Cache coherence without a build number as folder name


victorhurdugaci:
Because we're caching it in a folder with the build number, we keep using disk space and never reclaim it.

If you need to know what coherence is cached, drop a text file.

PR #426, opened 2016-03-22
Delete previous Coherence cache and expand packages (#423, #425).

cesarbs:
#423
#425

cc @victorhurdugaci @pranavkm

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -21,6 +21,7 @@ functions
         const string DefaultBuildBranch = "dev";
         const string DefaultCoherenceCacheDir = ".coherence-cache";
         const string UniverseCommitsFileName = "commits-universe";
+        const string CoherenceCacheVersionFileName = "coherence-version";
 
         private static bool Quiet { get; set; }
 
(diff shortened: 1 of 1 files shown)


makefile.shade
```diff
@@ -231,27 +232,41 @@ var buildTarget = "compile"
...
-                File.Copy(Path.Combine(coherenceShare, latestBuild, UniverseCommitsFileName), Path.Combine(targetCacheDir, UniverseCommitsFileName));
+                Log.Info("Coherence build " + latestBuild + " already cached at " + coherenceCacheDir);
             }
             else
             {
-                Log.Info("Coherence build " + latestBuild + " already cached at " + targetCacheDir);
+                if (Directory.Exists(coherenceCacheDir))
+                {
+                    Log.Info("Deleting previous Coherence cache");
+                    Directory.Delete(coherenceCacheDir, true);
```
pranavkm: Does this need retries?

pranavkm:
:shipit:

makefile.shade
cesarbs: Don't think so; if we run into issues we might add it later.

PR #429, opened 2016-03-28
Allow a 'Gate branch' which is built instead of dev if it exists

ryanbrandenburg:
Allow a 'Gate branch' which is built instead of dev if it exists. The point of this is to prevent breaking changes made for changes in CoreFX packages from blocking people until the CoreCLR build passes and makes those available.

CC @pranavkm @victorhurdugaci

### DIFF
--- a/makefile.shade
+++ b/makefile.shade
@@ -45,6 +45,9 @@ functions
         static string repositoryInclude = GetEnvironmentParameter("KOREBUILD_REPOSITORY_INCLUDE");
         static string repositoryExclude = GetEnvironmentParameter("KOREBUILD_REPOSITORY_EXCLUDE");
 
+        static string gateBranch = GetEnvironmentParameter("KOREBUILD_GATE_BRANCH");
+        private static bool UseGateBranch = !string.IsNullOrEmpty(gateBranch);
+
         // Doesn't build on Mono since their contracts don't match
         string[] repositories = GetRepositoriesToBuild().ToArray();
         string[] noSrcRepositoryExclude = GetNoSrcDeleteRepositories().ToArray();
@@ -126,13 +129,56 @@ var buildTarget = "compile"
 
         Parallel.ForEach(repositories, repo =>
         {
+            var useBuildBranch = true;
             if (Directory.Exists(repo))
             {
-                GitCommand("", string.Format("pull --ff-only --quiet --depth 1 git@github.com:aspnet/{1}.git {0}:{0}", buildBranch, repo));
+                if (UseGateBranch)
+                {
+                    try
+                    {
+                        GitCommand("", string.Format("pull --ff-only --quiet --depth 1 git@github.com:aspnet/{1}.git {0}:{0}", gateBranch, repo));
+                        useBuildBranch = false;
+                        Log.Warn(string.Format(
+                            "{0} has a branch '{1}' which is overriding use of {2}. {1} must be deleted before {2} will be used.",
+                            repo,
+                            gateBranch,
+                            buildBranch));
+                    }
+                    catch (Exception ex)
+                    {
+                    	Log.Info(string.Format("{0} doesn't exist, falling back to {1}.", gateBranch, buildBranch));
+                    }
+                }
+
+                if (useBuildBranch)
+                {
+                    GitCommand("", string.Format("pull --ff-only --quiet --depth 1 git@github.com:aspnet/{1}.git {0}:{0}", buildBranch, repo));
+                }
             }
             else
             {
-                GitCommand("", string.Format("clone --quiet --depth 1 --branch {0} git@github.com:aspnet/{1}.git", buildBranch, repo));
+                if (UseGateBranch)
+                {
+                    try
+                    {
+                        GitCommand("", string.Format("clone --quiet --depth 1 --branch {0} git@github.com:aspnet/{1}.git", gateBranch, repo));
+                        useBuildBranch = false;
+                        Log.Warn(string.Format(
+                            "{0} has a branch '{1}' which is overriding use of {2}. {1} must be deleted before {2} will be used.",
+                            repo,
+                            gateBranch,
+                            buildBranch));
+                    }
+                    catch (Exception ex)
+                    {
+                    	Log.Info(string.Format("{0} doesn't exist, falling back to {1}.", gateBranch, buildBranch));
+                    }
+                }
+
+                if (useBuildBranch)
+                {
+                    GitCommand("", string.Format("clone --quiet --depth 1 --branch {0} git@github.com:aspnet/{1}.git", buildBranch, repo));
+                }
             }
         });
 
@@ -402,7 +448,7 @@ var buildTarget = "compile"
 
                 Log.Info(string.Format("Build {0} succeeded", repo));
             }
-            catch(Exception ex)
+            catch (Exception ex)
             {
                 Log.Warn(string.Format("Build {0} failed: {1}", repo, ex.Message));
                 failed[repo] = ex;
@@ -639,7 +685,7 @@ functions
         {
             var repoUrl = gitHubUriPrefix + repo + ".git";
 
-            if(Directory.Exists(repo))
+            if (Directory.Exists(repo))
             {
                 GitPull(repoUrl, buildBranch, repo);
             }


makefile.shade
```diff
@@ -795,6 +805,19 @@ functions
...
+                        GitPull(repoUrl, GateBranch, repo);
+                        return;
+                    }
+                    catch(Exception ex)
+                    {
+                        // Probably GateBranch doesn't exist, move on.
+                    }
+                }
+            
                 GitPull(repoUrl, BUILD_BRANCH, repo);
```
victorhurdugaci: Shouldn't this be in an else branch? Otherwise it will undo your previous pull

makefile.shade
```diff
@@ -20,7 +20,9 @@ functions
         const string UniverseCommitsFileName = "commits-universe";
 
         private static bool Quiet { get; set; }
-
+        private static bool UseGateBranch { get; set; }
+        private const string GateBranch = "gateDev";
```
victorhurdugaci: Can you make this an environment variable instead? I'm worried that once we create that branch someone will forget to delete it and it might sync again to it.

victorhurdugaci: Also, if you make it an environment variable, you don't need the new build target. If the variable is set sync to it, otherwise do nothing

makefile.shade
victorhurdugaci: Nevermind, I saw the return statement

makefile.shade
ryanbrandenburg: The basic idea I had when writing this is that if you call that target it will look for a specific pre-defined branch, use it if it's available, and if it's not use the normal branch. This way we can have "special" branches for only the affected repos and fail over to dev for the rest.

I share your concern about people leaving the branch laying around, but if I understand your suggestion correctly (that people would add the environment variable to the config while they're trying it then remove it by hand after they're finished) I think it will suffer from basically the same problem, possibly worse since branches are something most people look at more often then CI environment variables.

makefile.shade
```diff
@@ -795,6 +805,19 @@ functions
 
             if(Directory.Exists(repo))
             {
+                if(UseGateBranch)
```
pranavkm: `ci-build` has its own copy of the git clone script (on line 122). You need to update that bit rather than this.

ryanbrandenburg:
:up::date:

makefile.shade
```diff
@@ -125,13 +128,44 @@ var buildTarget = "compile"
...
+            var useBuildBranch = true;
             if (Directory.Exists(repo))
             {
-                GitCommand("", string.Format("pull --ff-only --quiet --depth 1 git@github.com:aspnet/{1}.git {0}:{0}", buildBranch, repo));
+                if(UseGateBranch)
+                {
+                    try
+                    {
+                        GitCommand("", string.Format("pull --f-only --quite --depth 1 git@github.com:aspnet/{1}.git {0}:{0}", gateBranch, repo));
+                        useBuildBranch = false;
```
pranavkm: We should leave this on. It could be likely that some repos don't have the gate-branch but others do.

makefile.shade
```diff
@@ -125,13 +128,44 @@ var buildTarget = "compile"
...
+                    }catch(Exception ex)
```
pranavkm: Formatting: catch on the next line.

makefile.shade
pranavkm: Oops, it's only used on a per - repo basis. Ignore the comment.

makefile.shade
```diff
@@ -44,6 +44,9 @@ functions
         static string repositoryInclude = GetEnvironmentParameter("KOREBUILD_REPOSITORY_INCLUDE");
         static string repositoryExclude = GetEnvironmentParameter("KOREBUILD_REPOSITORY_EXCLUDE");
         
+        static string gateBranch = GetEnvironmentParameter("GATE_BRANCH");
+        private static bool UseGateBranch { get{ return !string.IsNullOrEmpty(gateBranch); } }
```
pranavkm: Formatting. (space after `get`)

pranavkm:
:shipit:

makefile.shade
victorhurdugaci: Be consistent. Either use access modifiers everywhere or nowhere

makefile.shade
```diff
@@ -125,13 +128,44 @@ var buildTarget = "compile"
...
+                    {
+                        GitCommand("", string.Format("pull --f-only --quite --depth 1 git@github.com:aspnet/{1}.git {0}:{0}", gateBranch, repo));
+                        useBuildBranch = false;
+                    }catch(Exception ex)
+                    {
+                        //Probably gateBranch doesn't exist, fallback to buildBranch
+                    }
+                }
+
+                if(useBuildBranch)
```
muratg: nit: space after if :)

makefile.shade
```diff
@@ -125,13 +128,44 @@ var buildTarget = "compile"
...
             else
             {
-                GitCommand("", string.Format("clone --quiet --depth 1 --branch {0} git@github.com:aspnet/{1}.git", buildBranch, repo));
+                if(UseGateBranch)
+                {
+                    try
+                    {
+                        GitCommand("", string.Format("clone --quite --depth 1 --branch {0} git@github.com:aspnet/{1}.git", gateBranch, repo));
+                        useBuildBranch = false;
+                    }catch(Exception ex)
```
muratg: nit: catch on the next line

makefile.shade
```diff
@@ -125,13 +128,44 @@ var buildTarget = "compile"
...
+                    {
+                        //Probably gateBranch doesn't exist, fallback to buildBranch
+                    }
+                }
+
+                if(useBuildBranch)
```
muratg: nit: space

makefile.shade
```diff
@@ -125,13 +128,44 @@ var buildTarget = "compile"
 
         Parallel.ForEach(repositories, repo =>
         {
+            var useBuildBranch = true;
             if (Directory.Exists(repo))
             {
-                GitCommand("", string.Format("pull --ff-only --quiet --depth 1 git@github.com:aspnet/{1}.git {0}:{0}", buildBranch, repo));
+                if(UseGateBranch)
```
muratg: nit: space

ryanbrandenburg:
:up: :date:

makefile.shade
```diff
@@ -126,13 +129,46 @@ var buildTarget = "compile"
...
             {
-                GitCommand("", string.Format("pull --ff-only --quiet --depth 1 git@github.com:aspnet/{1}.git {0}:{0}", buildBranch, repo));
+                if (UseGateBranch)
+                {
+                    try
+                    {
+                        GitCommand("", string.Format("pull --f-only --quite --depth 1 git@github.com:aspnet/{1}.git {0}:{0}", gateBranch, repo));
+                        useBuildBranch = false;
+                    }
+                    catch(Exception ex)
```
pranavkm: Space after catch and after the `//` on the next line.

pranavkm:
:shipit:

makefile.shade
```diff
@@ -45,6 +45,9 @@ functions
         static string repositoryInclude = GetEnvironmentParameter("KOREBUILD_REPOSITORY_INCLUDE");
         static string repositoryExclude = GetEnvironmentParameter("KOREBUILD_REPOSITORY_EXCLUDE");
         
+        static string gateBranch = GetEnvironmentParameter("GATE_BRANCH");
```
pranavkm: Could we call this `KOREBUILD_GATE_BRANCH`?

makefile.shade
```diff
@@ -126,13 +129,46 @@ var buildTarget = "compile"
...
         {
+            var useBuildBranch = true;
             if (Directory.Exists(repo))
             {
-                GitCommand("", string.Format("pull --ff-only --quiet --depth 1 git@github.com:aspnet/{1}.git {0}:{0}", buildBranch, repo));
+                if (UseGateBranch)
+                {
+                    try
+                    {
+                        GitCommand("", string.Format("pull --f-only --quite --depth 1 git@github.com:aspnet/{1}.git {0}:{0}", gateBranch, repo));
```
victorhurdugaci: Did you cut an `f` here from `--ff-only`

makefile.shade
```diff
@@ -45,6 +45,9 @@ functions
         static string repositoryInclude = GetEnvironmentParameter("KOREBUILD_REPOSITORY_INCLUDE");
         static string repositoryExclude = GetEnvironmentParameter("KOREBUILD_REPOSITORY_EXCLUDE");
         
+        static string gateBranch = GetEnvironmentParameter("KOREBUILD_GATE_BRANCH");
```
victorhurdugaci: This is not KOREBUILD, it's UNIVERSE. Rename to `UNIVERSE_GATE_BRANCH`

victorhurdugaci:
:shipit: after addressing the comments

makefile.shade
pranavkm: There's a bunch of variable env names right above this line which start with that name though?

makefile.shade
ryanbrandenburg: Good eye, I must have bumped a key while moving it.

makefile.shade
```diff
@@ -126,13 +129,46 @@ var buildTarget = "compile"
...
+                if (UseGateBranch)
+                {
+                    try
+                    {
+                        GitCommand("", string.Format("clone --quite --depth 1 --branch {0} git@github.com:aspnet/{1}.git", gateBranch, repo));
+                        useBuildBranch = false;
+                    }
+                    catch (Exception ex)
+                    {
+                        // Probably gateBranch doesn't exist, fallback to buildBranch
```
muratg: Log.Info?

ryanbrandenburg:
:up::date:

@muratg's comment gave me the idea to add some warnings which will show when a gate branch exists so that it's more visible and less likely to lay around after it's needed.

pranavkm:
Any idea what's holding this up?

ryanbrandenburg:
Universe was broken each time I looked at merging it and I didn't want to confound the problem, but it looks like we got a build this morning so I'll go ahead and merge.

PR #432, opened 2016-04-06
save the .build folder during build

victorhurdugaci:
cc @pranavkm just rebased this on release instead of dev


victorhurdugaci:
Merged

PR #439, opened 2016-04-11
Only reset repos that were cloned.

cesarbs:
`sync-commits` fails when some repos were skipped because `UNIVERSE_SKIP_NO_CREDENTIALS` was set.

cc @victorhurdugaci


victorhurdugaci:
:shipit:

PR #445, opened 2016-05-20
Include nuget-verify in default ci-build target

pranavkm:


pranavkm:
cc @cesarbs  - we can remove it from the CI configuration by declaring it here.

PR #447, opened 2016-05-31
Print better error messages if project.json fails to parse

pranavkm:


pranavkm:
cc @sebastienros

sebastienros:
:shipit:

PR #449, opened 2016-06-08
Use aspnetexternal feed instead of CI share

pranavkm:


pranavkm:
cc @cesarbs

pranavkm:
We spoke about this a while ago. Finally got around to setting this feed up.

cesarbs:
:+1:

PR #456, opened 2016-08-26
Fail gracefully when repositories have cycles

### LINKED ISSUE #451: Circular Dependency causes stack overflow in repo graph analysis
Accidentally had a StaticFiles dependency in my BasicMiddleware sample, which cause a circular dependency and eventually caused a stack overflow. There should be a check to see if there is an infinite cycle and an appropriate error message rather than just crashing.

pranavkm:
Fixes #451


pranavkm:
cc @ZestyBread

makefile.shade
```diff
@@ -929,13 +929,27 @@ functions
...
-                    if (Dependencies.Count > 0)
-                    {
-                        return 1 + Dependencies.Max(d => d.Order);
-                    }
+                    return GetOrder(new HashSet<RepositoryInfo>(), this);
+                }
+            }
 
-                    return 1;
+            private static int GetOrder(HashSet<RepositoryInfo> visited, RepositoryInfo info)
```
ZestyBread: Though it would be a perf hit, it may be better to have this as IEnumerable to know the path that caused a cyclic dependency. With a hash set, it isn't guaranteed that string.Join will be in added order, and would make the error message more useful.

makefile.shade
```diff
@@ -929,13 +929,27 @@ functions
...
+            {
+                if (!visited.Add(info))
```
ZestyBread: Does RepositoryInfo override HashCode and Equals? May be using objects native one.

pranavkm: Nah, it would do reference equality. Should be fine here since we have a fixed set of instances.

pranavkm:
:up: :date:

makefile.shade
```diff
@@ -929,13 +929,29 @@ functions
...
                 {
-                    if (Dependencies.Count > 0)
-                    {
-                        return 1 + Dependencies.Max(d => d.Order);
-                    }
+                    return GetOrder(new List<RepositoryInfo>(), this);
+                }
+            }
+
+            private static int GetOrder(List<RepositoryInfo> visited, RepositoryInfo info)
```
ZestyBread: IList?

pranavkm: Not for private APIs

ZestyBread:
:ship: :it:

PR #457, opened 2016-09-19
React to deprecation of UserSecrets repo

natemcmaster:
https://github.com/aspnet/dotnet-watch/pull/163
https://github.com/aspnet/UserSecrets/pull/107

cc @pranavkm @Eilon


pranavkm:
:shipit:

pranavkm:
Does this affect 1.0.0 release in anyway?

natemcmaster:
For the moment, we are leaving aspnet/UserSecrets so we could build 1.0.x version. If we want to delete aspnet/UserSEcrets for real, we would move the 1.0.0 source code to the "master" branches of aspnet/Configuration and aspnet/DotNetTools

PR #462, opened 2016-11-08
Switch SignalR-* -> SignalR

pranavkm:


pranavkm:
cc @davidfowl \ @Eilon 

This is for 1.1.0, right?

Eilon:
I think it's technically for 1.2.0, no? @muratg when are we shipping these?

pranavkm:
Gotcha. Would we be releasing the SignalR-* bits for 1.1.0? Presumably I shouldn't tinker around with the CI configuration until we branch

muratg:
Yes, this will be a 1.2.0 RTM.  Per my last talks with @DamianEdwards, we do not want to release bits to nuget in 1.1 timeline.

pranavkm:
Good to go now that dev's open for 1.2.0?

muratg:
@pranavkm Yup. Ideally we won't need to branch out of dev branches until 1.2 previews.

PR #467, opened 2016-12-15
Add foundation for the testing from packages, not project references

natemcmaster:
Don't remove source directories for VS 2017 projects.
When ci-test is called, set a variable so KoreBuild will pick up properties required for ci-test to work correctly.

cc @pranavkm


pranavkm:
Haven't really looked at the changes, but @Eilon \ @muratg do we want to consider not doing this? Our CI configuration is riddled with exemptions for this since not all repos work without the sources present and you have to [add code](https://github.com/aspnet/MvcPrecompilation/commit/1cb4b8161bb62641c547eba65e4596dc58091a40) to make it specifically work without the src - something we don't test locally.

It's less nice that we'd test the src at the commit and not the packages if we don't do this, but assuming we didn't have a two pass build (build packages first followed by running test), the behavior would be no different.

natemcmaster:
For the record, I'm fine with either approach. In theory there is value in testing the actual binary in the nupkg, but how often has that actually revealed a problem not caught by just using regular project references?

It's worth noting that testing from packages will be opt-in. At the moment there is no way to turn a ProjectReference into a package reference magically. Test projects would looks something like this:

```xml
<ItemGroup Condition="'$(TestFromPackages)'=='true'">
    <ProjectReference Include="..\..\src\MyLibrary.csproj" />
</ItemGroup>
<ItemGroup Condition="'$(TestFromPackages)'!='true'">
    <PackageReference Include="MyLibrary" Version="1.2.0-*" />
</ItemGroup>
```

pranavkm:
> but how often has that actually revealed a problem not caught by just using regular project references?
I don't think a whole bunch. If we were to test packages, we should rely on E2E test scenarios (MusicStore \ Entropy) rather than a repo's tests. The former is specifically designed around consuming packages while the latter isn't. 

The ProjectReference looks kinda gross 😿

natemcmaster:
I'm okay tabling test-from-packages for now.

I would like to put the other part in of this PR, though...the update to 'remove-src-folders'. KOREBUILD_NOSRC_EXCLUDE has grown to include all VS 2017 projects, plus others: "Common,EntityFramework,MusicStore,DotNetTools,RazorTooling,BrowserLink,EntityFramework.Tools,PlatformAbstractions,FileSystem"

pranavkm:
Okies - do you want to send a separate PR for it sans the package-test changes?

PR #471, opened 2017-01-20
Remove RazorTooling from UniverseCoherence.

pranavkm:
RazorTooling will not be maintained in the current format and the tool is going away for tooling RTM.


natemcmaster:
RazorTooling is still referenced from aspnet/SignalR samples. We should remove it there or pin those samples to stable numbers

pranavkm:
I missed SignalR and Entropy when removing the reference, although Entropy references the 1.0.0 package which restores from NuGet.org.

PR #472, opened 2017-01-20
Workaround issue in solution targets in MSBuild RC3

natemcmaster:
Workaround https://github.com/Microsoft/msbuild/issues/1587

Required before we upgrade to RC.3


BrennanConroy:
:shipit:

PR #475, opened 2017-02-28
Update Universe to use the MSBuild-first KoreBuild

natemcmaster:
React to https://github.com/aspnet/KoreBuild/pull/180


build/repo.targets
```diff
@@ -0,0 +1,7 @@
+<Project>
+  <Target Name="CiBuild">
```
pranavkm: `CIBuild`

PR #478, opened 2017-03-07
Update to MSBuild

pranavkm:


build/RepositoryBuild.targets
```diff
@@ -0,0 +1,69 @@
...
+      Properties="
+        PackagesDirectory=$(RepositoryArtifactsDirectory);
+        DotNetPath=$(DotNetPath);
+        NuGetFeed=$(NuGetPublishVolatileFeed);
+        APIKey=$(APIKey)" />
+
+    <Message Text="============ Done building $(RepositoryToBuild) ============" Importance="High" />
+  </Target>
+
+  <Target Name="_PinVersions">
```
pranavkm: Summary of what we're doing here: prior to building a repository:

1. We look at the restore graph spec to determine packages with floating versions
2. Look at `artifacts/build/*.nupkg and `.deps/build/*.nupkg` to determine if we have a package that has the same `{Major.Minor.Build}` number as the floated version
3. Pin the build if we found something in either of the two packages. The resulting file looks like this:

```xml
<Project>
  <ItemGroup>
    <PackageReference Update="Microsoft.AspNetCore.Http.Abstractions" Version="1.2.0-preview1-t0041a40df" Condition="'$(TargetFramework)'=='net451'" />
    <PackageReference Update="Microsoft.AspNetCore.Http.Extensions" Version="1.2.0-preview1-t0041a40df" Condition="'$(TargetFramework)'=='net451'" />
    <PackageReference Update="Microsoft.AspNetCore.WebUtilities" Version="1.2.0-preview1-t0041a40df" Condition="'$(TargetFramework)'=='net451'" />
    <PackageReference Update="Microsoft.AspNetCore.Http.Abstractions" Version="1.2.0-preview1-t0041a40df" Condition="'$(TargetFramework)'=='netstandard1.3'" />
    <PackageReference Update="Microsoft.AspNetCore.Http.Extensions" Version="1.2.0-preview1-t0041a40df" Condition="'$(TargetFramework)'=='netstandard1.3'" />
    <PackageReference Update="Microsoft.AspNetCore.WebUtilities" Version="1.2.0-preview1-t0041a40df" Condition="'$(TargetFramework)'=='netstandard1.3'" />
  </ItemGroup>
</Project>
```
4. We drop this file in the project's obj directory with the right naming so that the sdk's auto include picks it up.

This removes the need for us to juggle around with the feeds to produce a coherent build. It also works around issues like the one Kestrel hit the other day where a higher version package (that we hadn't mirrored) ended up on NuGet.

build/RepositoryBuild.targets
```diff
@@ -0,0 +1,69 @@
...
+    <PropertyGroup>
+      <BuildScriptToExecute Condition="'$(OS)'!='Windows_NT'">build.sh</BuildScriptToExecute>
+      <BuildScriptToExecute Condition="'$(OS)'=='Windows_NT'">build.cmd</BuildScriptToExecute>
+      <RepositoryArtifactsDirectory>$(BuildRepositoryRoot)artifacts\build\</RepositoryArtifactsDirectory>
+    </PropertyGroup>
+
+    <Message Text="============ Building $(RepositoryToBuild) ============" Importance="High" />
+    <Exec
+      Command="./$(BuildScriptToExecute) $(_RepositoryBuildTargets) /p:BuildNumber=$(BuildNumber) /p:Configuration=$(Configuration)"
+      IgnoreStandardErrorWarningFormat="true"
```
pranavkm: For a full build, MSBuild takes super long to collate all our warnings at the end of the build. https://github.com/aspnet/BuildTools/issues/190 might help, but there's a plethora of `warning MSB3277: Found conflicts between different versions of the same dependent assembly that could not be resolved.` that I'm not sure we are doing anything about. Turning this off causes us to lose colorization, but I think it's an acceptable cost for faster builds.

build/repo.targets
```diff
@@ -1,56 +1,195 @@
...
   </Target>
 
-  <Target Name="CITest" DependsOnTargets="SetBuildNumber">
-  <PropertyGroup>
-      <_SakeTargets>$(_SakeTargets):ci-test</_SakeTargets>
-    </PropertyGroup>
-    <MSBuild Projects="$(MSBuildProjectFullPath)"
-      Targets="Sake"
-      Properties="SakeTargets=$(_SakeTargets);RepositoryRoot=$(RepositoryRoot)" />
+  <Target Name="BuildRepositories"
```
pranavkm: This always attempts to produce a coherent build from the list of cloned repositories. Doing this includes fiddling around with repo's NuGet.configs + pinning package versions. We could add a different (default?) target that does what `.\build` used to do with sake - i.e build the repositories in sequence (ever since we broke `build install` (https://github.com/aspnet/KoreBuild/issues/17). 

That said, it might be simpler to say this repo is primarily a way for our build system to co-ordinate builds rather than for users to build all of AspNet. I'm not sure there's a lot of value in doing this. Thoughts @davidfowl ?

build/repo.targets
```diff
@@ -1,56 +1,195 @@
...
+     DependsOnTargets="_FindDotNetPath;_GenerateRestoreGraphSpecs;_GenerateBuildGraph;_UpdateNuGetConfig;_CreateRepositoriesListWithCommits">
+
+    <MSBuild
+      Projects="$(MSBuildThisFileDirectory)RepositoryBuild.targets"
+      Targets="BuildRepositories"
+      Properties="
```
pranavkm: Is there a saner way to do this?

tools/BuildGraph/DGMLFormatter.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Xml.Linq;
+
+namespace BuildGraph
+{
+    public class DGMLFormatter : GraphFormatter
```
pranavkm: Stolen from https://github.com/aspnet/BuildTrain/blob/dev/src/aspnet-coupler/Graph/DgmlGraphFormatter.cs

tools/BuildGraph/Project.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public Repository Repository { get; set; }
+
+        public IList<string> PackageReferences { get; set; } = Array.Empty<string>();
+
+        public static Project Read(string name, string path)
+        {
+            var project = new Project(name)
+            {
+                // PackageReferences = MSBuildEval.GetPackageReferences(path),
```
pranavkm: Will remove

build/RepositoryBuild.targets
```diff
@@ -0,0 +1,69 @@
...
+        NuGetFeed=$(NuGetPublishVolatileFeed);
+        APIKey=$(APIKey)" />
+
+    <Message Text="============ Done building $(RepositoryToBuild) ============" Importance="High" />
+  </Target>
+
+  <Target Name="_PinVersions">
+    <PropertyGroup>
+      <PinToolBinary>$(RepositoryRoot)tools\PinVersions\bin\$(Configuration)\netcoreapp1.1\PinVersions.dll</PinToolBinary>
+      <PinVersionArgs>$(DotNetPath) $(PinToolBinary) --graph-specs-root &quot;$(_RestoreGraphSpecsDirectory) &quot; -s &quot;$(BuildDir) &quot; &quot;$(BuildRepositoryRoot) &quot;</PinVersionArgs>
```
pranavkm: https://github.com/dotnet/sdk/issues/951 🤕

build/RepositoryBuild.targets
```diff
@@ -0,0 +1,69 @@
...
+    <MSBuild
+      Projects="@(BatchedRepository)"
+      BuildInParallel="$(BatchBuilds)"
+      Targets="_BuildRepository"
+      Properties="BuildGroup=%(BatchedRepository.BuildGroup)" />
+  </Target>
+
+  <Target Name="_BuildRepository" DependsOnTargets="_PinVersions">
+    <PropertyGroup>
+      <BuildScriptToExecute Condition="'$(OS)'!='Windows_NT'">build.sh</BuildScriptToExecute>
```
natemcmaster: Instead of MSBuild => shell script => MSBuild, have you considered going straight to using MSBuild? KoreBuild is designed to key all locations off the `RepositoryRoot` property.

```xml
<MSBuild Projects="$(MSBuildProjectFullPath)" Properties="RepositoryRoot=$(BuildRepositoryRoot)" />
```

This avoids the pre-flight restore N+1 times for building N repositories, takes advantage of using the MSBuild already loaded in memory, and potentially avoids some thrashing by letting MSBuild orchestrate the entire build top to bottom.

build/RepositoryBuild.targets
```diff
@@ -0,0 +1,69 @@
...
+      IgnoreStandardErrorWarningFormat="true"
+      WorkingDirectory="$(BuildRepositoryRoot)" />
+
+    <ItemGroup>
+      <ArtifactsToCopy Include="$(RepositoryArtifactsDirectory)*" />
+    </ItemGroup>
+
+    <Copy
+      SourceFiles="@(ArtifactsToCopy)"
+      DestinationFolder="$(BuildDir)" />
```
natemcmaster: You can reduce disk usage by overriding the `BuildDir` setting for each repo so that it copies the packages into the Universe artifact folder directly.

build/RepositoryBuild.targets
natemcmaster: Could we take this one step farther and replace the PackageReference with a ProjectReference?

build/repo.targets
```diff
@@ -1,56 +1,195 @@
...
+        Remove="@(Repository)"
+        Condition="'@(Repository)'!='@(_RepositoriesToInclude)' AND '%(Identity)'!=''" />
+    </ItemGroup>
+
+    <ItemGroup Condition="'$(KOREBUILD_REPOSITORY_EXCLUDE)'!=''">
+      <RepositoriesToExclude Include="$(KOREBUILD_REPOSITORY_EXCLUDE)" />
+      <Repository Remove="@(RepositoriesToExclude)" />
+    </ItemGroup>
+
+    <Warning Text="KOREBUILD_REPOSITORY_EXCLUDE AND KOREBUILD_REPOSITORY_INCLUDE are specified."
```
natemcmaster: Error? What's the problem here?

build/repo.targets
```diff
@@ -1,56 +1,195 @@
...
+          CloneUrl=%(Repository.CloneUrl);
+          CloneBranch=%(Repository.Branch);
+          CloneRepositoryCommit=%(Repository.Commit);
+          ShallowClone=true;
+          UseGateBranch=$(UseGateBranch)
+        </AdditionalProperties>
+      </_CloneRepository>
+    </ItemGroup>
+
+    <RemoveDir Directories="$(_CloneRepositoryRoot)" Condition="Exists('$(_CloneRepositoryRoot)')" />
```
natemcmaster: Can we not do this? When doing cross-repo work, I work inside the Universe folder. It's nice to repeatedly run universe commands without it deleting all my work.

build/repo.targets
natemcmaster: You can build up a list of properties incrementally. 
https://github.com/aspnet/KoreBuild/blob/dev/build/targets/KoreBuild.SolutionBuild.targets#L148-L151

build/repo.targets
```diff
@@ -1,56 +1,195 @@
...
 
+  <Target Name="_CreateRepositoriesListWithCommits" DependsOnTargets="_FindDotNetPath">
+    <PropertyGroup>
+      <RepositoryFileWithCommit>$(BuildDir)Repositories.props</RepositoryFileWithCommit>
+    </PropertyGroup>
+    <Copy
+      SourceFiles="$(MSBuildThisFileDirectory)Repositories.props"
+      DestinationFiles="$(RepositoryFileWithCommit)" />
+
+    <Exec
```
natemcmaster: Did you consider using MSBuild instead of a console tool to do this? IIUC what the AddCommitInfo file does, you can probably reduce this to something like:
```xml
<GetGitCommitInfo WorkingDirectory="$(_CloneRepositoryRoot)">
  <Output TaskParameter="CommitHash" PropertyName="_Hash" />
</GetGitCommitInfo>
<XmlPoke2 XmlInputPath="$(RepositoryFileWithCommit)" 
    Query="//Repository[@Include='$(RepoName)']/@Commit" Value="$(_Hash)" />

File: tools/AddCommitInfo/Program.cs [hunk 13]
```diff
@@ -0,0 +1,104 @@
...
+                var commitHash = ReadCommitHash(repositoryDirectory);
+                repositoryElement.Add(new XAttribute("Commit", commitHash));
+            }
+
+            File.WriteAllText(repositoryFile, document.Root.ToString());
+        }
+
+        private static string ReadCommitHash(string repositoryPath)
+        {
+            // Based on https://github.com/aspnet/BuildTools/blob/dev/src/Internal.AspNetCore.BuildTools.Tasks/GetGitCommitInfo.cs
```
natemcmaster: I'm against copy + pasting code from other repos. If you don't my suggestion above (implementing this tool in MSbuild instead), let's instead make this tool depend on Internal.AspNetCore.BuildTools.Tasks directly. We can abstract what GetGitCommitInfo to make it more usable if necessary.

tools/shared/DotNetMuxer.cs
```diff
@@ -0,0 +1,57 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+
+// Copy of code from https://github.com/aspnet/DotNetTools/blob/047e4a8533bbcdf22831c40dd9d9aaf0c80d1953/shared/DotNetMuxer.cs
```
natemcmaster: Ahhh! We need to put this in a common place. https://github.com/aspnet/Common/issues/173

natemcmaster:
I think this is a great step forward. I'm in favor of the work, but I think we'll get even more benefit from the MSBuild upgrade to make our build system MSbuild top-to-bottom. That means instead of 'Exec'-ing to dotnet-run, let's make those tools MSBuild tasks.

The benefits we get: 
 - Makes build components usable.
   - Example: your `BuildGraph` tool could capture the the solution-to-solution build ordering logic in a reusable, abstract way. If implemented as a task, it could be defined in BuildTools and reused in Universe, Coherence, Coherence-Patch, CoreCLR, etc.
   - Saves some time on re-compiling the build graph tool for every run of Universe. This tool won't change much.
 - consistent programming model
 - unified logging. Tasks can call `Log.LogMessage`, `Log.LogError` etc and let MSbuild handle the file logger, coloring, verbosity, etc.
 - avoiding issues with PATH and .NET Core runtime installation. Tasks compile against .NET Standard and aren't bound to a specific version of .NET Core.
    - This opens the door to using MSBuild.exe on Windows. I'm not saying we should as dotnet.exe continues to be the easiest way to ensure we have the right SDKs ([for now](https://github.com/Microsoft/msbuild/issues/1493)), but we have some projects that still only work with MSBuild.exe, like Razor. It would be nice if we could just swap to MSBuild.exe for those projects.

build/Repositories.props
```diff
@@ -0,0 +1,58 @@
...
+    <Repository Include="Scaffolding" />
+    <Repository Include="Security" />
+    <Repository Include="ServerTests" />
+    <Repository Include="Session" />
+    <Repository Include="SignalR" />
+    <Repository Include="StaticFiles" />
+    <Repository Include="Testing" />
+    <Repository Include="WebSockets" />
+  </ItemGroup>
+</Project>
```
anurse: (totally serious btw)

build/RepositoryBuild.targets
anurse: I think this should totally be the end goal. If we can do it now that'd be great, but there may be some other vestigial things in the build system that prevent it right now. If we can't do it now though, let's file an issue to track it because I think Nate's right that it'll be significantly faster.

build/RepositoryBuild.targets
pranavkm: Could we try that as a follow up? I wanted to keep the `Update` as simple as possible for now.

build/repo.targets
pranavkm: I think you want to pick one or the other. It's likely a bad configuration when both are specified. I could make this an Error

build/repo.targets
pranavkm: I can move this to Clean.

build/repo.targets
pranavkm: I was under the impression that `XmlPoke(2)` can't add new attributes.

tools/shared/DotNetMuxer.cs
pranavkm: I could reference [CLI.Utils](https://www.nuget.org/packages/Microsoft.DotNet.Cli.Utils) here instead and use their version of this type. How about that?

build/RepositoryBuild.targets
pranavkm: This might have been completely unrelated. I think I ran in to this: https://github.com/Microsoft/msbuild/issues/1808#issuecomment-284809790

build/RepositoryBuild.targets
natemcmaster: > Maybe that's an issue we can deal with when it arises.

That's a good concern. I was actually just thinking about that. I don't expect we will hit this soon. In the long run, I'm hoping we can do something like https://github.com/aspnet/KoreBuild/issues/196. In the short run, we could fall back to invoking build.cmd if we need to.

build/RepositoryBuild.targets
natemcmaster: Yeah, I should have been clearer. This is okay for now. Was just wondering if you had looked into ProjectReference. At one point I played with generating an MSBuild file in `obj/` that removed PackageReference and added ProjectReference...but I ran into lingering issues in Package vs Project precedence in ResolveAssemblyReferences.

build/repo.targets
natemcmaster: Yeah, let's make it an error. Sometimes the only way to find bad configs is to force them to fail.

build/repo.targets
natemcmaster: Awesome, thanks!

build/repo.targets
natemcmaster: When you generate "RepositoryFileWithCommit", you could add an empty Commit attribute.

tools/shared/DotNetMuxer.cs
natemcmaster: Yeah, let's do that for now.

build/RepositoryBuild.targets
```diff
@@ -0,0 +1,83 @@
...
+    <PropertyGroup>
+      <BatchBuilds Condition="$(BuildInParallel) AND '$(CompileOnly)'=='true'">true</BatchBuilds>
+      <BatchBuilds Condition="'$(BatchBuilds)'==''">false</BatchBuilds>
+    </PropertyGroup>
+
+    <MSBuild
+      Projects="@(BatchedRepository)"
+      BuildInParallel="$(BatchBuilds)"
+      Targets="_BuildRepository"
+      Properties="BuildGroup=%(BatchedRepository.BuildGroup)" />
```
dougbu: What reads the `$(BuildGroup)` property?

tools/BuildGraph/DGMLFormatter.cs
dougbu: Any C# code we need to review in detail? Or was the build train code already reviewed thoroughly?

anurse: I'd take a look at the code. BuildTrain didn't really get any eyes because it was very prototypey and then kinda just got abandoned :)

PR #480, opened 2017-03-10
Enable UnloadProjectsOnCompletion to prevent OOM

pakrym:
Possibly fixes https://github.com/aspnet/Universe/issues/481


pranavkm:
:shipit:

PR #482, opened 2017-03-10
Use Exec instead of MsBuild task

pranavkm:


natemcmaster:
cc @pakrym have we opened an issue on MSBuild to deal with the OutOfMemoryException issue?

pakrym:
https://github.com/Microsoft/msbuild/issues/1846

PR #496, opened 2017-03-21
`StopOnFirstFailure` when doing repository builds

dougbu:
- avoids MSBuild continuing past failed repositories in sequential builds
- ignored when building in parallel (unfortunately)

### DIFF
--- a/build/RepositoryBuild.targets
+++ b/build/RepositoryBuild.targets
@@ -41,6 +41,7 @@
     <MSBuild
       Projects="@(BatchedRepository)"
       BuildInParallel="$(BatchBuilds)"
+      StopOnFirstFailure="true"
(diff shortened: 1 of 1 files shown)


dougbu:
Any objections?

natemcmaster:
I might be missing some context...but wasn't the "keep going even on failure" design intentional? e.g. sometimes we want the whole thing to run, even if some tests fail. This allows us to use TeamCity to mute a test.

dougbu:
This option doesn't work when `$(BuildInParallel)` is set. So, it should have no affect on CI builds.

In any case, the CI has been failing after the first repo with failed tests for ages.

natemcmaster:
Ok.

dougbu:
dd38898

PR #503, opened 2017-04-21
Resolve warnings about duplicate imports of Microsoft.Common.targets

natemcmaster:
Resolves https://github.com/aspnet/KoreBuild/issues/216

Depends on https://github.com/aspnet/BuildTools/pull/239


natemcmaster:
When this merges, will revert https://github.com/aspnet/KoreBuild/pull/227

PR #504, opened 2017-04-24
Remove Microsoft.Extensions.PlatformAbstractions

anurse:
Part of aspnet/PlatformAbstractions#50


anurse:
I'm building Universe (incl. running tests) on a VM to verify that this is safe. It should be fine, I've already verified that when referencing `Microsoft.AspNetCore.All`, **nobody** brings in `Microsoft.Extensions.PlatformAbstractions`

PR #509, opened 2017-05-11
Remove identity service from universe

javiercn:


pranavkm:
Removed the VCS trigger from the CI so Universe builds don't kick off when you commit to IdentityService.

javiercn:
Merged

PR #513, opened 2017-06-04
Copy Korebuild directory to cloned repoistories to avoid repeated dow…

kichalla:
…nloads

Fixes: https://github.com/aspnet/Coherence-Signed/issues/546

### DIFF
--- a/build/RepositoryBuild.targets
+++ b/build/RepositoryBuild.targets
@@ -57,6 +57,15 @@
 
     <Message Text="============ Building $(RepositoryToBuild) ============" Importance="High" />
 
+    <!-- Copy Korebuild folder to individual repos to avoid downloading it again and again and also to
+          prevent from taking in newer Korebuild changes when Universe is building midway -->
+    <ItemGroup>
+      <_KorebuildItems Include="$(RepositoryRoot).build\**\*.*" />
+    </ItemGroup>
+
+    <Message Text="Copying Korebuild from Universe to repository $(BuildRepositoryRoot)"/>
+    <Copy SourceFiles="@(_KorebuildItems)" DestinationFolder="$(BuildRepositoryRoot).build\%(RecursiveDir)" SkipUnchangedFiles="true" UseHardlinksIfPossible="true" />
+
     <Exec
       Command="./$(_BuildScriptToExecute) $(BuildArguments)"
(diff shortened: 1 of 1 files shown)


build/repo.targets
```diff
@@ -106,6 +106,15 @@
       Condition="'$(CloneRepositoryCommit)'!=''" />
   </Target>
 
+  <Target Name="_CopyKorebuild" AfterTargets="_CloneRepository">
```
pranavkm: Make this part of building repositories and not clone.

build/repo.targets
```diff
@@ -106,6 +106,15 @@
...
+    <ItemGroup>
+      <_CopyItems Include="$(RepositoryRoot).build\**\*.*" />
+    </ItemGroup>
+    
+    <Message Text="Copying Korebuild from Universe to repository $(_CloneRepositoryRoot)$(CloneRepository)" />
+    <Copy SourceFiles="@(_CopyItems)" DestinationFolder="$(_CloneRepositoryRoot)$(CloneRepository)\.build\%(RecursiveDir)" />
```
pranavkm: `SkipUnchangedFiles="true" UseHardlinksIfPossible="true"`

build/repo.targets
```diff
@@ -106,19 +106,11 @@
...
-    
-    <Message Text="Copying Korebuild from Universe to repository $(_CloneRepositoryRoot)$(CloneRepository)" />
-    <Copy SourceFiles="@(_CopyItems)" DestinationFolder="$(_CloneRepositoryRoot)$(CloneRepository)\.build\%(RecursiveDir)" />
-  </Target>
-  
   <Target Name="BuildRepositories"
      DependsOnTargets="_FilterRepositories;_FindDotNetPath;_GenerateRestoreGraphSpecs;_GenerateBuildGraph;_UpdateNuGetConfig;_CreateRepositoriesListWithCommits">
 
     <PropertyGroup>
+      <_RepoDir>$(_CloneRepositoryRoot)%(Repository.Identity)</_RepoDir>
```
pranavkm: I'm not sure this works. This would basically create one property with the identity of the last repository. You might need a separate target that loops over individual projects for this.

kichalla: You are right. While testing, I took a shortcut by setting the environment variable to include a single repsository which is why it worked, but it fails when there are multiple. I moved this script to a different location now.

PR #514, opened 2017-06-05
Copy msbuild artifacts from individual repos to universe artifacts fo…

kichalla:
…lder

### DIFF
--- a/build/RepositoryBuild.targets
+++ b/build/RepositoryBuild.targets
@@ -49,7 +49,9 @@
   <Target Name="_BuildRepository" DependsOnTargets="_PinVersions">
     <PropertyGroup>
       <BuildArguments>$(_RepositoryBuildTargets) /p:BuildNumber=$(BuildNumber) /p:Configuration=$(Configuration)</BuildArguments>
-      <RepositoryArtifactsDirectory>$(BuildRepositoryRoot)artifacts\build\</RepositoryArtifactsDirectory>
+      <RepositoryArtifactsRoot>$(BuildRepositoryRoot)artifacts</RepositoryArtifactsRoot>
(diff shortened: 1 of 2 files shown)


natemcmaster:
Why not just set TeamCity to gather `.r/*/artifacts/**/*`?

kichalla:
> Why not just set TeamCity to gather .r/*/artifacts/**/*?

I wanted to do that but wouldn't it require to do this on all the nodes which build Universe or is there a better way? (like inheriting team city Project?...but probably all nodes might not have a common parent project node)

natemcmaster:
By nodes I assume you mean build configurations. Currently, most universe builds are unique, so yes you'd have to edit several configs. 

But I'd prefer that to adding a lot of disk IO required to copy what might be thousands of files up a few folders.

kichalla:
> But I'd prefer that to adding a lot of disk IO required to copy what might be thousands of files up a few folders.

I initially wanted to avoid 'copying' files as it also adds time to the build but based on the current universe msbuild log performance summary I noticed that the time taken by Copy task is insignificant compared to others. I was also trying to figure out if I could just pass in a property to individual repos so that they could write the artifacts directly to the root's artifacts folder, but the problem is the msbuild artifacts folder is being created in Korebuild scripts. 

Not clear about 'thousands of files'...we are copying 2 files from the 'msbuild' folder of all repos to the root, which would be like ~90 files I think

We could see if this change makes the perf worse and if it does could revert it.

natemcmaster:
Yeah, 'thousands of files' is a exaggeration. Still seems to me like the simpler answer is to just change teamcity settings, but I'll sign of if you change the Copy to a Move instead. At least with a Move, we don't run the risk of redundant copies in the artifacts captured by TeamCity.

kichalla:
Sure I updated to use Move and checked it in. I will keep an eye on the build time it adds
88f81f00e91c41bf3f3e5274837c56c404fbda1c

PR #519, opened 2017-06-20
Update PackagePublisher version

ryanbrandenburg:
React to aspnet/BuildTools#268.


pranavkm:
:shipit: as long as it works

PR #520, opened 2017-06-27
Port PackagePublisher to be a repo-task instead of a command-line tool

natemcmaster:
Preparing for changes coming to a KoreBuild which will drop support for pre-flight restore.

This ports the PackagePublisher console tool from https://github.com/aspnet/BuildTools/tree/dev/src/PackagePublisher into the repo task project.

Advantages:
 - task libraries are runtime-independent (e.g. we don't have to worry about which version of Microsoft.NETCore.App is installed on the machine)
 - integrated logging with MSBuild
 - changes to package publisher are tied to Universe. No more floating package versions to PackagePublisher.

Disadvanages:
 - We'll end up with duplicates of this code. (Maybe we can produce a shared sources package?).


build/tasks/PackagePublisher.cs
```diff
@@ -0,0 +1,152 @@
...
+using Microsoft.Build.Framework;
+using NuGet.Common;
+using NuGet.Packaging;
+using NuGet.Protocol;
+using NuGet.Protocol.Core.Types;
+using BuildTask = Microsoft.Build.Utilities.Task;
+
+namespace UniverseTools
+{
+    public class PackagePublisher : BuildTask, ICancelableTask
```
pranavkm: Can we push this task in to the SDK? I really don't want us to maintain 4 copies of this (CoreCLR, Coherence, UniverseCoherence, Coherence-Signed).

natemcmaster:
Yeah that seems better. This will have to wait, though. Build improvements are being punted.

PR #521, opened 2017-06-27
Change bootstrapper to install the compiled version of KoreBuild

natemcmaster:
Changes the bootstrappers to use the "compiled" KoreBuild. Part of https://github.com/aspnet/KoreBuild/issues/258

This will also override the build.cmd/build.sh in individual repos to use the build.ps1 in the Unverse repo. This is temporary until we've updated all individual repos to use the new bootstrapper.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1,66 +1,177 @@
-$ErrorActionPreference = "Stop"
+#!/usr/bin/env powershell
+#requires -version 4
 
-function DownloadWithRetry([string] $url, [string] $downloadLocation, [int] $retries)
-{
-    while($true)
-    {
-        try
-        {
-            Invoke-WebRequest $url -OutFile $downloadLocation
-            break
+<#
+.SYNOPSIS
+Build this repository
+
+.DESCRIPTION
+Downloads korebuild if required. Then builds the repository.
+
+.PARAMETER Path
+The folder to build. Defaults to the folder containing this script.
+
+.PARAMETER Channel
+The channel of KoreBuild to download. Overrides the value from the config file.
+
+.PARAMETER DotNetHome
+The directory where .NET Core tools will be stored.
+
+.PARAMETER ToolsSource
+The base url where build tools can be downloaded. Overrides the value from the config file.
+
+.PARAMETER Update
+Updates KoreBuild to the latest version even if a lock file is present.
+
+.PARAMETER ConfigFile
+The path to the configuration file that stores values. Defaults to version.xml.
+
+.PARAMETER MSBuildArgs
+Arguments to be passed to MSBuild
+
+.NOTES
+This function will create a file $PSScriptRoot/korebuild-lock.txt. This lock file can be committed to source, but does not have to be.
+When the lockfile is not present, KoreBuild will create one using latest available version from $Channel.
+
+The $ConfigFile is expected to be an XML file. It is optional, and the configuration values in it are optional as well.
+
+.EXAMPLE
+Example config file:
+```xml
+<!-- version.xml -->
+<Project>
+  <PropertyGroup>
+    <KoreBuildChannel>dev</KoreBuildChannel>
+    <KoreBuildToolsSource>https://aspnetcore.blob.core.windows.net/buildtools</KoreBuildToolsSource>
+  </PropertyGroup>
+</Project>
+```
+#>
+[CmdletBinding(PositionalBinding = $false)]
+param(
+    [string]$Path = $PSScriptRoot,
+    [Alias('c')]
+    [string]$Channel,
+    [Alias('d')]
+    [string]$DotNetHome,
+    [Alias('s')]
+    [string]$ToolsSource,
+    [Alias('u')]
+    [switch]$Update,
+    [string]$ConfigFile = (Join-Path $PSScriptRoot 'version.xml'),
+    [Parameter(ValueFromRemainingArguments = $true)]
+    [string[]]$MSBuildArgs
+)
+
+Set-StrictMode -Version 2
+$ErrorActionPreference = 'Stop'
+
+#
+# Functions
+#
+
+function Get-KoreBuild {
+
+    $lockFile = Join-Path $Path 'korebuild-lock.txt'
+
+    if (!(Test-Path $lockFile) -or $Update) {
+        Get-RemoteFile "$ToolsSource/korebuild/channels/$Channel/latest.txt" $lockFile
+    }
+
+    $version = Get-Content $lockFile | Where-Object { $_ -like 'version:*' } | Select-Object -first 1
+    if (!$version) {
+        Write-Error "Failed to parse version from $lockFile. Expected a line that begins with 'version:'"
+    }
+    $version = $version.TrimStart('version:').Trim()
+    $korebuildPath = Join-Paths $DotNetHome ('buildtools', 'korebuild', $version)
+
+    if (!(Test-Path $korebuildPath)) {
+        Write-Host -ForegroundColor Magenta "Downloading KoreBuild $version"
+        New-Item -ItemType Directory -Path $korebuildPath | Out-Null
+        $remotePath = "$ToolsSource/korebuild/artifacts/$version/korebuild.$version.zip"
+
+        try {
+            $tmpfile = Join-Path ([IO.Path]::GetTempPath()) "KoreBuild-$([guid]::NewGuid()).zip"
+            Get-RemoteFile $remotePath $tmpfile
+            if (Get-Command -Name 'Expand-Archive' -ErrorAction Ignore) {
+                # Use built-in commands where possible as they are cross-plat compatible
+                Expand-Archive -Path $tmpfile -DestinationPath $korebuildPath
+            }
+            else {
+                # Fallback to old approach for old installations of PowerShell
+                Add-Type -AssemblyName System.IO.Compression.FileSystem
+                [System.IO.Compression.ZipFile]::ExtractToDirectory($tmpfile, $korebuildPath)
+            }
         }
-        catch
-        {
-            $exceptionMessage = $_.Exception.Message
-            Write-Host "Failed to download '$url': $exceptionMessage"
-            if ($retries -gt 0) {
-                $retries--
-                Write-Host "Waiting 10 seconds before retrying. Retries left: $retries"
-                Start-Sleep -Seconds 10
-
-            }
-            else
-            {
-                $exception = $_.Exception
-                throw $exception
-            }
+        catch {
+            Remove-Item -Recurse -Force $korebuildPath -ErrorAction Ignore
+            throw
+        }
+        finally {
+            Remove-Item $tmpfile -ErrorAction Ignore
         }
     }
+
+    return $korebuildPath
 }
 
-cd $PSScriptRoot
-
-$repoFolder = $PSScriptRoot
-$env:REPO_FOLDER = $repoFolder
-$koreBuildZip="https://github.com/aspnet/KoreBuild/archive/dev.zip"
-if ($env:KOREBUILD_ZIP)
-{
-    $koreBuildZip=$env:KOREBUILD_ZIP
+function Join-Paths([string]$path, [string[]]$childPaths) {
+    $childPaths | ForEach-Object { $path = Join-Path $path $_ }
+    return $path
 }
 
-$buildFolder = ".build"
-$buildFile="$buildFolder\KoreBuild.ps1"
-
-if (!(Test-Path $buildFolder)) {
-    Write-Host "Downloading KoreBuild from $koreBuildZip"
-
-    $tempFolder=$env:TEMP + "\KoreBuild-" + [guid]::NewGuid()
-    New-Item -Path "$tempFolder" -Type directory | Out-Null
-
-    $localZipFile="$tempFolder\korebuild.zip"
-
-    DownloadWithRetry -url $koreBuildZip -downloadLocation $localZipFile -retries 6
-
-    Add-Type -AssemblyName System.IO.Compression.FileSystem
-    [System.IO.Compression.ZipFile]::ExtractToDirectory($localZipFile, $tempFolder)
-
-    New-Item -Path "$buildFolder" -Type directory | Out-Null
-    copy-item "$tempFolder\**\build\*" $buildFolder -Recurse
-
-    # Cleanup
-    if (Test-Path $tempFolder) {
-        Remove-Item -Recurse -Force $tempFolder
+function Get-RemoteFile([string]$RemotePath, [string]$LocalPath) {
+    if ($RemotePath -notlike 'http*') {
+        Copy-Item $RemotePath $LocalPath
+        return
     }
+
+    $retries = 10
+    while ($retries -gt 0) {
+        $retries -= 1
+        try {
+            Invoke-WebRequest -UseBasicParsing -Uri $RemotePath -OutFile $LocalPath
+            return
+        }
+        catch {
+            Write-Verbose "Request failed. $retries retries remaining"
+        }
+    }
+
+    Write-Error "Download failed: '$RemotePath'."
 }
 
-&"$buildFile" @args
+#
+# Main
+#
+
+# Load configuration or set defaults
+
+if (Test-Path $ConfigFile) {
+    [xml] $config = Get-Content $ConfigFile
+    if (!($Channel)) { [string] $Channel = Select-Xml -Xml $config -XPath '/Project/PropertyGroup/KoreBuildChannel' }
+    if (!($ToolsSource)) { [string] $ToolsSource = Select-Xml -Xml $config -XPath '/Project/PropertyGroup/KoreBuildToolsSource' }
+}
+
+if (!$DotNetHome) {
+    $DotNetHome = if ($env:DOTNET_HOME) { $env:DOTNET_HOME } `
+        elseif ($env:USERPROFILE) { Join-Path $env:USERPROFILE '.dotnet'} `
+        elseif ($env:HOME) {Join-Path $env:HOME '.dotnet'}`
+        else { Join-Path $PSScriptRoot '.dotnet'}
+}
+
+if (!$Channel) { $Channel = 'dev' }
+if (!$ToolsSource) { $ToolsSource = 'https://aspnetcore.blob.core.windows.net/buildtools' }
+
+# Execute
+
+$korebuildPath = Get-KoreBuild
+Import-Module -Force -Scope Local (Join-Path $korebuildPath 'KoreBuild.psd1')
+
+try {
+    Install-Tools $ToolsSource $DotNetHome
+    Invoke-RepositoryBuild $Path @MSBuildArgs
+}
+finally {
+    Remove-Module 'KoreBuild' -ErrorAction Ignore
+}
--- a/build.sh
+++ b/build.sh
@@ -1,46 +1,195 @@
 #!/usr/bin/env bash
-repoFolder="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-cd $repoFolder
 
-koreBuildZip="https://github.com/aspnet/KoreBuild/archive/dev.zip"
-if [ ! -z $KOREBUILD_ZIP ]; then
-    koreBuildZip=$KOREBUILD_ZIP
-fi
+set -euo pipefail
 
-buildFolder=".build"
-buildFile="$buildFolder/KoreBuild.sh"
+#
+# variables
+#
 
-if test ! -d $buildFolder; then
-    echo "Downloading KoreBuild from $koreBuildZip"
+RESET="\033[0m"
+RED="\033[0;31m"
+MAGENTA="\033[0;95m"
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+[ -z "${DOTNET_HOME:-}"] && DOTNET_HOME="$HOME/.dotnet"
+config_file="$DIR/version.xml"
+verbose=false
+update=false
+repo_path="$DIR"
+channel=''
+tools_source=''
 
-    tempFolder="/tmp/KoreBuild-$(uuidgen)"
-    mkdir $tempFolder
+#
+# Functions
+#
+__usage() {
+    echo "Usage: $(basename ${BASH_SOURCE[0]}) [options] [[--] <MSBUILD_ARG>...]"
+    echo ""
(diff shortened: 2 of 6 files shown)


build.ps1
```diff
@@ -1,66 +1,140 @@
...
+    $version = Get-Content $lockFile -Tail 1
+    $korebuildPath = __join_paths $DotNetHome ('buildtools', 'korebuild', $version)
+
+    if (!(Test-Path $korebuildPath)) {
+        Write-Host "Downloading KoreBuild $version"
+        New-Item -ItemType Directory -Path $korebuildPath | Out-Null
+        $remotePath = "$ToolsSource/korebuild/artifacts/$version/korebuild.$version.zip"
+
+        try {
+            if ($PSVersionTable.PSVersion -ge '5.0.0.0') {
```
pakrym: Get-Command -Name "Expand-Archive" -ErrorAction SilentlyContinue

build.ps1
```diff
@@ -1,66 +1,140 @@
...
+.NOTES
+This function will create a file $PSScriptRoot/korebuild-lock.txt. This lock file can be committed to source, but does not have to be.
+When the lockfile is not present, KoreBuild will create one using latest available version from $Channel.
 
-    Add-Type -AssemblyName System.IO.Compression.FileSystem
-    [System.IO.Compression.ZipFile]::ExtractToDirectory($localZipFile, $tempFolder)
+#>
+[CmdletBinding(PositionalBinding = $false)]
+param(
+    [Alias('p')]
```
pakrym: Would this conflict with `/p:Property=Value` msbuild argument?

build.ps1
```diff
@@ -1,66 +1,140 @@
...
+
+function __get_korebuild {
+
+    $lockFile = Join-Path $Path 'korebuild-lock.txt'
+
+    if (!(Test-Path $lockFile) -or $Update) {
+        __fetch "$ToolsSource/korebuild/channels/$Channel/latest.txt" $lockFile
     }
+
+    $version = Get-Content $lockFile -Tail 1
```
pakrym: Do we need a way to pass it in or make it fixed in the repo?

build.ps1
```diff
@@ -1,66 +1,140 @@
...
+$ErrorActionPreference = 'Stop'
 
-    # Cleanup
-    if (Test-Path $tempFolder) {
-        Remove-Item -Recurse -Force $tempFolder
+#
+# Functions
+#
+
+function __get_korebuild {
```
pakrym: nit: Does not look like a powershell naming scheme

build.ps1
natemcmaster: We intend to commit this lockfile to source when we branch or need to pin for other reasons. Would a `$KoreBuildVersion` flag also be useful?

build.ps1
natemcmaster: No. I've verified this works: `build.ps1 -p ../OtherRepo /p:Prop=value`

build.ps1
pakrym: We can always add it later.

pakrym:
Poweshell LGTM, @anurse @halter73 please take a look too if you have time.

natemcmaster:
Also, if this wasn't clear, these bootstrapper scripts is what I will end up coping into all individual repos.

build.ps1
```diff
@@ -1,66 +1,140 @@
...
+
+    $lockFile = Join-Path $Path 'korebuild-lock.txt'
+
+    if (!(Test-Path $lockFile) -or $Update) {
+        __fetch "$ToolsSource/korebuild/channels/$Channel/latest.txt" $lockFile
     }
+
+    $version = Get-Content $lockFile -Tail 1
+    $korebuildPath = __join_paths $DotNetHome ('buildtools', 'korebuild', $version)
```
pakrym: Why isn't channel included here? Is it included in version?

natemcmaster: Channel is only used to fetch a specific version.

Example:  https://aspnetcore.blob.core.windows.net/buildtools/korebuild/channels/dev/latest.txt

build.sh
```diff
@@ -1,46 +1,141 @@
...
 
-    # Cleanup
-    if test -d $tempFolder; then
-        rm -rf $tempFolder
+    failed=false
+    if __machine_has curl ; then
+        curl --retry 10 -sSL -f --create-dirs -o $local_path $remote_path || failed=true
+    elif __machine_has wget; then
+        wget --tries 10 -O $local_path $remote_path || failed=true
+    else
```
anurse: Check these at the top of the file (before arg parsing). You can just duplicate the check a little (`if ! __machine_has curl && ! __machine_has wget; then ...`, double check the precedence there though I'm not positive what order those are invoked ;))

build.sh
```diff
@@ -1,46 +1,141 @@
...
+            return 1
         fi
-        retries=$((retries - 1))
-        echo "Waiting 10 seconds before retrying. Retries left: $retries"
-        sleep 10s
-    done
 
-    unzip -q -d $tempFolder $localZipFile
+        tmpfile="$(mktemp)"
+        echo "Downloading KoreBuild $version"
```
anurse: Colors maybe? You can use something like https://gist.github.com/elucify/c7ccfee9f13b42f11f81 (there are many others out there if you search for 'bash color variables' :)

build.sh
```diff
@@ -1,46 +1,141 @@
...
+        __fetch $remote_path $tmpfile
+        unzip -q -d "$korebuild_path" $tmpfile
+        rm $tmpfile || :
```
anurse: What's the `:` do here? I'm not familiar.

anurse: Prevent `pipefail`?

build.sh
```diff
@@ -1,46 +1,141 @@
...
+#
+# main
+#
+
+while [[ $# > 0 ]]; do
+    case $1 in
+        -\?|-h|--help)
+            __usage
+            ;;
+        -c|--channel|-Channel)
```
anurse: `-Channel` (and similar) are for equivalent syntax to PowerShell? Comment please :)

build.sh
```diff
@@ -1,46 +1,141 @@
...
+}
+
+#
+# main
+#
+
+while [[ $# > 0 ]]; do
+    case $1 in
+        -\?|-h|--help)
+            __usage
```
anurse: We should exit with `0` exit code in this case. `__usage` should take a flag or variable for the exit code.

build.sh
```diff
@@ -1,46 +1,141 @@
...
+        *)
+            break
+            ;;
+    esac
+    shift
+done
+
+get_korebuild
+install_tools "$tools_source" "$DOTNET_HOME"
+invoke_repository_build "$repo_path" $@
```
anurse: same comment as in PowerShell, can we reduce the "public API" surface here to just one function?

build/RepositoryBuild.proj
```diff
@@ -60,20 +60,14 @@
...
-          prevent from taking in newer Korebuild changes when Universe is building midway -->
-    <ItemGroup>
-      <_KorebuildItems Include="$(RepositoryRoot).build\**\*.*" />
-    </ItemGroup>
-
-    <Message Text="Copying Korebuild from Universe to repository $(BuildRepositoryRoot)"/>
-    <Copy SourceFiles="@(_KorebuildItems)" DestinationFolder="$(BuildRepositoryRoot).build\%(RecursiveDir)" SkipUnchangedFiles="true" UseHardlinksIfPossible="true" />
+    <!-- Copy Korebuild lock file to individual repos to align version if the repo doesn't already have one -->
+    <Message Text="Copying KoreBuild lockfile from Universe to repository $(BuildRepositoryRoot)"/>
+    <Copy SourceFiles="$(RepositoryRoot)korebuild-lock.txt" DestinationFolder="$(BuildRepositoryRoot)" />
```
anurse: Does KoreBuild write it's version number to the `artifacts` directory? If not, perhaps we should do that too in order to make sure we know what version built a particular set of artifacts.

build.ps1
```diff
@@ -1,66 +1,140 @@
...
+
+function Get-KoreBuild {
+
+    $lockFile = Join-Path $Path 'korebuild-lock.txt'
+
+    if (!(Test-Path $lockFile) -or $Update) {
+        Get-RemoteFile "$ToolsSource/korebuild/channels/$Channel/latest.txt" $lockFile
     }
+
+    $version = Get-Content $lockFile -Tail 1
```
anurse: I do think a version option would be a good idea. It seems trivial to add here but if it's more than just adding the switch and a couple lines, then it can wait.

build.ps1
```diff
@@ -1,66 +1,140 @@
...
+    $korebuildPath = Join-Paths $DotNetHome ('buildtools', 'korebuild', $version)
+
+    if (!(Test-Path $korebuildPath)) {
```
anurse: Can we add a switch to force re-download of the version even if it exists? -ResetKoreBuild?

build.ps1
```diff
@@ -1,66 +1,140 @@
...
+        Write-Host "Downloading KoreBuild $version"
```
anurse: Adding a little color to the KoreBuild messages helps them stand out from the MSBuild messages. It might be a nice addition here.

build.ps1
```diff
@@ -1,66 +1,140 @@
...
+        New-Item -ItemType Directory -Path $korebuildPath | Out-Null
+        $remotePath = "$ToolsSource/korebuild/artifacts/$version/korebuild.$version.zip"
+
+        try {
+            if (Get-Command -Name 'Expand-Archive' -ErrorAction Ignore) {
+                # Use built-in commands where possible as they are cross-plat compatible
+                $tmpfile = "$(New-TemporaryFile).zip"
```
anurse: [This calls `Path.GetTempFileName`](https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs#L26) which creates the file it returns in advance. If `Expand-Archive` (and `ZipFile` below) don't care about the extension, we could just use it without creating extra files (though granted, they are 0-byte).

Alternatively, the way you create the temp file name in the other branch is totally safe, so there's no reason not to just do it here.

Wish we had `Path.GetTempFileName(string extension)`.

build.ps1
```diff
@@ -1,66 +1,140 @@
...
+        finally {
+            remove-item $tmpfile -ErrorAction Ignore
+        }
+    }
+
+    return $korebuildPath
 }
 
-&"$buildFile" @args
+function Join-Paths($path, $childPaths) {
```
anurse: I hate that this isn't standard... Put type annotations on the params though ;)

build.ps1
```diff
@@ -1,66 +1,140 @@
...
+
+#
+# Main
+#
+
+try {
+    $korebuildPath = Get-KoreBuild
+    Import-Module -Force -Scope Local (Join-Path $korebuildPath 'KoreBuild.psd1')
+    Install-Tools $ToolsSource $DotNetHome
+    Invoke-RepositoryBuild $Path @MSBuildArgs
```
anurse: Can these steps be collapsed? Changing the template is a pain and involves changing every repo, I'd like the build script to be as rock-solid and flexible as possible and just delegate everything to a script inside KoreBuild as soon as possible to shift logic to the place we can centrally-update.

build.sh
```diff
@@ -1,46 +1,141 @@
...
-        echo "Waiting 10 seconds before retrying. Retries left: $retries"
-        sleep 10s
-    done
 
-    unzip -q -d $tempFolder $localZipFile
+        tmpfile="$(mktemp)"
+        echo "Downloading KoreBuild $version"
+        __get_remote_file $remote_path $tmpfile
+        unzip -q -d "$korebuild_path" $tmpfile
+        rm $tmpfile || :
```
anurse: What's the `:` do here? I'm not familiar with it. Does it prevent pipefail from triggering? Comment please (since we are mostly `sh`noobs)

build.ps1
natemcmaster: Can you clarify why you mean by redownload? There are two redownloads possible: (1) the download that fetches the `korebuild-lock.txt` file from the channel, and (2) the re-download of the korebuild artifact. The `-Update` addresses (1) -- it overwrite the local version of korebuild-lock.txt. 
(2) shouldn't be necessary. Once a version is determined from the lockfile, the appropriate korebuild artifact will always be downloaded if it is not already present in `$DOTNET_HOME/buildtools/korebuild/$version`.

build.ps1
natemcmaster: It's more than a few lines. We'd need to make the expected lockfile an artifact available by version too. The lockfile includes more than just the version. Currently it includes the commithash of aspnet/BuildTools which will be important for pinning.

build.ps1
natemcmaster: Regrettable, `Expand-Archive` does care. It has to be .zip :-( I just use the alternative creating of the temp file in both places.

build.sh
natemcmaster: Yes, these are meant to align with powershell. Inspiration: dotnet-install.ps1/sh does the same thing. Saves me now and then when muscle memory does the wrong thing.

build/RepositoryBuild.proj
natemcmaster: korebuild.$version.zip includes a `.version` file inside it that is identical to the `korebuild-lock.txt` file. 

See https://github.com/aspnet/BuildTools/tree/dev/sdk/KoreBuild#layout

build.sh
natemcmaster: `:` is a builtin equivalent to `true`. It makes this line is functionally equivalent to `rm $tmpfile >/dev/null 2>&1`, except that we'll get some console output indicating there was a problem.

Try this:
```sh
rm some_bogus_file || :
echo $?
```
It should return 0, but still show you that some_bogus_file did not exist.


FYI - https://unix.stackexchange.com/questions/31673/what-purpose-does-the-colon-builtin-serve

build.ps1
natemcmaster: This was very deliberate. Let me explain my thinking:
 - make it clearer to those reading the script which steps will happen when build.ps1 executes.
 - `Install-Tools` and `Invoke-RepositoryBuild` should be treated like an "API". Any changes to arguments or behavior is 'breaking' and should not be done lightly. These will rarely if ever change, so our bootstrappers can delegate the majority of stuff to these two functions.
 - I'm totally okay with one extra line in build.ps1.

FYI - I'm also thinking forward to what we discussed in our runtime-engineering meeting last week: https://github.com/aspnet/specs/pull/77. One of the problems we have to day is that our build is a mix of console tools, tasks, and scripts, scattered across nuget packages, source code, and stuff that gets downloaded on the fly. We want KoreBuild capture the "state" of build tools. Decomposing KoreBuild into a module (KoreBuild.ps**m**1) gives us more flexibility to write scripts that re-use tools from KoreBuild. `Korebuild.ps1` was too inflexible. For example, I want to add a function to KoreBuild.psd1 to `Push-Packages`. This could delegate to `nuget.exe`, `dotnet nuget`, or `PackagePublisher.dll`, but it doesn't matter to the caller since all they need is the powershell function.

build.sh
natemcmaster: Updated

build.sh
natemcmaster: updated

natemcmaster:
Improvements to build infrastructure are being delayed until after we branch for 2.0.0. I've captured this feedback and ported it back to aspnet/BuildTools: https://github.com/aspnet/BuildTools/commit/d5fd2ee2fc23117ed0978761434d23e3d44708c2

Closing for now. We can re-open this later this summer.

natemcmaster:
:up: :date: - re opening now that we have branched for 2.1.0. Since last review, I've update the bootstrappers a little.

Another important change: because this version of KoreBuild does not do pre-flight restore, any "PackageReference" in repo.targets will not be restored. This means PackagePublisher is not available from the default KoreBuild lifecycle. Instead, I've added a new file `push.cmd` that will push the packages after build is done.

You can try a dry-run push locally by running:
```
push.cmd -WhatIf
```

anurse:
Is there a reason why `build.ps1` was renamed to `run.ps1`?

pranavkm:
>  I've added a new file push.cmd that will push the packages after build is done.

Where does it push to? Why do we need something in addition to the package publisher we already have?

natemcmaster:
> Another important change: because this version of KoreBuild does not do pre-flight restore, any "PackageReference" in repo.targets will not be restored. This means PackagePublisher is not available from the default KoreBuild lifecycle. Instead, I've added a new file push.cmd that will push the packages after build is done.

build/RepositoryBuild.targets
```diff
@@ -71,13 +65,6 @@
        SourceFiles="@(RepositoryMSBuildArtifacts)"
        DestinationFolder="$(ArtifactsDir)msbuild\$(RepositoryToBuild)\%(RecursiveDir)" />
 
-    <Message Text="Publishing the following packages to the volatile feed: @(RepositoryNupkgs -> '%(Filename)%(Extension)', ', ')"
-      Condition="'$(PublishPackages)'=='true' AND '@(RepositoryNupkgs)' != ''" />
-
-    <Exec
```
pranavkm: What happened to this?

run.ps1
```diff
@@ -0,0 +1,216 @@
...
+$korebuildPath = Get-KoreBuild
+Import-Module -Force -Scope Local (Join-Path $korebuildPath 'KoreBuild.psd1')
+
+try {
+    Install-Tools $ToolsSource $DotNetHome
+
+    if ($CommandName -eq 'build') {
+        Invoke-RepositoryBuild $Path @MSBuildArgs
+    }
+    elseif ($CommandName -eq 'push') {
```
natemcmaster: Look here.

pranavkm:
Wasn't the plan to make it a task?

natemcmaster:
It was until I realized that it would introduce binary compatibility issues due to how MSBuild loads tasks. PackagePublisher currently uses NuGet 4.0.0. When I made PackagePublisher a task, it fails to load because the SDK bundles NuGet 4.3.0 which is binary incompatible. Rather than tangle with MSBuild, we can push packages from console/shell only. This is what the `Push-NuGetPackage` cmdlet does under the covers.

natemcmaster:
FYI - an alternative I discussed with @pakrym this morning was that instead of push.cmd, we leave the bootstrapper the same and add `/t:PushPackages` instead. We can bundle PackagePublisher in KoreBuild and add a KoreBuild target that pushes all \*.nupkg from $(BuildDir)

natemcmaster:
Scratch all that. Changed my mind. We have enough dependencies on NuGet anyways (like NuGetPackageVerifier), so we're unlikely to get much value from the powershell-only approach. Returning this PR to the original approach. Namely:

 -  PackagePublisher is now a built-in task in KoreBuild (https://github.com/aspnet/BuildTools/pull/306)
 - This PR uses the standard bootstrapper build.sh and build.ps1 scripts as found here: https://github.com/aspnet/BuildTools/tree/dev/scripts
 - Added a push.cmd file, which is just a short cut for `build.ps1 /t:PushPackages`

natemcmaster:
:bell: I've squashed and rebased this.

pranavkm:
Not sure about others, but I'm a bit busy with 2.0 stuff today so unlikely I'd have a chance to look at this. Would be good to have @anurse \ @pakrym have a look at this before this goes in.

build.ps1
```diff
@@ -1,66 +1,177 @@
...
-                $exception = $_.Exception
-                throw $exception
+            else {
+                # Fallback to old approach for old installations of PowerShell
+                Add-Type -AssemblyName System.IO.Compression.FileSystem
+                [System.IO.Compression.ZipFile]::ExtractToDirectory($tmpfile, $korebuildPath)
             }
         }
+        catch {
+            remove-item -Recurse -Force $korebuildPath -ErrorAction Ignore
```
pranavkm: Casing consistency `Remove-Item` (in the finally block too)

build.ps1
```diff
@@ -1,66 +1,177 @@
...
+    [string]$Path = $PSScriptRoot,
+    [Alias('c')]
+    [string]$Channel,
+    [Alias('d')]
+    [string]$DotNetHome,
+    [Alias('s')]
+    [string]$ToolsSource,
+    [Alias('u')]
+    [switch]$Update,
+    [string]$ConfigFile = (Join-Path $PSScriptRoot 'version.props'),
```
pranavkm: Can we rename the file to `xml` across repos if we intend to treat it as xml and not an msbuild evaluated file?

build.ps1
```diff
@@ -1,66 +1,177 @@
...
+if (!$DotNetHome) {
+    $DotNetHome = if ($env:DOTNET_HOME) { $env:DOTNET_HOME } `
+        elseif ($env:USERPROFILE) { Join-Path $env:USERPROFILE '.dotnet'} `
+        elseif ($env:HOME) {Join-Path $env:HOME '.dotnet'}`
+        else { Join-Path $PSScriptRoot '.dotnet'}
+}
 
-    DownloadWithRetry -url $koreBuildZip -downloadLocation $localZipFile -retries 6
+if (!$Channel) { $Channel = 'dev' }
+if (!$ToolsSource) { $ToolsSource = 'https://aspnetcore.blob.core.windows.net/buildtools' }
```
pranavkm: Why not set these inline when they're defined?

build.sh
```diff
@@ -1,46 +1,195 @@
...
+    fi
 
-    chmod +x $buildFile
+    failed=false
+    if __machine_has curl ; then
+        curl --retry 10 -sSL -f --create-dirs -o $local_path $remote_path || failed=true
+    elif __machine_has wget; then
+        wget --tries 10 -O $local_path $remote_path || failed=true
+    else
+        failed=true
```
pranavkm: Print an error message saying the tool couldn't be found?

natemcmaster: That was added already. Line 176 below.

build.ps1
natemcmaster: To preserve this order of precedence:
1. Explicit CLI flag. `./build.ps1 -ToolsSource C:\dev\aspnet\BuildTools\artifacts`
2. Path in version.props `<KoreBuildToolsSource>https://something/else/</KoreBuildToolsSource>`
3. If neither 1 or 2, fallback to https://aspnetcore.blob.core.windows.net/buildtools

build.ps1
natemcmaster: Fine with me.

PR #526, opened 2017-07-07
Use clone uris and commit hash that the CI specifies

pranavkm:


build/repo.targets
```diff
@@ -196,27 +196,45 @@
       XmlInputPath="$(RepositoryFileWithCommit)"
       Query="//Repository[@Include='%(Repository.Identity)']/@Commit"
       Value="%(Repository.Commit)" />
+
+    <XmlPoke
```
natemcmaster: Starting to think xmlpoke isn't the right solution. This is fine for now, but maybe a RepoTasks project would simplify this.

PR #527, opened 2017-07-10
Make the calculation of the build graph a task so we can pass metadata around

pranavkm:


pranavkm:
@Eilon - we might need this for the 2.0.0 release if we want the repository sources to be determined by the CI.

build/tasks/RepoTasks.csproj
```diff
@@ -0,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
```
natemcmaster: This is how other RepoTasks.csproj are setup. Helps us keep task dependencies consistent with what ships in KoreBuild.

```xml
<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="$(RepoTasksSdkPath)\Sdk.props" Condition="'$(RepoTasksSdkPath)' != '' "/>

   <PropertyGroup>
       <!-- still required to keep VS happy -->
   </PropertyGroup>

  <Import Project="$(RepoTasksSdkPath)\Sdk.targets" Condition="'$(RepoTasksSdkPath)' != '' "/>
</Project>
```

build/tasks/BuildGraph/Repository.cs
```diff
@@ -26,26 +26,20 @@ public Repository(string name)
...
-
-                var repository = Read(provider, directoryInfo.Name, directoryInfo.FullName);
+                var repositoryPath = repositoryPaths[i];
+                var repositoryName = Path.GetFileName(repositoryPath);
+                var repository = Read(provider, repositoryName, repositoryPath);
                 repositories[i] = repository;
-
-                Console.WriteLine($"Done gathering dependency information from {directoryInfo.Name}.");
             });
             sw.Stop();
```
natemcmaster: Is the stopwatch used anymore?

build/tasks/BuildGraph/CalculateBuildGraph.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            RepositoriesToBuildInOrder = repositoriesWithOrder
+                .OrderBy(r => r.order)
+                .Select(r => r.repository)
+                .ToArray();
+
+            return true;
+        }
+    }
+}
```
natemcmaster: Nit: add trailing new lines

build/repo.targets
```diff
@@ -107,69 +111,33 @@
...
 
+      <RestoreGraphOutputs Include="$(_RestoreGraphSpecsDirectory)%(Solution.Repository)\%(Solution.FileName)%(Solution.Extension).json" />
+
       <Solution>
         <AdditionalProperties>RestoreGraphOutputPath=$(_RestoreGraphSpecsDirectory)%(Solution.Repository)\%(Solution.FileName)%(Solution.Extension).json</AdditionalProperties>
       </Solution>
     </ItemGroup>
+  </Target>
 
+  <Target Name="_GenerateRestoreGraphSpecs" DependsOnTargets="_PrepareRestoreGraphSpecs" Inputs="@(Solution)" Outputs="@(RestoreGraphOutputs)">
```
natemcmaster: This looks buggy. Solution files can stay the same even when the repository dependency graph changes.

build/repo.targets
```diff
@@ -154,22 +122,26 @@
       <Solution>
         <AdditionalProperties>RestoreGraphOutputPath=$(_RestoreGraphSpecsDirectory)%(Solution.Repository)\%(Solution.FileName)%(Solution.Extension).json</AdditionalProperties>
       </Solution>
+
+      <GraphSpecInputs Include="
```
pranavkm: Not super accurate since we could emit files as part of generating the spec file, but it's better than re-running it every time now.

pranavkm:
🆙 📅

build/repo.targets
natemcmaster: Seems okay enough for now as long as these cached files wouldn't cause a build to pass that shouldn't. If there are any changes in the repo-to-repo graph, a bad cache will fail to build, right?

pranavkm: > If there are any changes in the repo-to-repo graph, a bad cache will fail to build, right?

Unfortunately not. You would get an incorrect build if we have a cache that's gotten in a state where it no longer reflects the current build order. But the issue I pointed out was with us over-doing the task. Plus we could fine tune this if we see local incremental builds (which is the only thing that should be affected by this) get in a bad state.

natemcmaster: Well, as long as we always force clean on CI we're okay. Still a little concerned as this will be hard to debug...but let's give it a try.

PR #530, opened 2017-07-13
Pin CLI tool references for all projects in a solution

pranavkm:


tools/PinVersions/PinVersionUtility.cs
```diff
@@ -188,20 +198,32 @@ private NuGetVersion GetExactVersion(string name, VersionRange range)
...
+                IDictionary<string, ProjectPinVersionMetadata> pinVersionLookup,
+                List<(LibraryRange, NuGetVersion)> cliToolReferences)
+            {
+                PinVersionLookup = pinVersionLookup;
+                CLIToolReferences = cliToolReferences;
+            }
+
+            public IDictionary<string, ProjectPinVersionMetadata> PinVersionLookup { get; }
+
+            public List<(LibraryRange, NuGetVersion)> CLIToolReferences { get; }
```
pranavkm: They changed the structure to list one CLI tool reference per restore spec file. We'll eagerly generate update nodes. This should no-op for projects without the reference and fix the issue for projects with.

PR #533, opened 2017-08-01
Add ANCM

ryanbrandenburg:
Fixes aspnet/CoreCLR#253. Since we're taking over development on AspNetCoreModule we should build it in Universe instead of in the CoreCLR mirror.

### DIFF
--- a/build/Repositories.props
+++ b/build/Repositories.props
@@ -7,6 +7,7 @@
 
   <ItemGroup>
     <Repository Include="Antiforgery" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
+    <Repository Include="AspNetCoreModule" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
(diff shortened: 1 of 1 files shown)


build/Repositories.props
```diff
@@ -7,6 +7,7 @@
 
   <ItemGroup>
     <Repository Include="Antiforgery" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
+    <Repository Include="AspNetCoreModule" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
```
pranavkm: I don't think we can do this as yet. This requires the C++ sdk which we don't have installed on all our build agents.

pranavkm: We'll also need to update https://github.com/aspnet/AspNetCoreModule/blob/dev/test/AspNetCoreModule.TestSites.Standard/AspNetCoreModule.TestSites.Standard.csproj#L15 to point to the right build of packages if we intend for it to be built as part of our build process.

Eilon:
See https://github.com/aspnet/CoreCLR/issues/253#issuecomment-319762365

PR #535, opened 2017-08-02
Move PinVersions into RepoTasks

natemcmaster:
Removes this repo's dependency on .NET Core 1.1.

cref https://github.com/aspnet/BuildTools/issues/347


build/tasks/PinVersions.cs
```diff
@@ -0,0 +1,74 @@
...
+                ? new DependencyGraphSpecProvider(GraphSpecsRoot)
+                : DependencyGraphSpecProvider.Default;
+
+            var sources = new List<string>();
+            sources.AddRange(PackageSources.Select(i => i.ItemSpec));
+
+            foreach (var item in AdditionalPackageSources)
+            {
+                var path = item.ItemSpec;
+                if (path.StartsWith("http", StringComparison.OrdinalIgnoreCase))
```
pranavkm: This feels weird. It'd be cleaner to construct `PackageSources` in the target with the right set of sources instead.

build/tasks/PinVersions.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        public override bool Execute()
+        {
+            if (PackageSources?.Length == 0)
+            {
+                Log.LogError($"Missing PackageSources. At least one item source must be specified.");
+                return false;
+            }
+
+            if (string.IsNullOrEmpty(BuildRepositoryRoot))
```
pranavkm: Does `Required` not cover empty strings?

build/tasks/RepoTasks.tasks
```diff
@@ -4,4 +4,5 @@
   </PropertyGroup>
 
   <UsingTask TaskName="RepoTasks.CalculateBuildGraph" AssemblyFile="$(_RepoTaskAssembly)" />
+  <UsingTask TaskName="RepoTasks.PinVersions" AssemblyFile="$(_RepoTaskAssembly)" />
```
pranavkm: New line at EOF as long as you're editing this file.

build/tasks/PinVersions.cs
natemcmaster: updated

build/tasks/PinVersions.cs
natemcmaster: It does. Updated to remove the check

PR #536, opened 2017-08-05
Clone repositories over https

pranavkm:


build/repo.targets
```diff
@@ -97,7 +97,9 @@
     <Exec
       Command="$(_CloneArguments) --branch $(CloneBranch)"
       Condition="'$(GateBranchExitCode)'!='0'"
-      WorkingDirectory="$(_CloneRepositoryRoot)" />
+      WorkingDirectory="$(_CloneRepositoryRoot)"
+      EnvironmentVariables="GIT_TERMINAL_PROMPT=0"
+      Timeout="180000" />
```
mikeharder: Are these variables related to the HTTPS change?

build/Repositories.props
```diff
@@ -6,49 +6,49 @@
...
-    <Repository Include="Routing" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
-    <Repository Include="Scaffolding" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
-    <Repository Include="Security" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
-    <Repository Include="ServerTests" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
-    <Repository Include="Session" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
-    <Repository Include="SignalR" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
-    <Repository Include="StaticFiles" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
-    <Repository Include="Testing" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
-    <Repository Include="WebSockets" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
+    <Repository Include="Antiforgery" Commit="" CloneUrl="https://github.com/aspnet/%(Identity)" />
```
pranavkm: We'll require fewer setup pieces on dev boxes to clone repositories (you cannot clone over ssh unless you have a key that's registered with github), but pushing would require having git credentials setup (which I haven't as yet because ssh). We used to be clever and try to clone over the protocol Universe was cloned in. Is that preferable?

build/repo.targets
pranavkm: Yes. The first's to prevent git for asking for a password if the repository is private. This specifically works for Git v2.3 or higher (https://github.com/blog/1957-git-2-3-has-been-released#user-content-the-credential-subsystem-is-now-friendlier-to-scripting). The latter is to force a 3 minute timeout in case we're running git < 2.3 (like I was on my local box) and are stuck with a prompt that won't be responded to.

build/Repositories.props
mikeharder: Ideally I agree we would use the protocol Universe was cloned in, so developers can choose between ssh key or saved password, whichever they prefer.  If we need to choose one, I would take a vote of the engineering team.

pranavkm:
🆙 📅

PR #537, opened 2017-08-10
Remove unused TeamCity logger

natemcmaster:
Removing dead code. The TeamCity service messages writer has been disabled for a while because it's impossible to make it look completely right. MSBuild's default console logger prefixes all stdout lines with whitespace, but TC expects service messages to be at the beginning of the line. This led to the TC log appearing to have large chucks empty lines. We'll need to tie in a TC logger a lower level and disable MSBuild's console logger altogether to make this work right.

see https://github.com/aspnet/BuildTools/issues/345


pranavkm:
Oh well :shipit:

PR #538, opened 2017-08-10
Produce a lineup package containing all Universe packages (plus a few others)

natemcmaster:
Part of https://github.com/aspnet/KoreBuild/issues/239

This will produce a "lineup" package containing a list of all packages produced in the universe build, along with a few other external dependencies.

Example result:
[Internal.AspNetCore.Universe.Lineup.2.1.0-preview1-t004e82595.nupkg.zip](https://github.com/aspnet/Universe/files/1216125/Internal.AspNetCore.Universe.Lineup.2.1.0-preview1-t004e82595.nupkg.zip)

cc @mikeharder @ryanbrandenburg

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -1,0 +1,26 @@
+<Project>
+
+  <!-- ASP.NET Core (non-Universe builds) -->
+  <PropertyGroup>
+    <AspNetCoreFeed>https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json</AspNetCoreFeed>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ExternalDependency Include="Libuv" Version="1.10.0" Source="$(AspNetCoreFeed)" />
+  </ItemGroup>
+
+  <!-- nuget.org -->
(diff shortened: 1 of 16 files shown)


build/dependencies.props
```diff
@@ -0,0 +1,26 @@
...
+    <ExternalDependency Include="Libuv" Version="1.10.0" Source="$(AspNetCoreFeed)" />
+  </ItemGroup>
+
+  <!-- nuget.org -->
+  <PropertyGroup>
+    <DefaultNuGetFeed>https://api.nuget.org/v3/index.json</DefaultNuGetFeed>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ExternalDependency Include="BenchmarkDotNet" Version="0.10.3" Source="$(DefaultNuGetFeed)"/>
```
smitpatel: Need 0.10.9 at least

natemcmaster: For now, we're only experimenting with lineups in Kestrel, which is still using 0.10.3. I'll update Kestrel to 0.10.9 as a separate step so that we unify with EF's benchmarks projects.

PR #540, opened 2017-08-15
Upgrade xunit to 2.3.0-beta4

natemcmaster:
Also, expands the list of xunit packages to include commonly used sub-packages, such as xunit.core.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -19,8 +19,11 @@
     <ExternalDependency Include="Microsoft.NET.Test.Sdk" Version="15.3.0" Source="$(DefaultNuGetFeed)"/>
     <ExternalDependency Include="Moq" Version="4.7.49" Source="$(DefaultNuGetFeed)"/>
     <ExternalDependency Include="Newtonsoft.Json" Version="10.0.1" Source="$(DefaultNuGetFeed)"/>
-    <ExternalDependency Include="xunit.runner.visualstudio" Version="2.3.0-beta3-build3705" Source="$(DefaultNuGetFeed)"/>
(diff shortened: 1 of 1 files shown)


build/dependencies.props
```diff
@@ -19,8 +19,11 @@
     <ExternalDependency Include="Microsoft.NET.Test.Sdk" Version="15.3.0" Source="$(DefaultNuGetFeed)"/>
     <ExternalDependency Include="Moq" Version="4.7.49" Source="$(DefaultNuGetFeed)"/>
     <ExternalDependency Include="Newtonsoft.Json" Version="10.0.1" Source="$(DefaultNuGetFeed)"/>
-    <ExternalDependency Include="xunit.runner.visualstudio" Version="2.3.0-beta3-build3705" Source="$(DefaultNuGetFeed)"/>
-    <ExternalDependency Include="xunit" Version="2.3.0-beta3-build3705" Source="$(DefaultNuGetFeed)"/>
+    <ExternalDependency Include="xunit.analyzers" Version="0.6.1" Source="$(DefaultNuGetFeed)"/>
+    <ExternalDependency Include="xunit.assert" Version="2.3.0-beta4-build3742" Source="$(DefaultNuGetFeed)"/>
```
mikeharder: Why explicitly specify `xunit.analyzers` and `xunit.assert`?

mikeharder: I meant `xunit.assert` and `xunit.core`.  I see you are explicitly specifying a newer version of `xunit.assert`.

natemcmaster: xunit is a metapackage of xunit.{assert, core, analyzers}. In some repros, we may want to reference only some of these but not all (e.g. some people may want to opt-out of analyzers).

mikeharder: OK I get it now, this is just a list of versions, and each project chooses which subset they want.

PR #541, opened 2017-08-15
Add external packages require for aspnet/Common

natemcmaster:
Add more packages for the Universe lineup.


natemcmaster:
See https://github.com/aspnet/Common/pull/259

PR #542, opened 2017-08-17
Remove SignalR from Universe

mikaelm12:


mikaelm12:
Context: https://github.com/aspnet/SignalR/pull/725#discussion_r133564959

PR #551, opened 2017-08-31
Fix `First()` failure when `$(BuildGraphOf)` is set

dougbu:
- still must build entire Universe locally before using `$(BuildGraphOf)`

### DIFF
--- a/build/tasks/CalculateBuildGraph.cs
+++ b/build/tasks/CalculateBuildGraph.cs
@@ -50,7 +50,13 @@ namespace RepoTasks
             foreach (var repositoryTaskItem in Repositories)
             {
                 var repositoryName = repositoryTaskItem.ItemSpec;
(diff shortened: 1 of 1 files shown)


pranavkm:
@natemcmaster's more familiar with this code. Assigning it to him. Looks good to me though.

natemcmaster:
@dougbu what are the exact command line parameters you're trying to use? This fix suppresses an error, but I want to know what leads to it.

dougbu:
@natemcmaster `.\build.cmd /p:CompileOnly=true /p:ShallowClone=true /p:BuildGraphOf=Mvc` for example. I had edited `Repositories.props` before running that but had done nothing else.

The specific problem I'm working around is that `GraphBuilder` returns a graph missing most of what's in the `Repositories` collection in the case above. Actually, this happens in almost any case involving a non-empty `$(BuildGraphOf)` value.

dougbu:
3f8b319

PR #555, opened 2017-09-14
Begin stubbing out the framework build

natemcmaster:
Basically deletes most of what is in universe, except config data for how 'framework build' should work.


natemcmaster:
Discussed more in triage. Re-organizing the build like this is likely going to take more work than we have time for. Let's instead try to do the minimal possible work to get the patch building.

PR #556, opened 2017-09-14
Start preparing list of repositories that need patching

natemcmaster:
Pending patch approvals on bugs for different repos

cc @Eilon


build/Repositories.props
```diff
@@ -1,11 +1,16 @@
 <Project>
-  <ItemDefinitionGroup>
-    <Repository>
-      <Branch>rel/2.0.0</Branch>
-    </Repository>
-  </ItemDefinitionGroup>
-
   <ItemGroup>
+    <!-- Repos being patched -->
+    <Repository Include="EntityFramework" Branch="rel/2.0.1" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
```
smitpatel: Just a minor thing and it should be working regardless. Should this (and may be other references) be `EntityFrameworkCore` now since we renamed repo?

natemcmaster: Good catch. Yeah, looks like dev is still "EF" not "EFCore". Can you make a PR to update it? Also you'll need to update TC  VCS roots so commithashs can still be found.

https://github.com/aspnet/Universe/blob/425c1b83e65953b6dc3bdcbf89e2f4d295fbf855/build/repo.targets#L262-L269

natemcmaster:
:up: :date:

build/Repositories.props
```diff
@@ -1,55 +1,55 @@
 <Project>
-  <ItemDefinitionGroup>
-    <Repository>
-      <Branch>patch/2.0.1</Branch>
```
natemcmaster: FYI - I want to see if we can create our patch system in such a way that we don't have to create and build these patch branches.

build/Repositories.props
```diff
@@ -1,55 +1,55 @@
...
+    <Repository Include="MetaPackages" Branch="rel/2.0.1" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
+
+    <!-- TODO compute downstream repos that should be updated. -->
+
+    <!--
+    Repos not building this patch.
+    Build tools will *verify* that these repos will be unaffected
+    by the patch update and do not need updating.
+    -->
+    <VerifyRepository Include="Antiforgery" Branch="rel/2.0.0" Commit="" CloneUrl="git@github.com:aspnet/%(Identity)" />
```
natemcmaster: We still need the list of packages that these produced so we can build a runtime store with the right set of packages.

PR #560, opened 2017-09-18
Copy packages to ship/noship and verify coherent versions

natemcmaster:
Copies some verification code from aspnet/Coherence into the current repository.

Changes 
 - Copies files to artifacts/{ship, noship, shipoop]
 - Ensures compiled packages have self-consistent versions
 - Replaces packages.csv with an item group

TODO:
 - Add external dependency version verification
 - Update TeamCity configuration

PR #561, opened 2017-09-19
Verify external dependencies

natemcmaster:
Follow up to #560.

This ensures the following:

1. Projects has cannot have a PackageReference to anything not listed as an ExternalDependency
1. Shipping packages has cannot have dependnecy to anything not listed as an ExternalDependency AND marked as Private=false
1. Warns when multiple versions of an external dependency are used

cc @ryanbrandenburg


build/dependencies.props
```diff
@@ -0,0 +1,195 @@
...
+    <ExternalDependency Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="2.6.0-beta1-61924-08" Source="$(RoslynFeed)" >
+      <!-- This version is required for Razor to build the VSIX for VS. -->
+      <NoWarn>KRB2004</NoWarn>
+      <Private>true</Private>
+    </ExternalDependency>
+    <ExternalDependency Include="Microsoft.VisualStudio.LanguageServices.Razor.RemoteClient" Version="2.6.0-beta1-61924-08" Source="$(RoslynFeed)">
+      <Private>true</Private>
+    </ExternalDependency>
+    <ExternalDependency Include="Microsoft.VisualStudio.LanguageServices" Version="2.6.0-beta1-61924-08" Source="$(RoslynFeed)">
+      <Private>true</Private>
```
mikeharder: Some items in this file set `Private` as a child element while others set it inline.  Should we be consistent?

natemcmaster: yeah. I'll update

PR #565, opened 2017-09-22
Generate runtime store

JunTaoLuo:
TODO:
- [x] Build All metapackage
  - Using artifacts.props
- [x] Build RuntimeStore
- [x] Build HostingDeps
- [x] Create delta zips
  - using previous manifests
- [x] Cleanup

I will port the installer builds steps in a separate PR.


build/RuntimeStore.targets
```diff
@@ -0,0 +1,211 @@
+<Project>
+  <Import Project="common.props" />
+  <Import Project="existingStores.props" />
```
JunTaoLuo: Note, this will no longer be needed after moving to manifest based delta zip calculations.

build/RuntimeStore.targets
```diff
@@ -0,0 +1,199 @@
...
+    <_ExistingStoreDirectory>$(_WorkRoot).es\</_ExistingStoreDirectory>
+    <_RuntimeStoreReferencePackageDirectory>$(_WorkRoot)bin\Release\</_RuntimeStoreReferencePackageDirectory>
+    <_TemplatesDirectory>$(MSBuildThisFileDirectory)tools\templates\</_TemplatesDirectory>
+    <_SrcDirectory>$(RepositoryRoot)src\</_SrcDirectory>
+    <_ExistingManifestsDirectory>$(_SrcDirectory)Microsoft.AspNetCore.All\build\</_ExistingManifestsDirectory>
+    <_ArtifactsZipDirectory>$(ArtifactsDir)zip\</_ArtifactsZipDirectory>
+    <_DepsOutputDirectory>$(ArtifactsDir)deps\</_DepsOutputDirectory>
+  </PropertyGroup>
+
+  <Target Name="BuildAllMetapackage" DependsOnTargets="ResolveRepoInfo">
```
JunTaoLuo: This doesn't feel like the right approach. Maybe we can run the repo task as part of building the package instead of having this separate target. All that's needed is to populate the references before building.

src/Microsoft.AspNetCore.All/Microsoft.AspNetCore.All.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <VersionPrefix Condition="'$(MetapackageVersionPrefix)' != ''">$(MetapackageVersionPrefix)</VersionPrefix>
```
JunTaoLuo: Can just use the version prefix.

JunTaoLuo:
Ready for review.

build/RuntimeStore.targets
```diff
@@ -0,0 +1,176 @@
...
+    <_AllMetapackageDirectory>$(_SrcDirectory)Microsoft.AspNetCore.All\</_AllMetapackageDirectory>
+    <_AllMetapackageProject>$(_AllMetapackageDirectory)Microsoft.AspNetCore.All.csproj</_AllMetapackageProject>
+    <_ExistingManifestsDirectory>$(_AllMetapackageDirectory)build\</_ExistingManifestsDirectory>
+    <_ArtifactsZipDirectory>$(ArtifactsDir)zip\</_ArtifactsZipDirectory>
+    <_StoreZipDirectory>$(_ArtifactsZipDirectory)rs\</_StoreZipDirectory>
+    <_SymbolsZipDirectory>$(_ArtifactsZipDirectory)symbols\</_SymbolsZipDirectory>
+    <_DepsOutputDirectory>$(ArtifactsDir)deps\</_DepsOutputDirectory>
+  </PropertyGroup>
+
+  <Target Name="BuildAllMetapackage" DependsOnTargets="ResolveRepoInfo">
```
JunTaoLuo: This doesn't feel like the right approach. Maybe we can run the repo task as part of building the package instead of having this separate target. All that's needed is to populate the references before building.

build/RuntimeStore.targets
```diff
@@ -0,0 +1,176 @@
...
+    <!-- Generate RS.Reference project -->
+    <RepoTasks.AddMetapackageReferences
+      ReferencePackagePath="$(_AllMetapackageProject)"
+      BuildArtifacts="@(ArtifactInfo)"
+      PackageArtifacts="@(PackageArtifact)" />
+
+    <!-- Restore and pack-->
+    <MSBuild Projects="$(_AllMetapackageProject)"
+      Targets="Restore;Pack"
+      Properties="Configuration=Release" />
```
natemcmaster: `Configuration=$(Configuration)`

build/RuntimeStore.targets
```diff
@@ -0,0 +1,176 @@
...
+  </Target>
+
+  <Target Name="UpdateNuGetConfig">
+    <UpdatePackageSource
+      NuGetConfigPath="$(RepositoryRoot)NuGet.config"
```
natemcmaster: Can we either make a non-tracked copy of this file or use MSBuild properties instead? I'd like to avoid the UpdatePackageSource task going forward because it makes changes to a file tracked by git. I've mistakenly committed the changes it makes a few times now.

If you want to use MSBuild props, these properties can be used to control the feeds used to restore a project:

RestoreSources
RestoreFallbackFolders
RestoreAdditionalProjectSources
RestoreAdditionalProjectFallbackFolders
RestoreConfigFile

build/RuntimeStore.targets
```diff
@@ -0,0 +1,176 @@
...
+  <Target Name="_ResolveRuntimeStoreRID">
+    <Error Text="Please specify the platform for the runtime package store via RUNTIMESTORE_PLATFORM: {x64|x86}." Condition="'$(RUNTIMESTORE_PLATFORM)' == ''"/>
+
+    <GetOSPlatform>
+      <!-- Returns {Linux, macOS, Windows} -->
+      <Output TaskParameter="PlatformName" PropertyName="OSPlatform" />
+    </GetOSPlatform>
+
+    <PropertyGroup>
+      <RID Condition="'$(OSPlatform)' == 'Windows'">win7-$(RUNTIMESTORE_PLATFORM)</RID>
```
natemcmaster: Nit: naming. How about `RuntimeStoreRuntimeIdentifier`?

build/RuntimeStore.targets
```diff
@@ -0,0 +1,176 @@
...
+      <RID Condition="'$(OSPlatform)' == 'Linux'">linux-$(RUNTIMESTORE_PLATFORM)</RID>
+      <RID Condition="'$(OSPlatform)' == 'macOS'">osx-$(RUNTIMESTORE_PLATFORM)</RID>
+    </PropertyGroup>
+  </Target>
+
+  <Target Name="BuildRuntimeStore" DependsOnTargets="UpdateNuGetConfig;_ResolveRuntimeStoreRID;_PrepareRuntimeStoreBuildAssets;_ResolveCurrentSharedFrameworkVersion">
+    <!-- Build runtime store -->
+    <Exec Command="dotnet store -m $(_WorkRoot)RS.Manifest.csproj -f netcoreapp2.0 -r $(RID) -o $(_RuntimeStoreOutputDirectory) --framework-version 2.0.2-* -w $(_RuntimeStoreWorkDirectory) $(RUNTIMESTORE_ADDITIONAL_ARGS)" />
```
natemcmaster: What would it take to remove the floating version? Floating dependency versions is evil.

build/RuntimeStore.targets
```diff
@@ -0,0 +1,176 @@
...
+
+  <Target Name="BuildHostingDeps" DependsOnTargets="ResolveRepoInfo;_ResolveCurrentSharedFrameworkVersion">
+    <!-- Generate RS.Reference project -->
+    <RepoTasks.ResolveHostingStartupPackages
+      BuildArtifacts="@(ArtifactInfo)"
+      PackageArtifacts="@(PackageArtifact)">
+      <Output TaskParameter="HostingStartupArtifacts" ItemName="HostingStartupArtifacts" />
+    </RepoTasks.ResolveHostingStartupPackages>
+
+    <!--- MSBuild caches things if you run inproc so have to use Exec -->
```
natemcmaster: What are you having trouble cache-busting? inproc MSBuild calls are ideal because it preserves logging, parameters, etc.

build/RuntimeStore.targets
```diff
@@ -0,0 +1,176 @@
...
+
+    <!-- Copy and update build assets -->
+    <Copy SourceFiles="$(RepositoryRoot)NuGet.config" DestinationFiles="$(_WorkRoot)NuGet.config" />
+    <Copy SourceFiles="$(_TemplatesDirectory)RS.Manifest\RS.Manifest.csproj" DestinationFiles="$(_WorkRoot)RS.Manifest.csproj" />
+  </Target>
+
+  <Target Name="_ResolveRuntimeStoreRID">
+    <Error Text="Please specify the platform for the runtime package store via RUNTIMESTORE_PLATFORM: {x64|x86}." Condition="'$(RUNTIMESTORE_PLATFORM)' == ''"/>
+
+    <GetOSPlatform>
```
natemcmaster: MSBuild 15.3 has a built-in function for this now :)
```
Condition="$([MSBuild]::IsOSPlatform('Windows'))"
Condition="$([MSBuild]::IsOSPlatform('Linux'))"
Condition="$([MSBuild]::IsOSPlatform('OSX'))"
```

build/RuntimeStore.targets
```diff
@@ -0,0 +1,176 @@
...
+    <!-- Clear working directory -->
+    <RemoveDir Directories="$(_WorkRoot)" />
+
+    <!-- Copy and update build assets -->
+    <Copy SourceFiles="$(RepositoryRoot)NuGet.config" DestinationFiles="$(_WorkRoot)NuGet.config" />
+    <Copy SourceFiles="$(_TemplatesDirectory)RS.Manifest\RS.Manifest.csproj" DestinationFiles="$(_WorkRoot)RS.Manifest.csproj" />
+  </Target>
+
+  <Target Name="_ResolveRuntimeStoreRID">
+    <Error Text="Please specify the platform for the runtime package store via RUNTIMESTORE_PLATFORM: {x64|x86}." Condition="'$(RUNTIMESTORE_PLATFORM)' == ''"/>
```
natemcmaster: Nit: naming. How about `RuntimeStoreArch`?

build/RuntimeStore.targets
```diff
@@ -0,0 +1,176 @@
...
+    <ZipArchive File="$(OutputSymbolZip)" SourceFiles="@(OutputSymbolZipFiles)" WorkingDirectory="$(_SymbolsZipDirectory)" Overwrite="true" Condition="'@(OutputSymbolZipFiles)' != ''"/>
+  </Target>
+
+  <Target Name="CreateCommonManifest" DependsOnTargets="_PrepareRuntimeStoreBuildAssets">
+    <PropertyGroup>
+      <CommonManifestFileName>aspnetcore-store-$(VersionPrefix)-$(VersionSuffix)-common.xml</CommonManifestFileName>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <PackagesToTrim Include="runtime.win-arm64.runtime.native.system.data.sqlclient.sni" />
```
natemcmaster: Can you add a comment about why this is excluded? Also, is this temporary or a permanent workaround? If temporary, can we put this somewhere more visible so we remember to undo in the future?

build/RuntimeStore.targets
```diff
@@ -0,0 +1,176 @@
...
+
+    <MSBuild Projects="$(_WorkRoot)RS.Manifest.csproj" Targets="Restore;GetPackageDefinitions" >
+      <Output TaskParameter="TargetOutputs" ItemName="_PackageDefinitions" />
+    </MSBuild>
+
+    <RepoTasks.CreateCommonManifest DestinationFilePath="$(ArtifactsDir)$(CommonManifestFileName)" PackageDefinitions="@(_PackageDefinitions)" Packages="@(PackagesToTrim)"/>
+  </Target>
+
+  <Target Name="AddManifestsToMetapackage">
+    <Error Text="COHERENCE_DROP_LOCATION is not specified. This requires a path like \\aspnetci\drops\Coherence\{branch}\{build-number}"
```
natemcmaster: This will need to change in the future We cannot impose our TeamCity and network-share settings on external partners who want to build aspnetcore from source. They should be able to clone Universe and execute `build.sh` and get the entire set of packages and a runtime store in one command.

build/RuntimeStore.targets
```diff
@@ -0,0 +1,176 @@
...
+        Condition=" '$(COHERENCE_DROP_LOCATION)' == '' " />
+
+    <ItemGroup>
+      <MetaPackageNupkg Include="$(COHERENCE_DROP_LOCATION)\ship\Microsoft.AspNetCore.All.*.nupkg" />
+      <ManifestFiles Include="$(ArtifactsDir)*.xml"/>
+    </ItemGroup>
+
+    <RemoveDir Directories="@(MetaPackageNupkg->'$(_WorkRoot)%(Filename)')" />
+    <Copy SourceFiles="@(MetaPackageNupkg)" DestinationFolder="$(ArtifactsDir)" />
+    <UnzipArchive File="$(ArtifactsDir)%(MetaPackageNupkg.FileName)%(MetaPackageNupkg.Extension)" Destination="@(MetaPackageNupkg->'$(_WorkRoot)%(Filename)')" />
```
natemcmaster: Instead of unzip/rezipping the nupkg, can we defer the nupkg creation until available files are ready?

build/tasks/AddMetapackageReferences.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            var xmlDoc = new XmlDocument();
+            xmlDoc.Load(ReferencePackagePath);
+
+            // Project
+            var projectElement = xmlDoc.FirstChild;
+
+            // Items
+            var itemGroupElement = xmlDoc.CreateElement("ItemGroup");
+            Log.LogMessage(MessageImportance.High, $"Runtime store will include the following packages");
```
natemcmaster: Nit: MessageImportance.Normal.

build/tasks/AddMetapackageReferences.cs
```diff
@@ -0,0 +1,69 @@
...
+            foreach (var package in metapackageArtifacts)
+            {
+                var packageName = package.GetMetadata("Identity");
+                var packageVersion = buildArtifacts
+                    .Single(p => string.Equals(p.PackageInfo.Id, packageName, StringComparison.OrdinalIgnoreCase))
+                    .PackageInfo.Version.ToString();
+                Log.LogMessage(MessageImportance.High, $" - Package: {packageName} Version: {packageVersion}");
```
natemcmaster: Also MessageImportance.Normal. MessageImportance.High should be used sparingly.

build/tasks/AddMetapackageReferences.cs
```diff
@@ -0,0 +1,69 @@
...
+
+                var packageReferenceElement = xmlDoc.CreateElement("PackageReference");
+
+                var includeAttribute = xmlDoc.CreateAttribute("Include");
+                var versionAttribute = xmlDoc.CreateAttribute("Version");
+                includeAttribute.Value = packageName;
+                versionAttribute.Value = packageVersion;
+                packageReferenceElement.Attributes.Append(includeAttribute);
```
natemcmaster: I think this can be simplified to `packageReferenceElement.Attributes["Include"] = packageName;`

build/RuntimeStore.targets
```diff
@@ -0,0 +1,176 @@
...
+    <_ArtifactsZipDirectory>$(ArtifactsDir)zip\</_ArtifactsZipDirectory>
+    <_StoreZipDirectory>$(_ArtifactsZipDirectory)rs\</_StoreZipDirectory>
+    <_SymbolsZipDirectory>$(_ArtifactsZipDirectory)symbols\</_SymbolsZipDirectory>
+    <_DepsOutputDirectory>$(ArtifactsDir)deps\</_DepsOutputDirectory>
+  </PropertyGroup>
+
+  <Target Name="BuildAllMetapackage" DependsOnTargets="ResolveRepoInfo">
+    <!-- Generate RS.Reference project -->
+    <RepoTasks.AddMetapackageReferences
+      ReferencePackagePath="$(_AllMetapackageProject)"
```
natemcmaster: I have the same concern here as with the NuGet.config file. Editing tracked-files in place makes builds messy. Generated files that are intermediates shouldn't be tracked by git. Can you copy this file to obj/ instead?

build/tasks/ComposeNewStore.cs
```diff
@@ -0,0 +1,131 @@
...
+
+                    File.Copy(storeFile.GetMetadata("FullPath"), Path.Combine(destinationDir, $"{storeFile.GetMetadata("Filename")}{storeFile.GetMetadata("Extension")}"), overwrite: true);
+                }
+            }
+
+            // Insert new runtime store files
+            foreach (var symbolFile in RuntimeStoreSymbolFiles)
+            {
+                // format: {bitness}}/{tfm}}/{id}/{version}}/...
+                var recursiveDir = symbolFile.GetMetadata("RecursiveDir");
```
natemcmaster: Any time you use file paths coming from MSBuild items, normalize dir separators first to '/'. We've had lots of bugs in the past on Linux from missing this step.

build/tasks/ConsolidateManifests.cs
```diff
@@ -0,0 +1,62 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Xml;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace RepoTasks
+{
+    public class ConsolidateManifests : Task
```
natemcmaster: Nit: document what this task does

build/tasks/TrimDeps.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.IO;
+using System.Linq;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace RepoTasks
+{
+    public class TrimDeps : Task
```
natemcmaster: What are we trimming and why? Is this a workaround for limitations in dotnet-store?

build/tools/scripts/GetSharedFrameworkVersion.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
natemcmaster: Use env in shebangs for machines where bash is not here.

`#!/usr/bin/env bash`

build/RuntimeStore.targets
JunTaoLuo: RuntimeStoreRID?

build/RuntimeStore.targets
JunTaoLuo: Good call, should be able to use `$(SharedFrameworkVersion)` now that I parse it. It was hard coded due to limitations that is now ancient history.

build/RuntimeStore.targets
JunTaoLuo: I'm not sure. I copied @pakrym's code here

build/RuntimeStore.targets
JunTaoLuo: Permanent, so I'll add a comment. The reason is that the sqlclient package brings in this dependency for non-standalone apps. However since we do not build a runtime store for `win-arm64` this package isn't included in our manifest that dictates which packages to trim. However we know for sure that this package will be included in the runtime so we should trim it from the output.

build/RuntimeStore.targets
JunTaoLuo: Let me give this a shot. I think it should work.

build/RuntimeStore.targets
JunTaoLuo: Might be able to work around this. But in general, as discussed offline, we will most likely need a way to flow artifacts between builds.

build/tasks/AddMetapackageReferences.cs
JunTaoLuo: It seems to be read-only: 
```
public XmlAttribute this[string name] { get; }
```

build/RuntimeStore.targets
pakrym: We use same project file to restore and build deps file per hosting startup, so we were cache busting restore targets and build targets when using same csproj but different dependency.

build/tasks/AddMetapackageReferences.cs
natemcmaster: oh, right. What about XmlElement.SetAttributeValue(string, object)?

build/tasks/TrimDeps.cs
JunTaoLuo: This trims the first entry which is the entry assembly, i.e. `HostingStartup` which is the template used to generate the deps file and we don't want to include it.

build/tasks/AddMetapackageReferences.cs
JunTaoLuo: Ah yes, that works  `XmlElement.SetAttribute(string, string)`, nice.

build/RuntimeStore.targets
natemcmaster: Ok. You can cache-bust MSBuild's stored result by passing in new properties.

build/RuntimeStore.targets
JunTaoLuo: I've moved the nuget.config to the work directory. We can't really just pass in properties because this is required for dotnet-store which looks for a NuGet.config inside the working dir.

build/RuntimeStore.targets
JunTaoLuo: Discussed offline, not worth it to pursue this.

build/RuntimeStore.targets
pakrym: I tried, we are passing lots of things as properties, it didn't work correctly.

build/tasks/ComposeNewStore.cs
JunTaoLuo: What do you mean normalize? https://github.com/dotnet/dotnet-cli-archiver/pull/3/files#diff-e872d285abbe85ecc125a21cdbee8021R7? I'm not doing anything that assumes slash or back-slash since I'm just splitting by DirectorySeparatorChar. Does that not work?

build/RuntimeStore.targets
JunTaoLuo: Building in work dir instead. Where should the built artifacts, such as the nupkg, go?

build/RuntimeStore.targets
natemcmaster: I'm fine with this for now. I've been [prototyping](https://github.com/aspnet/Universe/tree/namc/repoproj) another approach to building universe that shaves a lot of time off build. We can address this again when/if we do it.

build/RuntimeStore.targets
```diff
@@ -0,0 +1,181 @@
...
+      NuGetConfigPath="$(_WorkRoot)NuGet.config"
+      SourceName="Dependencies"
+      SourceUri="$(_DependencyBuildDirectory)"
+      Condition="Exists('$(_DependencyBuildDirectory)')" />
+  </Target>
+
+  <Target Name="_ResolveRuntimeStoreRID">
+    <Error Text="Please specify the platform for the runtime package store via RuntimeStoreArch: {x64|x86}." Condition="'$(RuntimeStoreArch)' == ''"/>
+
+    <GetOSPlatform>
```
natemcmaster: You can remove this task call.

src/Microsoft.AspNetCore.All/build/aspnetcore-store-2.0.0-linux-x64.xml
```diff
@@ -0,0 +1,171 @@
+<StoreArtifacts>
+   <Package Id="libuv"  Version ="1.10.0"/>
```
natemcmaster: Nit: is the extra space intentional? `Version=` vs `Version =`

JunTaoLuo: Never paid attention to the output of the dotnet-store manifests. This is a dotnet-store bug in their format.

PR #569, opened 2017-09-29
Generate the package version props file

natemcmaster:
This should allow us to override RuntimeFrameworkVersion automatically based on TFM.


natemcmaster:
Example output: https://gist.github.com/natemcmaster/263a30f0593526d0bfbd971a13c49f12

PR #582, opened 2017-10-06
Removing Microsoft.AspNetCore.Server.IIS package

jkotalik:
Reaction to: https://github.com/aspnet/IISIntegration/pull/443. Keeping this in the same namespace and extension will allow for a seamless transition.


natemcmaster:
Like always, wait to merge this until your change to aspnet/IISIntegration is in.

PR #585, opened 2017-10-06
[2.0.x] Improvements to build reproducibility

natemcmaster:
- Remove myget feeds during build
 - Update the method in which we select package versions
 - Remove obsolete 'GenerateLineup' targets/tasks

Example of changes in repos: https://github.com/aspnet/EntityFrameworkCore/pull/10005


src/Microsoft.AspNetCore.All/Microsoft.AspNetCore.All.csproj
```diff
@@ -5,11 +5,14 @@
 
   <PropertyGroup>
     <RestoreSources>$(DotNetRestoreSources)</RestoreSources>
-    <RestoreSources Condition="'$(DotNetBuildOffline)' != 'true'">
+    <RestoreSources Condition="'$(DotNetBuildOffline)' != 'true' AND '$(AspNetUniverseBuildOffline)' != 'true' ">
       $(RestoreSources);
       https://dotnet.myget.org/F/aspnet-2-0-2-october2017-patch/api/v3/index.json;
       https://dotnet.myget.org/F/aspnetcore-master/api/v3/index.json;
       https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json;
```
JunTaoLuo: Seems like we are removing some of these feeds in the base NuGet.config. We can probably remove them from here as well? Care to give that a shot?

natemcmaster: Yeah, we should be okay to remove these. This project should almost always be using locally-built packages anyways.

PR #587, opened 2017-10-09
Copy symbol packages to artifacts/symbols

### LINKED ISSUE #584: Move symbol packages to artifacts/symbols
These packages are consumed in https://github.com/aspnet/Coherence-Signed/blob/dev/build/Prepare.targets#L50.

I think we'll need this.

natemcmaster:
Resolves #584

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -209,7 +209,7 @@
 
   <Target Name="SplitPackages">
     <ItemGroup>
-      <PackageArtifactFile Include="$(BuildDir)*.nupkg" Exclude="$(BuildDir)*.symbols.nupkg" />
+      <PackageArtifactFile Include="$(BuildDir)*.nupkg" />
       <ExternalDependencyPackage Include="@(ExternalDependency->WithMetadataValue('Mirror', 'true'))" Category="mirror" />
       <_MirroredPackageFiles Include="$(IntermediateMirrorPackageDir)*.nupkg" />
     </ItemGroup>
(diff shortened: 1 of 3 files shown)


build/repo.targets
```diff
@@ -229,7 +229,14 @@
 
   <Target Name="SplitPackages">
     <ItemGroup>
-      <PackageArtifactFile Include="$(BuildDir)*.nupkg" Exclude="$(BuildDir)*.symbols.nupkg" />
+      <PackageSymbolsArtifactFile Include="$(BuildDir)*.symbols.nupkg" />
+    </ItemGroup>
+
+    <MakeDir Directories="$(ArtifactsDir)symbols\" />
+    <Move SourceFiles="@(PackageSymbolsArtifactFile)" DestinationFolder="$(ArtifactsDir)symbols\" />
```
JunTaoLuo: Sure we want to move instead of copy?

natemcmaster: Good question. What does the signing step expect?

JunTaoLuo: Well it doesn't really matter for signing since they look under the `.deps\symbols` directory. The question is do we want to keep the files under the original build directory. There's no harm either way.

JunTaoLuo: Okay maybe TC needs to unzip more files but that's not really an issue.

natemcmaster: One problem we have today is that the TC artifacts contains duplicate copies of our .nupkg files. I'm wondering if we should reduce duplication altogether by changing the splitpackages step into a move, not copy.

natemcmaster: The obvious downside is that we have to change downstream builds to be aware of all the subfolders in artifacts/. Today, we just assume everything is in artifacts/build

JunTaoLuo: Yea let's be consistent and choose one way or the other.

natemcmaster: Ok, for now, I'll make this a copy to stay consistent with dev and our current approach.

PR #598, opened 2017-10-12
Add aspnet/Templating to the universe build

### LINKED ISSUE #579: [2.0.x] Add aspnet/templating
We should add the aspnet/templating repo to the main product build so we can ensure templates and new runtime builds align.

cc @SteveSandersonMS

natemcmaster:
Build the templates in a universe build.

Depends on https://github.com/aspnet/templating/pull/58

Resolves #579

cc @SteveSandersonMS

### DIFF
--- a/build/Templating.targets
+++ b/build/Templating.targets
@@ -1,0 +1,69 @@
+<Project>
+  <PropertyGroup>
+    <TemplatingProjectRoot>$(MSBuildThisFileDirectory)..\modules\Templating\</TemplatingProjectRoot>
+  </PropertyGroup>
+
+  <Target Name="BuildTemplates" DependsOnTargets="GeneratePropsFiles">
+    <PropertyGroup>
+      <GeneratedNoTimestampPackageVersionPropsPath>$(IntermediateDir)dependencies.notimestamp.props</GeneratedNoTimestampPackageVersionPropsPath>
+      <TemplateProjCommmonProperties>
+        RepositoryRoot=$(TemplatingProjectRoot);
+        DotNetRestoreSourcesPropsPath=$(GeneratedRestoreSourcesPropsPath);
+        BuildNumber=$(BuildNumber);
+        Configuration=$(Configuration);
+      </TemplateProjCommmonProperties>
+      <TemplateProjProperties>
+        $(TemplateProjCommmonProperties);
+        DotNetPackageVersionPropsPath=$(GeneratedPackageVersionPropsPath);
+        VersionMetadata=timestamped;
+      </TemplateProjProperties>
+      <TemplateProjNoTimestampProperties>
+        $(TemplateProjCommmonProperties);
+        DotNetPackageVersionPropsPath=$(GeneratedNoTimestampPackageVersionPropsPath);
+        BuildNumber=$(BuildNumber)-final;
+        VersionMetadata=notimestamp;
+      </TemplateProjNoTimestampProperties>
+    </PropertyGroup>
+
+    <!-- Produce regular, timestamped templates for pre-release builds -->
+    <MSBuild Projects="$(MSBuildProjectFullPath)"
+             Targets="Restore;Compile;Package"
+             Properties="$(TemplateProjProperties)" />
+
+    <ItemGroup>
+      <TemplateArtifacts Include="$(TemplatingProjectRoot)artifacts\build\*" />
+    </ItemGroup>
+
+    <Copy SourceFiles="@(TemplateArtifacts)" DestinationFolder="$(BuildDir)" />
+
+    <!-- Gather artifact info as if this were a final build -->
+    <MSBuild Projects="$(MSBuildProjectFullPath)"
+             Targets="GetArtifactInfo"
+             Properties="RepositoryRoot=%(Repository.RootPath);Configuration=$(Configuration);BuildNumber=$(BuildNumber);IsFinalBuild=true"
+             ContinueOnError="WarnAndContinue">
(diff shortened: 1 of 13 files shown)


build/Templating.targets
```diff
@@ -0,0 +1,68 @@
...
+      <Output TaskParameter="TargetOutputs" ItemName="FinalBuildArtifactInfo" />
+    </MSBuild>
+
+    <ItemGroup>
+      <_NoTimestampPackages Include="@(ExternalDependency)" />
+      <_NoTimestampPackages Include="%(FinalBuildArtifactInfo.PackageId)" Version="%(FinalBuildArtifactInfo.Version)" Condition=" '%(FinalBuildArtifactInfo.ArtifactType)' == 'NuGetPackage' " />
+      <_NoTimestampPackages Include="Microsoft.AspNetCore.All" Version="$(PackageVersionNoTimestamp)" />
+    </ItemGroup>
+
+    <RepoTasks.GeneratePackageVersionPropsFile
```
natemcmaster: FYI @SteveSandersonMS  - this is the piece that updates version numbers. It spits out a file with MSBuild properties containing version overrides. It's passed into aspnet/Templating via the DotNetPackageVersionPropsPath property.

natemcmaster:
For now, I have not added aspnet/Templating to the tests. I couldn't get them to pass on Windows or Linux. Once we stabilize the tests, we can add this to our build as well.

build/artifacts.props
```diff
@@ -113,6 +113,7 @@
     <PackageArtifact Include="Microsoft.AspNetCore.Server.Kestrel" Category="ship" Metapackage="true" />
     <PackageArtifact Include="Microsoft.AspNetCore.Session" Category="ship" Metapackage="true" />
     <PackageArtifact Include="Microsoft.AspNetCore.SpaServices" Category="ship" Metapackage="true" />
+    <PackageArtifact Include="Microsoft.AspNetCore.SpaTemplates" Category="shipoob" />
```
SteveSandersonMS: What happens differently for `shipoob` packages versus `ship` ones?

build/Templating.targets
```diff
@@ -0,0 +1,68 @@
...
+      </TemplateProjProperties>
+      <TemplateProjNoTimestampProperties>
+        $(TemplateProjCommmonProperties);
+        DotNetPackageVersionPropsPath=$(GeneratedNoTimestampPackageVersionPropsPath);
+        VersionMetadata=notimestamp;
+      </TemplateProjNoTimestampProperties>
+    </PropertyGroup>
+
+    <!-- Produce regular, timestamped templates for pre-release builds -->
+    <MSBuild Projects="$(MSBuildProjectFullPath)"
```
SteveSandersonMS: For templates, I'm not sure we need to have this distinction between timestamped and no-timestamped package versions. Since end users don't install the template packages manually, there's no drawback to having the build numbers in the packages even in the final releases. In fact our current template packages do have build numbers in their names when shipped.

Would it make things much simpler for you if you didn't need to make that distinction?

build/artifacts.props
JunTaoLuo: Once the final packages are built packages marked as ship will be uploaded to nuget. shipoob packages are still produced by builds but are only available on the aspnetci drop share. Generally shipoob packages are released via a different mechanism.

natemcmaster: Will these ship on nuget.org when we release the runtime? If so, they should be `ship`. I wasn't sure so I picked "shipoob".

build/Templating.targets
natemcmaster: Yes, would love to simplify this if possible. I was under the impression the final nupkg we gave to the dotnet/CLI team needed to contain `content/*/*.csproj` files that had code like this
```xml
 <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.3" />
```
Is this not the case?

build/artifacts.props
SteveSandersonMS: Got it - thanks for the info.

The `SpaTemplates` one will ship on NuGet, since that's the only way customers can acquire it (it's not in the runtime package store and doesn't ship in the CLI). The others can be `shipoob`.

build/Templating.targets
SteveSandersonMS: Oh I see. Yes that is right so I think what you have here should stay.

I meant that the template packages themselves (but not the runtime dependencies for generated projects) would be OK to have timestamped versions for release.

build/artifacts.props
SteveSandersonMS: To make sure this is completely clear:

* `Microsoft.AspNetCore.SpaTemplates` should go to NuGet.org (`ship`)
* But `Microsoft.DotNet.Web.Spa.ProjectTemplates` and the other packages can be `shipoob` as they are bundled into the CLI.

build/Templating.targets
natemcmaster: Thanks for clarifying.

The only possible problem I see with this is how the CLI consumes our packages. Do they pull from a myget feed? If so, while we were preparing the final build of CLI and had to reset a non-timestamped package version, it comes really tricky to invalid nuget caches with a new version of `Microsoft.AspNetCore.SpaTemplates/2.0.3`. Maybe something like this?

Microsoft.AspNetCore.SpaTemplates 2.0.3-rtm-build-123-final => contains templates with non-timestamped versions
Microsoft.AspNetCore.SpaTemplates 2.0.3-rtm-build123 => contains templates with prerelease labels

natemcmaster:
Merged in https://github.com/aspnet/Universe/commit/446de15f0bc1b8d7d651c5e90c11c0479512e8b8

PR #599, opened 2017-10-12
Produce separate reference package for runtime store generation

JunTaoLuo:
There is a lot of code duplication here. For example BuildAllMetapackage and BuildRSReferencePackage targets uses mostly the same logic and can be cleaned up. Same with the Add*References.cs build tasks. This change will also allow us to avoid building then unpacking and repacking the .All metapackage. These changes will follow in a cleanup PR later (I'll file an issue) but I want to do the minimal amount of required work for now to get the builds running.

### DIFF
--- a/build/tools/templates/RS.References/RS.References.csproj
+++ b/build/tools/templates/RS.References/RS.References.csproj
@@ -1,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="$(DotNetRestoreSourcePropsPath)" Condition="'$(DotNetRestoreSourcePropsPath)' != ''"/>
+
+  <PropertyGroup>
+    <RestoreSources>$(DotNetRestoreSources)</RestoreSources>
+    <RestoreSources Condition="'$(DotNetBuildOffline)' != 'true' AND '$(AspNetUniverseBuildOffline)' != 'true' ">
+      $(RestoreSources);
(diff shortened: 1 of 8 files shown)


build/tools/templates/RS.References/RS.References.csproj
```diff
@@ -0,0 +1,19 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\build\common.props" />
```
natemcmaster: I made a contending change in #598 to use `src/Directory.Build.props` instead of the import to this file.

natemcmaster:
I'd rather implement it "right" the first time. It's actually more work to refactor this later, and these kinds of cleanup often get dropped. But I understand the time crunch. Can I help on this one?

natemcmaster:
Talked in person. :shipit:  for now. We'll cleanup next week.

JunTaoLuo:
Followup filed in https://github.com/aspnet/Universe/issues/600.

build/tools/templates/RS.References/RS.References.csproj
JunTaoLuo: That's src though? This project gets copied over to .w/RS.References/RS.REferences.csproj

natemcmaster: Yup, I copy directory.build.props too. https://github.com/aspnet/Universe/pull/598/files#diff-20ec4e3e95e20c986d83f609fb34ba1aR34

PR #603, opened 2017-10-13
Johluo/installers

JunTaoLuo:
Porting the last bit of infrastructure from https://github.com/aspnet/MetaPackages/blob/dev/build/repo.targets.

This PR will enable the generation of the linux installers and will be the last missing configuration on the CI. The test scripts for the installers are still missing so these artifacts are not verified. I'll look into testing them if I have time before I get back otherwise I'll test them out on Wednesday.


natemcmaster:
Didn't review carefully, but if it works, go for it. We can fixup stuff that doesn't later.

JunTaoLuo:
Yup feel free to comment further on this PR and I'll try to respond when I can.

PR #606, opened 2017-10-16
Put ApplicationInsights.HostingStartup back into metapackage

pakrym:


muratg:
cc @glennc FYI

JunTaoLuo:
Do we want to continue marking this as privateassets=none? If not we need to change this as well https://github.com/aspnet/Universe/blob/rel/2.0.3/build/tasks/AddMetapackageReferences.cs#L57

pakrym:
At least in patch we do.

natemcmaster:
Everything in 2.0.0 was PrivateAssets=None, so we should leave this as is to maintain compatibility.

PR #613, opened 2017-10-19
Fix rpm packages

JunTaoLuo:
3 issues:
1 - Need to add dependency on 2.0.0 runtime store for RPM installers
2 - Hosting dependencies weren't being set corrrectly
3 - typo rhel.rh.7 => rh.rhel.7


JunTaoLuo:
Verified debian installers are fine

JunTaoLuo:
WIP - need to fix dotnet-hosting tar.gz to include 2.0.0 store as well

JunTaoLuo:
Merged

PR #615, opened 2017-10-20
Generate the PackageVersionProps file and put it in the lineup package

natemcmaster:
Depends on https://github.com/aspnet/BuildTools/pull/432

This is the approached to setting package versions that we've adopted in the rel/2.0.3 branch. Instead of PackageLineup, PackageReferences are available as MSBuild variables.

### DIFF
--- a/build/RepositoryBuild.targets
+++ b/build/RepositoryBuild.targets
@@ -32,6 +32,7 @@
       <RepositoryBuildArguments>$(RepositoryBuildArguments) '/p:Universe_LineupBuildDir=$(LineupBuildDir)'</RepositoryBuildArguments>
       <RepositoryBuildArguments>$(RepositoryBuildArguments) '/p:Universe_AdditionalRestoreSources=$(IntermediateMirrorPackageDir)'</RepositoryBuildArguments>
       <RepositoryBuildArguments>$(RepositoryBuildArguments) '/p:Universe_IntermediateDir=$(IntermediateDir)'</RepositoryBuildArguments>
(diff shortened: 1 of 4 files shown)


natemcmaster:
Example output: https://gist.github.com/natemcmaster/f8b94e5d78cd4a406d0a396dccbb1f4d

ryanbrandenburg:
So the workflow here would be:

1. Clone repo
2. Run `.\build.cmd` to get a dependencies.props file
3. Run your build as normal, it won't be changing package versions until you clean.
4. If you want to use a different version locally, edit the file.
5. If you want to update the version permanently you make a change in Universe?

natemcmaster:
Slight tweak. Our repos will have build/dependencies.props files too.

The desired workflow will be

1. clone repo
2. build.cmd works 

Also works

1. clone repo
2. open VS and it works (no more `build /t:Pin`).

Upgrade workflow:

 - you can do it manually 
 - you can run `run.ps1 upgrade-deps`. It will pull the latest lineup package and attempt to update your repo's dependencies.props file

natemcmaster:
Commit noise is an unfortunate downside of making builds deterministic and reproducible, but we'll adjust our automation to keep it balanced.

ryanbrandenburg:
Yeah, I would much rather have commit noise than build indeterminacy.

natemcmaster:
Merged https://github.com/aspnet/Universe/commit/e91af13a7de7b7df348bb3b860f593cff344fdb0

Issue #577, opened 2017-10-04
Automate dependency updates


natemcmaster:
Because we plan to start committing repo hashes, build tool versions, and have already started committing exact versions of dependencies, we should make it easier to automate updates to these.

Workflow:
 - Run a scheduled or trigger-based update script
 - Ensure build still passes
 - Commit and push changes

cref #576 , #574 , #566

ryanbrandenburg:
As per conversation in person, we would want this to work with either a NuGet feed or ProdCon as the "source of truth".

natemcmaster:
cc @pranavkm

natemcmaster:
@ryanbrandenburg I have a bunch of information now on how we can automate this now that prodcon is stabilizing. Is this work item one you still want to own?

ryanbrandenburg:
Yeah, let's swap info about how exactly we want this to function and I'll drive it.

ryanbrandenburg:
Some areas for to look at when doing this work:

- https://github.com/dotnet/versions/blob/cdce4ee1842eca91db64a1531c6fc4458c3e0bcf/Maestro/subscriptions.json#L559-L580
- https://github.com/dotnet/versions/blob/master/build-info/dotnet/product/cli/release/2.1/build.xml
- https://github.com/dotnet/corefxlab/blob/master/scripts/update-dependencies/update-dependencies.csproj

ryanbrandenburg:
This work is "done" but we haven't seen a run of it in Maestro because ProdCon builds have not gotten to us in a while. We won't be able to learn more on this until they are flowing again, so it's blocked.

ryanbrandenburg:
We successfully ran this script and merged the PR it created.

Issue #575, opened 2017-10-04
Reproducible builds


natemcmaster:
Tracking issue for multiple issue related to making the build reproducible.

 - Commit the exact sha of repositories - https://github.com/aspnet/Universe/issues/574
 - Commit the version of build tools - https://github.com/aspnet/Universe/issues/576
 - Forbid floating versions in PackageReference - https://github.com/aspnet/BuildTools/issues/398

Pinning versions in repos:

- [x] Antiforgery - https://github.com/aspnet/Antiforgery/commit/df486e54a8fc7e68b75b28d9076b5a9f1c29f49e
- [x] AzureIntegration
- [x] BasicMiddleware - https://github.com/aspnet/BasicMiddleware/commit/76af959e99a1e509bd596393a592f177a0b74740
- [x] BrowserLink - https://github.com/aspnet/BrowserLink/commit/87c6dad76408830f31496d4ff19f951a0ff8691c
- [x] Caching - https://github.com/aspnet/Caching/commit/c6a9cbbe4566d1d757b27d74a1aad5da8d12aed1
- [x] Common - https://github.com/aspnet/Common/pull/272
- [x] Configuration - https://github.com/aspnet/Configuration/pull/749
- [x] CORS -https://github.com/aspnet/CORS/commit/8fd36866f9427b8a09cecdd6520ba54b064424c6
- [x] DataProtection https://github.com/aspnet/DataProtection/commit/49b2e22ab30239ed63d3f3b9478e7563e9a93ed9
- [x] DependencyInjection https://github.com/aspnet/DependencyInjection/commit/88297e3c1116f952aa22013172ef2f9396f716e0
- [x] Diagnostics - https://github.com/aspnet/Diagnostics/commit/7ac0e06abf822738437dba3bfc5b3b01325045f7
- [x] DotNetTools - https://github.com/aspnet/DotNetTools/commit/1836201beecf81c3e2b4a50cc1666f5aa51f5786
- [x] EntityFrameworkCore - https://github.com/aspnet/EntityFrameworkCore/pull/10199
- [x] EventNotification https://github.com/aspnet/EventNotification/commit/3f881f18d811b65046c1df798d4d97c98aa39086
- [x] FileSystem
- [x] Hosting - https://github.com/aspnet/Hosting/pull/1252
- [x] HtmlAbstractions
- [x] HttpAbstractions
- [x] HttpClientFactory
- [x] HttpSysServer
[...]

PR #618, opened 2017-10-24
Add scripts to auto-update build tools and submodules

natemcmaster:
Part of https://github.com/aspnet/Universe/issues/577

These scripts will run on CI and update to new versions of submodules and build tools.

CI flow:

1. Run the ./script/Update*.ps1 script and save changes to local copy
2. Run some kind of test to ensure things still build. Examples: run the compilation step, ensure compile passes, etc.
3. If the test passes, `git push` updates


scripts/UpdateBuildTools.ps1
```diff
@@ -0,0 +1,86 @@
...
+    return $version
+}
+
+Push-Location $RepoRoot
+try {
+    Ensure-Git
+
+    $oldVersion = Get-KoreBuildVersion
+
+    & "$RepoRoot/run.ps1" -Update default-build '/t:Cow'
```
ryanbrandenburg: Isn't running default-build with `/t:Cow` the same as running `install-tools`? If not I feel like it's worth clarifying what all we expect this to do.

scripts/UpdateBuildTools.ps1
```diff
@@ -0,0 +1,86 @@
...
+    | % {
+        if ($_ -like '*<InternalAspNetCoreSdkPackageVersion>*') {
+            "    <InternalAspNetCoreSdkPackageVersion>$newVersion</InternalAspNetCoreSdkPackageVersion>"
+        } else {
+            $_
+        }
+    }
+    $deps | Set-Content -Encoding UTF8 "$RepoRoot/build/dependencies.props"
+
+    & git add "$RepoRoot/korebuild-lock.txt"
```
ryanbrandenburg: Ah, I see, having accepted our fates to have to update `build/dependencies.props` constantly we're also now committing `korebuild-lock.txt`.

scripts/UpdateSubmodules.ps1
```diff
@@ -0,0 +1,111 @@
...
+        }
+        finally {
+            Pop-Location
+        }
+    } `
+    | ? { $_.changed } `
+    | % { "$($_.module) to $($_.newCommit.Substring(0, 8))" }
+
+    if ($changes) {
+        $message = "Updating submodule $( $changes -join ' ,' )"
```
ryanbrandenburg: `submodule(s)`?

scripts/UpdateBuildTools.ps1
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env powershell
```
ryanbrandenburg: We expect this script and the one below to mostly only be run by the CI, correct?

natemcmaster: Yes

scripts/UpdateBuildTools.ps1
natemcmaster: Yeah. I just need some kind of no-op command so the korebuild-lock.txt file gets updated. I'll add a comment.

scripts/UpdateBuildTools.ps1
natemcmaster: Yup. #575 requires it.

PR #620, opened 2017-10-25
Use TeamCity to select commits

ryanbrandenburg:
My attempt to integrate this script with TeamCity's VCS settings.


scripts/UpdateSubmodules.ps1
```diff
@@ -50,13 +50,22 @@ try {
...
+            if ($commit == $null)
+            {
+                $commit = $(git rev-parse HEAD)
+            }
+
             $data = @{
                 path      = $_
                 module    = $_.Name
-                commit    = $(git rev-parse HEAD)
+                commit    = $commit
```
natemcmaster: $data.commit is meant to represent the current commit. What we're actualy interested in getting from TC is the newCommit value (See the next loop down)

ryanbrandenburg:
:up::date:

PR #621, opened 2017-10-26
Use submodules for source dependencies

### LINKED ISSUE #574: Commit the exact sha of repositories
Currently, the build clones sub repositories by pulling a branch. We want to switch to having the exact sha committed to the source in universe.

Justification:
 - reproducible builds. 
   - Make it easier for external contributors to build
   - Make it easier for us to build patches
 - synchronizing changes. Make it easier to synchronize a change in sub repos with required changes in aspnet/Universe

Implementation plan:
 - As a first step, we'll try using git submodules. 
 - If this causes unanticipated problems, we'll look into using some other mechanism.

natemcmaster:
Resolves #574

As a part of making builds more reproducible, we want to start committing the exact sha of source dependencies used to build aspnetcore. This implements that using git submodules, and removes targets that were previously used to clone source on demand.

FYI - this is the same change we made to the 2.0.x branch. See https://github.com/aspnet/Universe/pull/592

cc @anurse @pakrym  - you may interested, per our discussion in person today


natemcmaster:
:watch: before merging, I need to update submodules. The dev branch of repos has moved ahead

scripts/UpdateSubmodules.ps1
```diff
@@ -26,6 +26,7 @@ $ModuleDirectory = Join-Path $RepoRoot "modules"
 Import-Module "$PSScriptRoot/common.psm1" -Scope Local -Force
 
 function Get-GitChanges([string]$Path) {
+    Write-Verbose "Get-GitChanges"
```
ryanbrandenburg: Is this output intended or left over?

scripts/UpdateSubmodules.ps1
```diff
@@ -36,65 +37,72 @@ function Get-GitChanges([string]$Path) {
     return $LastExitCode -ne 0
 }
 
+Push-Location $RepoRoot | Out-Null
```
ryanbrandenburg: I didn't notice that these were producing output.

scripts/UpdateSubmodules.ps1
```diff
@@ -36,65 +37,72 @@ function Get-GitChanges([string]$Path) {
...
-    $submodules = Get-Submodules $ModuleDirectory
+    $submodules = Get-Submodules $RepoRoot -Verbose:$VerbosePreference
 
-    $changes = $submodules `
-    | % {
-        Push-Location $_.path
-        try {
-            $vcs_name = "BUILD_VCS_NUMBER_" + $_.module
-            $newCommit = [environment]::GetEnvironmentVariable($vcs_name)
+    foreach ($submodule in  $submodules) {
```
ryanbrandenburg: Yeah, I think the `foreach` pattern is a lot more readable.

scripts/common.psm1
```diff
@@ -16,23 +16,31 @@ function Invoke-Block([scriptblock]$cmd) {
...
-{
-    Invoke-Block { & git submodule update --init }
+function Get-Submodules {
+    param(
+        [Parameter(Mandatory = $true)]
+        [string]$RepoRoot
+    )
 
-    $gitModules = Join-Path $RepoRoot ".gitmodules"
+    if ((Get-ChildItem "$RepoRoot/modules/*" -Directory | Measure-Object) -eq 0) {
```
ryanbrandenburg: So for each subdirectory in `$RepoRoot/modules` which is empty  we update it?

scripts/common.psm1
```diff
@@ -16,23 +16,31 @@ function Invoke-Block([scriptblock]$cmd) {
...
+        Write-Verbose "git submodule update --init"
+        Invoke-Block { & git submodule update --init }
```
ryanbrandenburg: Nit: might be nice if the contents of this block and Write-Verbose were driven by the same variable to prevent drift.

scripts/UpdateSubmodules.ps1
natemcmaster: left over. The script was hanging for 30 seconds with no output and I was trying to figure out what was causing it

scripts/common.psm1
natemcmaster: actually, tested this more and doesn't work. `git submodules update --init` takes 30-40 seconds. I'm trying to skip it if necessary. I'm going to remove this though as it's possible to have only some git submodules checked out

PR #623, opened 2017-10-26
Add script to update each Repo

ryanbrandenburg:
This is my first stab at a script to update individual repos. I have not fully tested it since I'm trying to set up a safe way to do so on my local machine, but I thought I'd put it out there in case I've gone way off base.


scripts/UpdateRepos.ps1
```diff
@@ -0,0 +1,67 @@
...
+$ErrorActionPreference = 'Stop'
+Set-StrictMode -Version 2
+
+Import-Module "$PSScriptRoot/common.psm1" -Scope Local -Force
+
+$RepoRoot = Resolve-Path "$PSScriptRoot\.."
+$ModuleDirectory = Join-Path $RepoRoot "modules"
+
+try {
+    Push-Location $ModuleDirectory
```
natemcmaster: Lift this above the 'try'. If it were to fail, the Pop-Location would move you to the wrong place.

scripts/UpdateRepos.ps1
```diff
@@ -0,0 +1,67 @@
...
+    $submodules = Get-Submodules $ModuleDirectory
+
+    $submodules `
+    | % {
+        try {
+            Push-Location $_.path
+
+            $depsFile = Join-Path (Join-Path $($_.path) "build") "dependencies.props"
+            $gitModules = Join-Path $RepoRoot ".gitmodules"
+            # Move to latest commit on tracked branch
```
natemcmaster: 👍

scripts/UpdateRepos.ps1
```diff
@@ -0,0 +1,67 @@
...
+            $branch = $(git config -f $gitModules --get submodule.modules/$($_.module).branch )
+            Write-Output "Branch: $branch"
+
+            Invoke-Block { & git checkout --quiet $branch }
+
+            Invoke-Block { & .\run.ps1 upgrade deps --source $Source --id $LineupID --version $LineupVersion --deps-file $depsFile }
+            Invoke-Block { & git add $depsFile }
+            Invoke-Block { & git commit --quiet --author "ASP.NET Push Bot <asplab@microsoft.com>" -m "[auto]Update dependencies.props" }
```
natemcmaster: Make the commit author an optional parameter of the script. Example: https://github.com/aspnet/BuildTools/blob/dev/scripts/UpdateDependencyVersions.ps1#L50

natemcmaster: Can you change the message to be consistent with the auto-updated tag we've added in the other universe update scripts?
```
-m "Update dependencies.props`n`n[auto-updated: dependencies]"
```

scripts/UpdateRepos.ps1
```diff
@@ -0,0 +1,66 @@
...
+            $sshUrl = "git@github.com:aspnet/$($_.module)"
+            Invoke-Block { & git remote set-url --push origin $sshUrl }
+
+            # TODO: How to push commits?
```
natemcmaster: Like this?
```
git push origin $_.branch
```

scripts/UpdateRepos.ps1
```diff
@@ -0,0 +1,66 @@
...
+        }
```
natemcmaster: I'd recommend doing push after you confirmed that you can make changes to the deps.props file successfully. Something like this:

```ps1
$update_errors = @()
foreach ($sm in $submodules) {
   try {
       # (upgrade-dependencies-props)
   }
   catch {
      $update_errors += $_
   }
}

if ($update_errors.Count -gt 0 ) {
  throw 'failed to update'
}

$push_errors = @()
foreach ($sm in $submodules) {
   try {
       # git-push
   }
   catch {
      $push_errors += $_
   }
}
```

ryanbrandenburg:
:up::date:

scripts/UpdateRepos.ps1
```diff
@@ -0,0 +1,87 @@
...
+            Pop-Location
+        }
+    }
+
+    if ($update_errors.Count -gt 0 )
+    {
+        throw 'Failed to update'
+    }
+
+    $push_errors = @()
```
natemcmaster: Also assert this is empty at the end of the script

scripts/UpdateRepos.ps1
```diff
@@ -0,0 +1,87 @@
...
+    # Init all submodules
+    Invoke-Block { & git submodule update --init }
+
+    $update_errors = @()
+    $submodules = Get-Submodules $ModuleDirectory
+    foreach($submodule in $submodules)
+    {
+        Push-Location $submodule.path
+        try {
+            $depsFile = Join-Path (Join-Path $($submodule.path) "build") "dependencies.props"
```
natemcmaster: We should skip the update if this file doesn't exist

ryanbrandenburg:
:up::date:

scripts/UpdateRepos.ps1
ryanbrandenburg: Addressed.

scripts/UpdateRepos.ps1
ryanbrandenburg: Addressed.

PR #625, opened 2017-10-30
Add a switch to protect from unintended pushes

ryanbrandenburg:


scripts/UpdateRepos.ps1
```diff
@@ -18,6 +18,7 @@ param(
     [string]$LineupID,
     [Parameter(Mandatory=$true)]
     [string]$LineupVersion,
+    [switch]$NoPush,
```
natemcmaster: Nit: update the doc block.

PR #626, opened 2017-11-01
Temporarily remove SpaServices.Extensions as it will be built separately for OOB shipping

SteveSandersonMS:


SteveSandersonMS:
Merging now to fix build.

PR #627, opened 2017-11-01
Updates to UpdateRepos.ps1

ryanbrandenburg:


scripts/UpdateRepos.ps1
```diff
@@ -46,25 +48,27 @@ try {
         try {
             $depsFile = Join-Path (Join-Path $($submodule.path) "build") "dependencies.props"
 
-            if (!Test-Path $depsFile)
+            # Move to latest commit on tracked branch
+            Invoke-Block { & git checkout --quiet $submodule.branch }
```
natemcmaster: hmm, this raises an interesting question. What to do if the dev branch of a module is ahead of the version of universe we're updating too....not sure of the right answer, but my hope was that we would do this.

```
*  Universe  xyx1  - Build + signing passes => Internal.AspNetCore.Universe.Lineup 2.1.0-preview1-xyz1
| \
|  *  foreach mod in modules/* => run.ps1 upgrade deps --version 2.1.0-preview1-xyz1
|  |
|  *  foreach mod in modules/* => git commit && git push
| /
* Universe -- git add modules/* && git commit -m "Updated submodules"
```

ryanbrandenburg: Maybe I was trying to make this too resilient should we try it your way and see if we run into any problems, or do we need to address a commit sneaking in the middle?

natemcmaster: We can use TC's shared resources locks to avoid a commit to Universe sneaking in from other automated builds. Manual commits to Universe should be rare so I think commits sneaking into universe should be rare.

natemcmaster: As far as commits the repos themselves though, yeah, we'll likely hit that now and then. Not sure we have a good way to handle this except to update universe to a new build and retry.

ryanbrandenburg:
:up::date:

scripts/UpdateRepos.ps1
```diff
@@ -40,78 +42,78 @@ $ModuleDirectory = Join-Path $RepoRoot "modules"
...
                 continue
             }
 
-            # Move to latest commit on tracked branch
-            Invoke-Block { & git checkout --quiet $submodule.branch }
-
-            Invoke-Block { & .\run.ps1 -Update upgrade deps --source $Source --id $LineupID --version $LineupVersion --deps-file $depsFile }
-            Invoke-Block { & git add $depsFile }
+            Write-Verbose "About to update dependencies.props for $($submodule.module)"
+            & .\run.ps1 -Update upgrade deps --source $Source --id $LineupID --version $LineupVersion --deps-file $depsFile | Out-Null
```
natemcmaster: I think it's useful to show the output instead of piping to `out-null`.

PR #630, opened 2017-11-07
Make UpdateRepos add dependencies.props to commits

ryanbrandenburg:


scripts/UpdateRepos.ps1
```diff
@@ -89,7 +93,12 @@ try {
 
     if ($update_errors.Count -gt 0 ) {
         foreach ($update_error in $update_errors) {
-            Write-Error "$update_error"
+            if ($update_error -eq $null) {
+                Write-Error "Error was null."
```
natemcmaster: when does this happen? seems like it never should

ryanbrandenburg: I didn't think it should either, but I ran into that case while trying it out. It seems like one of the processes exited with non-zero but didn't write to stdout.

PR #632, opened 2017-11-08
Add Force parameter to UpdateRepos

ryanbrandenburg:
Same thing we do in `UpdateSubmodules.ps1`, giving people a way to skip prompts without setting `-NoInteractive`.


natemcmaster:
Also update the doc block, then :shipit:

PR #639, opened 2017-11-09
Remove Healthchecks from Universe shipping list

anurse:
It's future in 2.1 is a little fuzzy and I don't want to accidentally ship it and lock ourselves in to supporting it.

@Eilon @muratg is this the right way to approach this? I'm nervous we'll forget that there's some work still to be done here, so I'd like to remove it sooner rather than later.


build/artifacts.props
```diff
@@ -53,7 +53,7 @@
     <PackageArtifact Include="Microsoft.AspNetCore.Diagnostics.Abstractions" Category="ship" Metapackage="true" RuntimeStore="true"/>
     <PackageArtifact Include="Microsoft.AspNetCore.Diagnostics.Elm" Category="noship" />
     <PackageArtifact Include="Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore" Category="ship" Metapackage="true" RuntimeStore="true"/>
-    <PackageArtifact Include="Microsoft.AspNetCore.Diagnostics.HealthChecks" Category="ship" Metapackage="true" RuntimeStore="true" />
+    <PackageArtifact Include="Microsoft.AspNetCore.Diagnostics.HealthChecks" Category="noship" Metapackage="true" RuntimeStore="true" />
```
natemcmaster: Noship packages cannot be in the metapackage or runtimestore. Remove those flags.

anurse: Will do, thanks for the heads-up.

PR #643, opened 2017-11-10
Update KoreBuild version

pranavkm:
FYI


natemcmaster:
I believe it was this change that is causing all the tests to fail. Can we revert? 

cc @pakrym

ryanbrandenburg:
I believe he attempted to fix that with https://github.com/aspnet/Logging/commit/7f4bcd95ba1d0c982ad42cc78426a23fd5065f39, and that commit is just now making its way to the Win2012 configs. We should learn pretty quickly if that fixed it.

pranavkm:
I'd rather we revert the test parallelization change in BuildTools (which is seemingly what's causing these failures).

ryanbrandenburg:
Yeah, now that we're having trouble with EF too I agree with @pranavkm, we can revisit test parallelization once our builds are in a more solid state.

natemcmaster:
I don't think reverting is the right move. I think we're uncovering legitimate product and test bugs. I ran EF and logging before I merged https://github.com/aspnet/BuildTools/pull/417, and they passed just fine. Before we rush to turn off parallelization, I think we should look at the errors and figure out _why_ the parallelization is bad.

For example, before merging I noticed that Kestrel always failed. The issue was port exhaustion and contention between libuv and managed sockets. We disabled parallelization there because we legitmately don't care if you can run 1000 kestrel servers in parallel at the same time on a machine.

Logging and EF on the other hand - seems like product bugs that analyzers fail if when run in parallel, and if you can't access a SQL database in parallel with other process.

pakrym:
@natemcmaster how long are we going to keep Universe broken for this? Can we make it opt in and do it repo by repo so we don't stall entire workflow?

natemcmaster:
@pakrym I'm not overly concerned with this. We're down to a handful of EF tests that fail to initialize a shared database in parallel (sometimes), and a few flaky NTLM and WebHostBuilder tests that were flaky before the parallelization change
.

PR #644, opened 2017-11-10
Add more packages required by the 2.1 runtime

pranavkm:


natemcmaster:
You'll likely end up needing all of these too:

https://github.com/aspnet/Universe/blob/b0de6da24fb726941371f1697afbf21bc7928f95/build/dependencies.props#L45-L79

pranavkm:
Ahh, I guess that's true.

PR #645, opened 2017-11-13
Add Microsoft.Extensions.ApplicationModelDetection

pakrym:


build/artifacts.props
```diff
@@ -171,6 +171,7 @@
     <PackageArtifact Include="Microsoft.EntityFrameworkCore.Tools" Category="ship" Metapackage="true" RuntimeStore="true" />
     <PackageArtifact Include="Microsoft.EntityFrameworkCore.Tools.DotNet" Category="ship" LZMATools="true" />
     <PackageArtifact Include="Microsoft.Extensions.ActivatorUtilities.Sources" Category="noship" />
+    <PackageArtifact Include="Microsoft.Extensions.ApplicationModelDetection" Category="noship" />
```
muratg: Is this temporarily "no-ship" or are we not going to use nuget.org as the distribution mechanism for this package?

PR #659, opened 2017-11-20
Add AuthSamples repo

Tratcher:
cc: @HaoK


Tratcher:
Yes, but it's hard to tell of the scripts are up to date, they change pretty quickly.

natemcmaster:
Which scripts? build.cmd/sh? You should be able to copy them from https://github.com/aspnet/BuildTools/tree/dev/scripts/bootstrapper

PR #660, opened 2017-11-20
Update ServiceController version for netstandard support

JunTaoLuo:
This is required for https://github.com/aspnet/Hosting/pull/1189

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -366,7 +366,7 @@
     <ExternalDependency Include="System.Security.Cryptography.Xml" Version="4.4.0" Source="$(DefaultNuGetFeed)" />
     <ExternalDependency Include="System.Security.Permissions" Version="4.4.0" Source="$(DefaultNuGetFeed)" />
     <ExternalDependency Include="System.Security.Principal.Windows" Version="4.4.0" Source="$(DefaultNuGetFeed)" />
-    <ExternalDependency Include="System.ServiceProcess.ServiceController" Version="4.4.0" Source="$(DefaultNuGetFeed)" />
+    <ExternalDependency Include="System.ServiceProcess.ServiceController" Version="4.5.0-preview1-25914-04" Source="$(DefaultNuGetFeed)" />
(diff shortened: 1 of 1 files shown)


build/dependencies.props
```diff
@@ -366,7 +366,7 @@
     <ExternalDependency Include="System.Security.Cryptography.Xml" Version="4.4.0" Source="$(DefaultNuGetFeed)" />
     <ExternalDependency Include="System.Security.Permissions" Version="4.4.0" Source="$(DefaultNuGetFeed)" />
     <ExternalDependency Include="System.Security.Principal.Windows" Version="4.4.0" Source="$(DefaultNuGetFeed)" />
-    <ExternalDependency Include="System.ServiceProcess.ServiceController" Version="4.4.0" Source="$(DefaultNuGetFeed)" />
+    <ExternalDependency Include="System.ServiceProcess.ServiceController" Version="4.5.0-preview1-25914-04" Source="$(DefaultNuGetFeed)" />
```
pranavkm: This comes from dotnet-core feed (not `DefaultNuGetFeed`). Also needs to be mirrored.

pranavkm: While we're at this, should we do a pass through repos and bump them up to use the latest CoreFx packages too? We updated the SDK \ runtime, presumably packages should be next. @Eilon \ @muratg thoughts?

JunTaoLuo: This one is actually on nuget: https://www.nuget.org/packages/System.ServiceProcess.ServiceController/4.5.0-preview1-25914-04

muratg: @pranavkm yes, I think we should bump the packages as well.

natemcmaster: @pranavkm chat with @ryanbrandenburg about #577. We need to make sure we make these version updates work well with Orchestrated Build.

pranavkm: Sure. But I was just hoping to do a one time update for these versions. Presumably the orchestrated build versions would be the latest (4.5.0-*) versions and not the shipped packages.

PR #664, opened 2017-11-21
Push a repo if it's build passed

ryanbrandenburg:
Rather than waiting for ALL repos tests to pass, let's push a repo as soon as it's own tests pass.


scripts/UpdateRepos.ps1
```diff
@@ -84,60 +80,46 @@ try {
...
+                    Write-Warning "Error in $($submodule.module): $_"
+                    $build_errors += @{
+                        Repo    = $submodule.module
+                        Message = $_
+                    }
+                    continue
+                }
+
+                # Push the changes
+                $shortMessage = "Pushing updates to repos."
```
natemcmaster: Not sure why this needs to be a variable

Issue #655, opened 2017-11-15
Produce only one version of assets


natemcmaster:
Instead of producing two on each build, only produce one version of packages, templates, runtime store, etc.

PR #665, opened 2017-11-21
Only produce one version of the runtime store, templates packages, and installers

natemcmaster:
Resolves https://github.com/aspnet/Universe/issues/655


build/Templating.targets
```diff
@@ -5,26 +5,16 @@
...
         RepositoryRoot=$(TemplatingProjectRoot);
         DotNetRestoreSourcesPropsPath=$(GeneratedRestoreSourcesPropsPath);
+        DotNetPackageVersionPropsPath=$(GeneratedPackageVersionPropsPath);
         BuildNumber=$(BuildNumber);
         Configuration=$(Configuration);
-      </TemplateProjCommmonProperties>
-      <TemplateProjProperties>
-        $(TemplateProjCommmonProperties);
-        DotNetPackageVersionPropsPath=$(GeneratedPackageVersionPropsPath);
-        VersionMetadata=timestamped;
```
JunTaoLuo: Just making sure VersionMetadata isn't needed anymore?

natemcmaster: Correct. Not needed.

natemcmaster:
Even if you're going to delete the archive and installer targets, I'm still going to merge so we can work out any other bugs we may have introduced in this change....separate from the upcoming "platforms" work.

PR #674, opened 2017-11-28
AspNetCore Shared Framework archives and installers

JunTaoLuo:


build/PlatformInstaller.targets
```diff
@@ -1,31 +1,31 @@
 <Project>
+  <Import Project="common.props" />
```
JunTaoLuo: Needed for `CoreSetupPackageVersion`. Maybe there's a better way to do this?

build/PlatformInstaller.targets
```diff
@@ -231,26 +151,26 @@
     <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="_BuildDockerImage" Properties="Image=$(Image)" />
 
     <!-- Common installer parameters -->
-    <JsonPeek Query="$.short_description" JsonInputPath="$(RSDebConfigFile)">
-      <Output TaskParameter="Result" ItemName="_RSInstallerSummary" />
+    <JsonPeek Query="$.short_description" JsonInputPath="$(SFDebConfigFile)">
```
JunTaoLuo: This is rather ugly, I'm going to create a props file that contains most of this information and populate config files instead. This way the truth lies in msbuild files instead of json files.

build/Platform.targets
```diff
@@ -0,0 +1,156 @@
+<Project>
+  <Import Project="common.props" />
```
JunTaoLuo: This feels odd and we should restructure the targets so they don't have to explicitly import props. Here, I needed to use a few version properties that were in dependency.props.

build/Platform.targets
```diff
@@ -0,0 +1,156 @@
...
+    <Copy SourceFiles="$(_SrcDir)Directory.Build.props" DestinationFolder="$(_WorkRoot)" />
+
+    <!-- Set _Target=Restore so the project will be re-evaluated to include Internal.AspNetCore.Sdk MSBuild properties on the next step. -->
+    <MSBuild Projects="$(PlatformWorkDirectory)Platform.csproj"
+      Targets="Restore"
+      Properties="$(_PlatformProps);MicrosoftAspNetCoreAllVersion=$(PackageVersion);_Target=Restore" />
+
+    <!-- Publish -->
+    <MSBuild Projects="$(PlatformWorkDirectory)Platform.csproj"
+      Targets="Publish"
```
JunTaoLuo: Still verifying a few properties with the dotnet teams. The produced output is able to run built and published apps but I see a few anomalies in the deps files and some versioning oddities. I'll file followup issues.

build/tasks/ProcessSharedFrameworkDeps.cs
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
JunTaoLuo: A few targets were copied from https://github.com/dotnet/core-setup/tree/be8d8e3486b2bf598ed69d39b1629a24caaba45e/tools-local/tasks. We'll need to keep these files in sync. Should we look for ways to share this code potentially?

src/Microsoft.AspNetCore.All/Microsoft.AspNetCore.All.csproj
```diff
@@ -24,8 +24,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Content Include="build\**\*.xml" PackagePath="%(Identity)" />
-    <Content Include="build\**\*.targets" PackagePath="%(Identity)" />
```
JunTaoLuo: We will no longer uses these targets. Projects will trim by setting the shared framework via `MicrosoftNETPlatformLibrary`. I'll need to see if we can conditionally set this automatically when > 2.1 versions of the metapackage is used. Otherwise the user needs to set this in their project file.

build/PlatformInstaller.targets
```diff
@@ -0,0 +1,298 @@
...
+    <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="_GenerateDeb" Properties="$(CommonProps);$(DebHostingProps)" />
+
+    <!-- Remove Docker Image to save disk space -->
+    <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="_RemoveDockerImage" Properties="Image=$(Image)" />
+  </Target>
+
+  <Target Name="GenerateDebs" DependsOnTargets="_EnsureInstallerPrerequisites">
+    <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="_GenerateDebOnPlatform" Properties="Version=$(Version);Image=debian.8" />
+    <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="_GenerateDebOnPlatform" Properties="Version=$(Version);Image=ubuntu.14.04" />
+    <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="_GenerateDebOnPlatform" Properties="Version=$(Version);Image=ubuntu.16.04" />
```
JunTaoLuo: I need to check which platform we need to build for in 2.1. Copying the current platforms minus the 16.10 which we know will no longer be supported.

build/PlatformInstaller.targets
```diff
@@ -0,0 +1,298 @@
...
+      <RPMVendor>.NET Foundation</RPMVendor>
+      <RHInstallRoot>/opt/rh/rh-dotnet20/root/usr/lib64/dotnet</RHInstallRoot>
+    </PropertyGroup>
+
+    <!-- Build Docker Image -->
+    <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="_BuildDockerImage" Properties="Image=$(Image)" />
+
+    <ItemGroup>
+      <RPMHostingDependencies Include="$(SFInstallerName)-$(PackageVersion)" Version="$(PackageVersion)" />
+      <RPMHostingDependencies Include="$(DotnetRuntimeInstallerName)" Version="$(CoreSetupPackageVersion)" />
```
JunTaoLuo: TODO: aspnetcore-shared installer needs to take a dependency on the previous 2.0.3 aspnetcore-store installer until "trampoline" is implemented for 2.1 preview2

JunTaoLuo:
Ready for review. Not all features are complete but this will produce artifacts that can be used for preliminary testing of downstream features.

build/tools/templates/Platform/Platform.csproj
```diff
@@ -0,0 +1,26 @@
...
+  <Import Project="$(DotNetRestoreSourcePropsPath)" Condition="'$(DotNetRestoreSourcePropsPath)' != ''" />
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
+    <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
+    <RestoreSources>$(DotNetRestoreSources)</RestoreSources>
+    <RestoreSources Condition="'$(DotNetBuildOffline)' != 'true' AND '$(AspNetUniverseBuildOffline)' != 'true' ">
+      $(RestoreSources);
+      https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json;
```
natemcmaster: This feed is obsolete. Use aspnetcore-dev instead.

src/Microsoft.AspNetCore.All/Microsoft.AspNetCore.All.csproj
natemcmaster: Since the `dev` branch is currently targeting 2.1, it seems like we should add this to the package:

```xml
<!-- build/netcoreapp2.1/Microsoft.AspNetCore.All.props -->
<PropertyGroup>
  <MicrosoftNETPlatformLibrary>Microsoft.AspNetCore.All</MicrosoftNETPlatformLibrary>
</PropertyGroup>
```

For netcoreapp2.0 projects, we'll have to figure something else out so users can migrate from the RS to shared fx.

build/tools/templates/Platform/Platform.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="$(DotNetRestoreSourcePropsPath)" Condition="'$(DotNetRestoreSourcePropsPath)' != ''" />
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
```
natemcmaster: netcoreapp2.1

muratg:
cc @DamianEdwards in case you're interested to take a look.

build/tools/packaging/debian_config.json
```diff
@@ -1,17 +1,17 @@
 {
-    "maintainer_name": "Microsoft",
-    "maintainer_email": "nugetaspnet@microsoft.com",
+    "maintainer_name": "MAINTAINER_NAME",
```
natemcmaster: Consider renaming this file to something other than `.json` (it's not a .json file). In other repos, we use `.in` as  way to indicate when a file is a template. e.g. `debian_config.json.in`.

build/tasks/RepoTasks.csproj
```diff
@@ -7,6 +7,7 @@
 
   <ItemGroup>
     <PackageReference Include="NuGet.Build.Tasks" Version="$(NuGetInMSBuildVersion)" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="2.0.0" PrivateAssets="All" />
```
natemcmaster: `Version="$(MicrosoftExtensionsDependencyModelPackageVersion)"`

build/Platform.targets
```diff
@@ -0,0 +1,156 @@
...
+      <RuntimeGraphGeneratorRuntime Condition="$([MSBuild]::IsOSPlatform('Windows'))">win</RuntimeGraphGeneratorRuntime>
+      <RuntimeGraphGeneratorRuntime Condition="$([MSBuild]::IsOSPlatform('Linux'))">linux</RuntimeGraphGeneratorRuntime>
+      <RuntimeGraphGeneratorRuntime Condition="$([MSBuild]::IsOSPlatform('OSX'))">osx</RuntimeGraphGeneratorRuntime>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <PlatformAssetsFile Include="$(PlatformWorkDirectory)**\project.assets.json" />
+    </ItemGroup>
+
+    <ProcessSharedFrameworkDeps
```
natemcmaster: Does the SDK's `GenerateDepsFile` task not work for us? Seems like we're duplicating effort.

https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/GenerateDepsFile.cs

build/Platform.targets
```diff
@@ -0,0 +1,156 @@
...
+    <MSBuild Projects="$(MetapackageWorkDirectory)Microsoft.AspNetCore.All.csproj"
+      Targets="Restore"
+      Properties="Configuration=$(Configuration);DotNetRestoreSourcePropsPath=$(GeneratedRestoreSourcesPropsPath);AspNetUniverseBuildOffline=true;_Target=Restore" />
+
+    <!-- Pack -->
+    <MSBuild Projects="$(MetapackageWorkDirectory)Microsoft.AspNetCore.All.csproj"
+      Targets="Pack"
+      Properties="Configuration=$(Configuration);DotNetRestoreSourcePropsPath=$(GeneratedRestoreSourcesPropsPath);AspNetUniverseBuildOffline=true" />
+
+    <!-- Copy to output directory -->
```
natemcmaster: It would be better to add the property `PackageOutputPath=$(BuildDir)` when calling the Pack target instead of globbing and copying files. Globbing might pickup files we don't actually want copied to the artifacts directory.

build/Platform.targets
```diff
@@ -0,0 +1,156 @@
...
+    <ItemGroup>
+      <AllMetapackageNupkgFile Include="$(MetapackageWorkDirectory)**\*.nupkg" />
+    </ItemGroup>
+
+    <Copy SourceFiles="@(AllMetapackageNupkgFile)" DestinationFolder="$(BuildDir)" />
+  </Target>
+
+  <Target Name="GetLatestCommitHash">
+    <!-- Get the latest commit hash -->
+    <Exec Command="git rev-parse HEAD 2>&amp;1" StandardOutputImportance="Low" IgnoreExitCode="true" IgnoreStandardErrorWarningFormat="true" ConsoleToMSBuild="true">
```
natemcmaster: Don't invoke git.exe. Use the `GetGitCommitHash` task instead or just add the `ResolveCommitHash` target as a dependency.

Example: https://github.com/aspnet/BuildTools/blob/cba083412c4668cfa74fd056e7152f410eac9c60/modules/BuildTools.Tasks/module.targets#L13-L17

build/Platform.targets
```diff
@@ -0,0 +1,156 @@
...
+    <!-- Clear working directory -->
+    <RemoveDir Directories="$(_WorkRoot)" />
+
+    <PropertyGroup>
+      <PlatformWorkDirectory>$(_WorkRoot)Platform\</PlatformWorkDirectory>
+      <PlatformOutputDirectory>$(_WorkRoot)Publish\</PlatformOutputDirectory>
+      <PlatformPublishDirectory>$(PlatformOutputDirectory)shared\Microsoft.AspNetCore.All\$(PackageVersion)\</PlatformPublishDirectory>
+
+      <_PlatformProps>Configuration=$(Configuration);RuntimeIdentifier=$(PlatformRID)</_PlatformProps>
+      <_PlatformProps>$(_PlatformProps);RuntimeFrameworkVersion=$(MicrosoftNETCoreApp20PackageVersion)</_PlatformProps>
```
natemcmaster: MicrosoftNETCoreApp21PackageVersion

build/Platform.targets
```diff
@@ -0,0 +1,156 @@
...
+      Targets="Publish"
+      Properties="$(_PlatformProps);GenerateRuntimeConfigurationFiles=true;SelfContained=false;PublishDir=$(PlatformPublishDirectory)" />
+
+    <!-- Clean deps.json -->
+    <RepoTasks.TrimDeps DepsFiles="$(PlatformPublishDirectory)/Platform.deps.json" />
+
+    <!-- Clean up artifacts that publish generates which we don't need -->
+    <ItemGroup>
+      <ToDelete Include="$(PlatformPublishDirectory)\Platform" />
+      <ToDelete Include="$(PlatformPublishDirectory)\Platform.dll" />
```
natemcmaster: Have you tried disabling the compilation step of this file altogether? I'm concerned this assembly will leak into deps.json files, nuspecs, or something else.

build/Platform.targets
```diff
@@ -0,0 +1,156 @@
...
+    <Move SourceFiles="$(PlatformPublishDirectory)\Platform.deps.json"
+          DestinationFiles="$(PlatformPublishDirectory)\Microsoft.AspNetCore.All.deps.json" />
+
+    <!-- Rename runtimeconfig.json file -->
+    <Move SourceFiles="$(PlatformPublishDirectory)\Platform.runtimeconfig.json"
+          DestinationFiles="$(PlatformPublishDirectory)\Microsoft.AspNetCore.All.runtimeconfig.json" />
+
+    <!-- Generate Runtime Graph -->
+    <PropertyGroup>
+      <RuntimeGraphGeneratorRuntime Condition="$([MSBuild]::IsOSPlatform('Windows'))">win</RuntimeGraphGeneratorRuntime>
```
natemcmaster: Other than crossgen, is there any reason I can't build and gather the Linux and macOS shared platforms on a Windows machine?

build/Platform.targets
```diff
@@ -0,0 +1,156 @@
...
+    </PropertyGroup>
+
+    <ItemGroup>
+      <PlatformAssetsFile Include="$(PlatformWorkDirectory)**\project.assets.json" />
+    </ItemGroup>
+
+    <ProcessSharedFrameworkDeps
+      AssetsFilePath="@(PlatformAssetsFile)"
+      DepsFilePath="$(PlatformPublishDirectory)\Microsoft.AspNetCore.All.deps.json"
+      Runtime="$(RuntimeGraphGeneratorRuntime)" />
```
natemcmaster: Why not use PlatformRID? What happens if the two don't match?

build/Platform.targets
```diff
@@ -0,0 +1,156 @@
...
+      <Output TaskParameter="ExitCode" PropertyName="LatestCommitExitCode" />
+    </Exec>
+    <!-- We shouldn't fail the build if we can't retreive the commit hash, so in this case just set it to N/A -->
+    <PropertyGroup Condition="'$(LatestCommitExitCode)'!='0'">
+      <LatestCommit>N/A</LatestCommit>
+    </PropertyGroup>
+  </Target>
+
+  <Target Name="BuildPlatform" DependsOnTargets="GetLatestCommitHash">
+    <Error Text="Please specify the RID for the aspnetcore platform via PlatformRID: {win7-x64|win7-x86|osx-x64|linux-x64}." Condition="'$(PlatformRID)' == ''"/>
```
natemcmaster: If there is a restricted set of values allowed, also add a check that PlatformRID matches one of these accepted values, in addition to the not-empty check.

build/PlatformInstaller.targets
```diff
@@ -0,0 +1,300 @@
+<Project>
+  <Target Name="BuildInstallers" DependsOnTargets="GenerateTargzs;GenerateRpms;GenerateDebs;GenerateRelabledInstallers" />
```
natemcmaster: Nit: typo. GenerateRelabledInstallers =>GenerateRelabeledInstallers

build/platform.props
```diff
@@ -0,0 +1,45 @@
...
+    <_InstallerSourceDir>$(RepositoryRoot).deps\Installers\</_InstallerSourceDir>
+    <_DockerRootDir>/opt/code/</_DockerRootDir>
+    <_InstallersOutputDir>$(ArtifactsDir)installers\</_InstallersOutputDir>
+
+    <!-- installers -->
+    <SFInstallerName>aspnetcore-shared</SFInstallerName>
+    <HostingInstallerName>dotnet-hosting</HostingInstallerName>
+    <DebConfigFile>$(_PackagingDir)debian_config.json</DebConfigFile>
+
+    <PublicCoreFeedPrefix>https://dotnetcli.blob.core.windows.net/dotnet</PublicCoreFeedPrefix>
```
natemcmaster: Need to update this to consume `DotNetAssetRootUrl` when building in orchestrated-build.

build/Platform.targets
JunTaoLuo: MicrosoftNETCoreApp21PackageVersion

build/PlatformInstaller.targets
```diff
@@ -0,0 +1,300 @@
...
+
+  <Target Name="_DownloadInstallers">
+    <!-- Download dotnet installers -->
+    <MakeDir Directories="$(_InstallerSource)" />
+    <!--
+      Note: KOREBUILD_DOTNET_FEED_CREDENTIAL is intentionally NOT an MSBuild variable.
+      MSBuild doesn't to the substitution correctly because the string contains %,
+      so we'll let bash do it instead.
+    -->
+    <Exec Command="curl --fail -sSL &quot;$(RuntimeTargzLink)$KOREBUILD_DOTNET_FEED_CREDENTIAL&quot; -o $(_InstallerSource)$(DotnetRuntimeInstallerArchiveName)" />
```
natemcmaster: Use DotNetAssetRootAccessTokenSuffix instead of KOREBUILD_DOTNET_FEED_CREDENTIAL. 

Would be better to switch to the [`DownloadFile`](https://github.com/aspnet/BuildTools/blob/dev/modules/KoreBuild.Tasks/DownloadFile.cs) task now. This has some additional protections, like timeouts, and support for `file://` URIs.

build/PlatformInstaller.targets
```diff
@@ -0,0 +1,300 @@
...
+    <ItemGroup>
+      <ChangelogItems Include="DATE" Replacement="$([System.DateTime]::UtcNow.ToString(ddd MMM dd yyyy))" />
+      <ChangelogItems Include="MAINTAINER_NAME" Replacement="$(RPMMaintainerName)" />
+      <ChangelogItems Include="MAINTAINER_EMAIL" Replacement="$(RPMMaintainerEmail)" />
+      <ChangelogItems Include="PACKAGE_VERSION" Replacement="$(RPMVersion)" />
+      <ChangelogItems Include="PACKAGE_REVISION" Replacement="$(RPMRevision)" />
+    </ItemGroup>
+
+    <!-- Update Date, Maintainer, Version, Revision and Changelog Message -->
+    <RepoTasks.ReplaceInFile Filename="$(_WorkRoot)templates/changelog" Items="@(ChangelogItems)" />
```
natemcmaster: (Can be done in cleanup later). Consider using the `GenerateFileFromTemplate` task to find/replace.

build/PlatformInstaller.targets
natemcmaster: Add this as a comment to the code and add a work item to remove it for 2.1 RTM.

build/PlatformInstaller.targets
```diff
@@ -0,0 +1,300 @@
...
+      so we'll let bash do it instead.
+    -->
+
+    <Exec Command="docker run
+      --rm
+      -v $(RepositoryRoot):$(_DockerRootDir)
+      -e DOTNET_SKIP_FIRST_TIME_EXPERIENCE=true
+      -e INSTALLER_NAME=$(DebPrefix)-$(DebVersion)
+      -e INSTALLER_VERSION=$(DebVersion)
+      -e 'KOREBUILD_DOTNET_VERSION=$(KOREBUILD_DOTNET_VERSION)'
```
natemcmaster: We plan to deprecate these variables completely. It would be better to switch to variables as defined in `dependencies.props`.

build/PlatformInstaller.targets
```diff
@@ -0,0 +1,300 @@
...
+      SkipUnchangedFiles="False"
+      UseHardlinksIfPossible="False" />
+  </Target>
+
+  <Target Name="_GenerateDebOnPlatform">
+    <!-- Build Docker Image -->
+    <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="_BuildDockerImage" Properties="Image=$(Image)" />
+
+    <ItemGroup>
+      <_DebSFDependencies Include="aspnetcore-store-2.0.3"/>
```
natemcmaster: Likewise, add a comment here why 2.1 depends on 2.0.x stuff.

Also, we should centralize the version as a variable in dependencies.props. Something like `MicrosoftAspNetCore20RuntimeStoreVersion`?

build/tasks/ProcessSharedFrameworkDeps.cs
natemcmaster: Yes, absolutely yes. Until then, add this comment into the affected files so we know where to look to update things.

build/tools/templates/Platform/Platform.csproj
JunTaoLuo: Yup, this is pretty out of date. This was what broke RAR when I rebased.

build/Platform.targets
JunTaoLuo: 👍  Didn't know we had a task for this.

build/Platform.targets
JunTaoLuo: This is trimmed from the deps files. I'm copying this logic from https://github.com/dotnet/core-setup/blob/d5e9580461cf73577f4882ce8755a038c907abfb/src/sharedFramework/sharedFramework.proj. I'd like to keep the same logic for now, at least for the first version, and gradually improve on it later.

build/Platform.targets
JunTaoLuo: Again, trying to use the same logic as core-setup. I don't have a clear understanding of the underlying mechanism. I can diff the outputs to ensure this doesn't break but I'd like to do that after we produce the first version.

build/PlatformInstaller.targets
JunTaoLuo: Just going to hardcode the versions for now. This is a temporary dependency anyway.

JunTaoLuo:
:up::date:

build/SharedFx.props
```diff
@@ -0,0 +1,44 @@
...
+    <DotnetRuntimeInstallerArchiveName>$(DotnetRuntimeInstallerName)-linux-x64.tar.gz</DotnetRuntimeInstallerArchiveName>
+    <RuntimeTargzLink>$(DotNetAssetRootUrl)/Runtime/$(MicrosoftNETCoreApp21PackageVersion)/$(DotnetRuntimeInstallerArchiveName)</RuntimeTargzLink>
+    <SharedFxLinuxArchiveFilePath>$(_SharedFxSourceDir)$(SharedFxInstallerName)-$(PackageVersion)-linux-x64.tar.gz</SharedFxLinuxArchiveFilePath>
+
+    <!-- installer metadata -->
+    <MaintainerName>Microsoft</MaintainerName>
+    <MaintainerEmail>nugetaspnet@microsoft.com</MaintainerEmail>
+    <Homepage>https://www.asp.net/</Homepage>
+    <InstallRoot>/usr/share/dotnet</InstallRoot>
+    <PackageRevision>1</PackageRevision>
```
natemcmaster: When (if ever) would this get changed?

build/SharedFx.targets
```diff
@@ -0,0 +1,142 @@
+<Project>
+  <Import Project="common.props" />
+  <Import Project="SharedFx.props" />
+
+  <Target Name="BuildMetapackage" DependsOnTargets="ResolveRepoInfo">
+    <!-- Clear working directory -->
+    <RemoveDir Directories="$(_WorkRoot)" />
+
+    <!-- Move to working dir -->
```
natemcmaster: Comment doesn't make sense here. What's moving?

build/SharedFx.props
JunTaoLuo: We will never change this really given the way we version installers. I've kept it as a property since this value is shared for both RPM and DEB installers. Could hardcode but I don't think that's necessary.

build/SharedFx.targets
JunTaoLuo: Moved comment to the copy task below. It used to make sense when the properties and itemgroups were defined for the copy task.

PR #675, opened 2017-11-28
Upgrade xunit to 2.3.1

### LINKED ISSUE #624: Update XUnit to version 2.3.1 due to bug in 2.3.0 that prohibit usage of testrunner inside VS
Is there a way that I can override the pinned xunit version? The pinned version is 2.3.0 and that cause runing tests in visual studio throwing the same error as https://github.com/xunit/xunit/issues/1514 and I should like to test with the 2.3.1 version instead to see if that solves the error. Working on an EFCore bug and it is hard to not been able to run and debug test during development. (dotnet test is still working but I lack the debug possibility)

natemcmaster:
Resolves #624


natemcmaster:
I'm running a full build right now to validate that the updated analyzers don't fail the compilation step.

natemcmaster:
Compilation passes. Can't tell if tests pass because they are failing due to issues unrelated to xunit.

PR #676, opened 2017-11-28
Add support for automatically installing new versions of the 2.1 runtime

natemcmaster:
Depends on https://github.com/aspnet/BuildTools/pull/495


natemcmaster:
Closing. We need to discuss more with core-setup how to acquire the shared runtime from the transport feed. Current efforts fail. May required core-setup to change how they layout files in Azure.

natemcmaster:
core-setup changed their transport feed layout https://github.com/dotnet/core-setup/commit/812a236404b5a47c92a94df2be34b22b524af414 so dotnet-install.ps1/sh should keep working.

natemcmaster:
:up: :date: should work now with orchestrated build.

PR #685, opened 2017-11-30
LZMA update for 2.0.4

JunTaoLuo:
I think we also need to update the config.dotnet.runtime.version parameter on the CI when we run.


build/dependencies.props
```diff
@@ -202,18 +202,18 @@
     <ExternalDependency Include="Microsoft.Web.Xdt" Version="1.4.0" Source="$(DefaultNuGetFeed)"/>
     <ExternalDependency Include="Microsoft.Win32.Registry" Version="4.4.0" Source="$(DefaultNuGetFeed)" />
     <ExternalDependency Include="Moq" Version="4.7.49" Source="$(DefaultNuGetFeed)" Private="true"/>
-    <ExternalDependency Include="NETStandard.Library" Version="2.0.0" Source="$(DefaultNuGetFeed)">
+    <ExternalDependency Include="NETStandard.Library" Version="2.0.1" Source="$(DefaultNuGetFeed)">
```
JunTaoLuo: Not sure if these are needed anymore.

JunTaoLuo:
So I should commit these changes to release/2.0.0 @natemcmaster ?

build/dependencies.props
natemcmaster: They are for the 2.0.x patch builds, at least for now.

build/tools/templates/Archive/Archive.csproj
```diff
@@ -5,6 +5,8 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp2.0</TargetFramework>
     <EnableApiCheck>false</EnableApiCheck>
+    <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
```
natemcmaster: +💯. Thanks for adding this

natemcmaster:
Yes, this should merge to release/2.0.0

build/tools/templates/Archive/Archive.csproj
JunTaoLuo: I think we should investigate whether we want to have this in other templates as well.

JunTaoLuo:
Merged in https://github.com/aspnet/Universe/commit/f7c3546b988ae2643b8b0cbdde6cb5cff3dc0e1a

PR #688, opened 2017-12-01
Add a design document for the bill of materials and incremental build

natemcmaster:
Wrote a spec for that describes my current thinking about how to do incremental builds using a bill of materials.

The motivation for this: @JunTaoLuo and I are having a difficult time producing a new LZMA without re-building the rest of aspnetcore. Part of the problem is that we have a handful of different caching mechanisms to avoid re-building parts of the stack. But these mechanisms are not consistent or easy to define. Some parts of the build can't be cached.

This design would make it possible to centralize our approach to caching build results, and to make it easier to rebuild only part of the stack.

cc @muratg @Eilon @ryanbrandenburg @pranavkm


natemcmaster:
Yes, just a spec. Definitely open to feedback.

docs/IncrementalBuild.md
```diff
@@ -0,0 +1,30 @@
...
+Each build will produce a new Bill of Materials (bom). Normally, this bom isn't added to source code right away.
+This bom should only be committed to source once the assets it describes are published and available from caching mechanisms.
+
+## Caching mechanisms
+
+These are some common caching mechanisms used to store build artifacts.
+
+ - NuGet servers. NuGet.org, MyGet.org, and file shares can be used to restore previously-built nupkg files
+ - Debian and RPM feeds. Some .deb and .rpm installers may be cached on https://packages.microsoft.com/repos.
+ - Azure blob feeds. Arbitrary buidl artifacts may be cached on https://aspnetcore.blob.core.windows.net/.
```
smitpatel: typo: build

docs/IncrementalBuild.md
```diff
@@ -0,0 +1,30 @@
+Incremental Build
+=================
+
+This repository can build the entire ASP.NET Core stack from scratch. However, in most cases, we only want to build part of the stack.
+
+## Building incrementally
```
pranavkm: I'm not very clear on what the workflow for this is (plus a little unsure as to why we're conflating the two topics here). Is the intent than an ordinary build of Universe completely no-ops if it has boms files checked in?

natemcmaster: "Incremental" is a poor word choice. I'll change this to `Building with cached artifacts`.

The workflow you described is exactly what we're after. Only build when we need to. If everything in universe is built and available from a cache, why rebuild?

docs/BillOfMaterials.md
```diff
@@ -0,0 +1,129 @@
...
+The bill of materials is an XML file. Here is a minimal example of a bom.
+
+```xml
+<Build Id="(commithash)+Release+Signed+123" Created="12/1/2017 5:10:06 PM +00:00">
+
+    <Artifacts Category="ship">
+        <Artifact Id="Microsoft.Extensions.DependencyInjection.2.0.0-rtm-123.nupkg" Type="NuGetPackage"
+            ShouldBeSigned="true"
+            PackageId="Microsoft.Extensions.DependencyInjection"
+            Version="2.0.0-rtm-123" />
```
muratg: Can we add repo name, branch name and commit hash as meta data to each package as well?

muratg: Also, based on our chat:
   - nupgk hash (so that we can do comparisons later if needed, for example while validating caches)
   - nupgk artifact location.

natemcmaster: > nupgk artifact location.

Actually, this is one piece we don't want in the bom. The BOM is just about files that were produced, but shouldn't define where to get them. The location we download them from could be private, so we don't want that commited to src.

natemcmaster: Updated the spec with examples of the other stuff though. yes, Commit hash, file hash, etc are good metadata to have about files.

docs/BillOfMaterials.md
```diff
@@ -0,0 +1,170 @@
...
+These are some common artifact types produced by the build.
+
+  - NuGetPackage
+  - NuGetSymbolsPackage
+  - DebianPackage
+  - RedhatPackage
+  - ZipArchive
+  - TarGzArchive
+  - TextFile
+  - TestBundle
```
JunTaoLuo: lzma files too

docs/BillOfMaterials.md
```diff
@@ -0,0 +1,170 @@
...
+Optional attributes:
+    - Others - any other attributes on the element are considered metadata values about the artifact.
+
+#### `<Dependencies>`
+
+Describes artifact dependencies.
+
+#### `<Link>`
+
+Describes a dependency between two artifacts.
```
JunTaoLuo: This file isn't meant to be human readable is it? I'm imagining that the list of dependencies can grow pretty large. Also, is this a full closure of dependencies or is it the top level dependencies?

docs/BuildCaching.md
```diff
@@ -0,0 +1,30 @@
...
+This will cause the build to always rebuild assets.
+
+## When to commit a new bill of material
+
+Each build will produce a new Bill of Materials (bom). Normally, this bom isn't added to source code right away.
+This bom should only be committed to source once the assets it describes are published and available from caching mechanisms.
+
+## Caching mechanisms
+
+These are some common caching mechanisms used to store build artifacts.
```
JunTaoLuo: I think there needs to be a way in the bom to specify where each artifact is cached. Also what if there are multiple possibilities? The installers are generally uploaded to the azure blob and then released on debian and rpm feeds later on. For example, the store archives are also available through the download center.

docs/BillOfMaterials.md
natemcmaster: Super-human readable only. The file is really meant to be consumed by other tools.

Just top-level dependencies.

docs/BillOfMaterials.md
```diff
@@ -0,0 +1,170 @@
...
+
+</Build>
+```
+
+### Example usage: signing
+
+In the example above, some of the `<Artifact>` items were marked `ShouldBeSigned="true"`. Our signing tool could use this as a way to
+determine which files should be passed on to signing.
+
+### Example usage: metadata
```
JunTaoLuo: Do we need mechanisms for appending to a bom or merging multiple boms for ingestion? For example, the runtime store installers run after UniverseCoherence. How do we coalesce the artifacts from those two builds for example?

docs/BillOfMaterials.md
```diff
@@ -0,0 +1,170 @@
...
+            FileHash="ABCDEF0123456789" />
+        <Artifact Id="Microsoft.AspNetCore.2.0.0-rtm-123.nupkg" Type="NuGetPackage"
+            ShouldBeSigned="true"
+            PackageId="Microsoft.AspNetCore"
+            Version="2.0.0-rtm-123" />
+        <Artifact Id="Microsoft.AspNetCore.2.0.0-rtm-123.symbols.nupkg" Type="NuGetSymbolsPackage"
+            ShouldBeSigned="true"
+            PackageId="Microsoft.AspNetCore"
+            Version="2.0.0-rtm-123" />
+        <Artifact Id="aspnetcore-store-2.0.0-rtm-123-win-x64.zip" Type="ZipArchive"
```
JunTaoLuo: Curious what we need to do to generate these entries.

docs/BuildCaching.md
natemcmaster: Yes. This kind of configuration should exist in the repo, but not in the bill of materials. We do something like this today with nuget packages. I could easily imagine expanding this approach to include blob feeds too. What this *might* look like:

```xml
   <PropertyGroup>

    <RestoreSources>$(DotNetRestoreSources)</RestoreSources>
    <RestoreSources Condition="'$(DotNetBuildOffline)' != 'true'  ">
      $(RestoreSources);
      https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json;
      https://api.nuget.org/v3/index.json;
    </RestoreSources>

    <BlobSources>$(DotNetBlobSources)</BlobSources>
    <BlobSources Condition="'$(DotNetBuildOffline)' != 'true'  ">
      $(BlobSources);
      https://aspnetcore.blob.core.windows.net/buildfeed/;
      https://dist.asp.net/buildfeed/
      file:///$(MSBuildThisFileDirectory)/../.deps/;
    </AzureBlobSources>

  </PropertyGroup>
```

docs/BillOfMaterials.md
natemcmaster: I'm working on an update to BuildTools to generate this. We already generate some of this metadata for nupkgs. https://github.com/aspnet/BuildTools/blob/2fdcbd3f8c1461caed668ebeaf42c9545d30fcbb/files/KoreBuild/modules/solutionbuild/Project.Inspection.targets

docs/BillOfMaterials.md
natemcmaster: Yes, appending, merging, etc are all good tools to have. As a starting point, we could throw if two boms being merged have clashing artifact IDs.

natemcmaster:
Thanks all for the feedback. I'm going to start working on an implementation of this plan. I assume more details will come up as I work on it.

PR #690, opened 2017-12-01
Add MonoDevelop.Sdk package to the mirror.

NTaylorMullen:


build/dependencies.props
```diff
@@ -75,6 +75,7 @@
     <MicrosoftWebXdtPackageVersion>1.4.0</MicrosoftWebXdtPackageVersion>
     <MicrosoftWin32RegistryPackageVersion>4.5.0-preview1-25914-04</MicrosoftWin32RegistryPackageVersion>
     <MonoAddinsPackageVersion>1.3.7</MonoAddinsPackageVersion>
+    <MonoDevelopSdkPackageVersion>1.0.0</MonoDevelopSdkPackageVersion>
```
NTaylorMullen: Given that the package name is `MonoDevelop.Sdk` will this break any of our pascal casing code?

natemcmaster: No, it should be fine. You can test this locally by running `build.cmd /t:GeneratePropsFiles`. Look in `obj/` for the file it creates.

natemcmaster:
cc @Eilon. Does MonoDevelop count as a third-party dependency?

NTaylorMullen:
> Sure. Could you link the staged PR that will be using this?

https://github.com/aspnet/Razor/pull/1790

build/dependencies.props
muratg: Just curious. Why do we need this?

NTaylorMullen: > Just curious. Why do we need this?

This is for VS for mac Razor support.

Eilon:
> cc @Eilon. Does MonoDevelop count as a third-party dependency?

It's not 3rd party, but what exactly is the concern?

natemcmaster:
OSS approval.

natemcmaster:
:shipit: I'm assuming we don't need OSS approval since this doesn't count as a 3rd party dependency.

PR #691, opened 2017-12-01
Removing certain, unreferenced, external dependencies.

ahsonkhan:
cc @natemcmaster, @muratg, @davidfowl, @KrzysztofCwalina, @pakrym


davidfowl:
Why aren't we also removing `System.IO.Pipelines.Text.Primitives`

ahsonkhan:
> Why aren't we also removing System.IO.Pipelines.Text.Primitives

We could if it isn't needed. I don't have enough context here to make that call.

Otherwise, is this PR good to merge?

build/external-dependencies.props
```diff
@@ -117,17 +117,14 @@
     <CoreFxLabFeed>https://dotnet.myget.org/f/dotnet-corefxlab/api/v3/index.json</CoreFxLabFeed>
   </PropertyGroup>
   <ItemGroup>
-    <ExternalDependency Include="System.Buffers.Experimental" Version="$(SystemBuffersExperimentalPackageVersion)" Source="$(CoreFxLabFeed)" Private="true" Mirror="true" />
```
natemcmaster: You can also delete these variables from dependencies.props if they are not used anywhere else in this repo.

https://github.com/aspnet/Universe/blob/dd2cc91c10333f08cb42f8bd6466123816f48d22/build/dependencies.props#L121

ahsonkhan:
>  assuming build.cmd /t:CheckGraph still passes

Do you mean ComputeGraph?

natemcmaster:
Yeah, that one.

PR #695, opened 2017-12-05
Clean up Universe

ryanbrandenburg:
We weren't using the same versions of `KoreBuild` and `Internal.AspNetCore.Sdk` (which are built together), which caused some problems for me when building locally.


scripts/UpdateRepos.ps1
```diff
@@ -105,7 +105,7 @@ try {
                 # Push the changes
                 if (-not $NoPush -and ($Force -or ($PSCmdlet.ShouldContinue("Pushing updates to repos.", 'Push the changes to these repos?')))) {
                     try {
-                        Invoke-Block { & git @gitConfigArgs push origin HEAD:$submodule.branch}
+                        Invoke-Block { & git @gitConfigArgs push origin HEAD:$(submodule.branch)}
```
ryanbrandenburg: We're still iterating on this script, which pushes new packages back to the repositories.

natemcmaster: Syntax error. Should be `HEAD:$($submodule.branch)`

build/dependencies.props
```diff
@@ -7,11 +7,11 @@
     <FSharpCorePackageVersion>4.2.1</FSharpCorePackageVersion>
     <FSharpNETSdkPackageVersion>1.0.5</FSharpNETSdkPackageVersion>
     <GoogleProtobufPackageVersion>3.1.0</GoogleProtobufPackageVersion>
-    <InternalAspNetCoreSdkPackageVersion>2.1.0-preview1-15616</InternalAspNetCoreSdkPackageVersion>
+    <InternalAspNetCoreSdkPackageVersion>$(KoreBuildVersion)</InternalAspNetCoreSdkPackageVersion>
```
mkArtakMSFT: KoreBuildVersion ? Why not CoreBuildVersion ? I understand this is a variable name, defined somewhere else, just wondering about why "Kore" ?

natemcmaster: Historical reasons. Long ago, before we branded as ASP.NET Core, our work was kalled Project K. Hence the 'K' instead of 'C'. KoreBuild, Kestrel, and [kvm](https://blogs.msdn.microsoft.com/webdev/2015/01/16/making-it-better-updates-for-asp-net-5-in-visual-studio-2015-ctp-5/). We just never kame up with a better name.

PR #696, opened 2017-12-05
Repoproj 1 - Add task to generate repoproj files based on package producers/consumers

natemcmaster:
Repoproj refactoring 1 . Targeting <kbd>feature/repo-proj</kbd> as this refactoring will come in stages. There will be more changes before this merges to release/2.0.0.

This target detects and generates repoproj files based on the PackageReferences and packages produced by a repo.

Also, for inspection, I added generation of modules/SubmoduleGraph.dgml. This produces a visualization in VS that can be used to inspect repo-to-repo dependencies. This currently looks like this:

image


src/Managed/Scaffolding.repoproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project>
```
ryanbrandenburg: I guess we check these in because they take a while to generate and won't change very often (basically never in a patch branch)?

natemcmaster: Correct. Think of these like our resx and Designer.cs files. It's generated, but checked in.

Issue #686, opened 2017-11-30
Split artifacts.props up into individual repos


natemcmaster:
Instead of one massive list here: https://github.com/aspnet/Universe/blob/dev/build/artifacts.props, each repo should contain a "repo manifest" indicating the shipping status of each artifact.

See also #679

natemcmaster:
I know there was some resistance to the idea because a few people like having a central list for all things, but trying to keep aspnet/Universe's 49 submodules and the build/artifacts.props list in sync has been a cause of more headaches and build breaks than I think it's worth. 

@Eilon @muratg - can we do this for rc1 instead of waiting for MQ? This issue is a continual source of pain.

cc @pakrym @anurse

anurse:
I definitely wasted like... 20mins of my time managing this. If we want a master list, have Universe generate it and spit it out at the end rather than requiring it be consumed.

> can we do this for rc1 instead of waiting for MQ? This issue is a continual source of pain

![https://media3.giphy.com/media/aZ4sQUpybai5y/giphy.gif](https://media3.giphy.com/media/aZ4sQUpybai5y/giphy.gif)

muratg:
@anurse I was thinking the same. We can get the same benefit in other ways.

Eilon:
Sure thing.

natemcmaster:
No plans to change this at the moment. If the mondo-repo plans fall through, we can revisit.

PR #697, opened 2017-12-06
Add Microsoft.AspNetCore.AzureKeyVault.HostingStartup

pakrym:
https://github.com/aspnet/AzureIntegration/pull/126


natemcmaster:
When you merge https://github.com/aspnet/AzureIntegration/pull/126, can you also update modules/AzureIntegration to the new commit that contains this new package?

pakrym:
@natemcmaster I can, but why is it required? Wouldn't universe build do it automatically?

natemcmaster:
Our update scripts will do it automatically, but I still recommend this for any change to git submodules that would break Universe. If you push without updating the submodule, aspnet/Universe is broken and won't build until that updater script succeeds.

natemcmaster:
FYI - for adding/removing packages, this should happen less once we do #686

PR #698, opened 2017-12-06
Cleaning up shared framework generation.

JunTaoLuo:
I changed the tfm to netcoreapp2.1 to avoid all the build warnings and didn't consider the aftermath. Update: decided to keep the tfm at 2.1.

### DIFF
--- a/build/tools/templates/SharedFx/SharedFx.csproj
+++ b/build/tools/templates/SharedFx/SharedFx.csproj
@@ -20,7 +20,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-   <PackageReference Include="Microsoft.AspNetCore.All" Version="$(PackageVersion)" />
+   <PackageReference Include="Microsoft.AspNetCore.All" Version="$(PackageVersion)" ExcludeAssets="Build" />
    <PackageReference Include="Microsoft.NETCore.App" Version="$(RuntimeFrameworkVersion)"/>
   </ItemGroup>
 </Project>
--- a/src/Microsoft.AspNetCore.All/Microsoft.AspNetCore.All.csproj
+++ b/src/Microsoft.AspNetCore.All/Microsoft.AspNetCore.All.csproj
@@ -23,7 +23,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Content Include="build\**\*.props" PackagePath="%(Identity)" />
+    <Content Include="build\$(TargetFramework)\*.props" PackagePath="%(Identity)" />
     <Content Include="lib\$(TargetFramework)\_._" PackagePath="%(Identity)" />
   </ItemGroup>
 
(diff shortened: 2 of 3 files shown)


natemcmaster:
Is this temporary?

natemcmaster:
Discussed in person. The association of target framework with package version is a little problematic. In the SDK today, is is the package version of $(MicrosoftNETPlatformLibrary) that determines what is generated in `runtimeconfig.json`. Here are some scenarios I think we need to consider carefully first.

### Target framework downgrade

Should this work?
```xml
<Project Sdk="Microsoft.NET.Sdk.Web">
  <PropertyGroup>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.1.0" />
     <!-- SDK adds implicit package reference to Microsoft.NETCore.App 2.0.0 -->
  </ItemGroup>
</Project>
```

The app expects to run on `shared/Microsoft.AspNetCore.All/2.1.0/`

### Target framework upgrade

Should this work?
```xml
<Project Sdk="Microsoft.NET.Sdk.Web">
  <PropertyGroup>
    <TargetFramework>netcoreapp2.1</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.0" />
     <!-- SDK adds implicit package reference to Microsoft.NETCore.App 2.1.0 -->
  </ItemGroup>
</Project>
```

The app expects to run on `shared/Microsoft.AspNetCore.All/2.0.0/`


cc @DamianEdwards

JunTaoLuo:
Note that $(MicrosoftNETPlatformLibrary) is only set by the Microsoft.AspNetCore.All 2.1.0 metapackage. The 2.0.0 version would not set this parameter.

JunTaoLuo:
Seems like there are a few combinations we need to consider here:

| Microsoft.AspNetCore.All TFM | Microsoft.AspNetCore.All package version | Microsoft.NETCore.App shared framework dependency in runtimeconfig.json | Options |
| --- | --- | --- | --- |
| netcoreapp2.0 | 2.0.0 | 2.0.0 | Future 2.x patch |
| netcoreapp2.0 | 2.1.0 | 2.1.0 | Option1 |
| netcoreapp2.1 | 2.1.0 | 2.1.0 | Option2 |

According to https://github.com/dotnet/corefx/issues/25766 it seems like the tfm and the package version of Microsoft.NETCore.App must match so if we follow that pattern, we should choose Option2. However that would mean that we violate our rule of not updating a tfm of a package during a minor release.

JunTaoLuo:
The decision here is to keep the metapackage's TFM as netcoreapp2.1. The metapackage as a special package that won't adhere to the rule that TFM do not change in minor releases.

build/tools/templates/SharedFx/SharedFx.csproj
```diff
@@ -20,7 +20,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-   <PackageReference Include="Microsoft.AspNetCore.All" Version="$(PackageVersion)" />
+   <PackageReference Include="Microsoft.AspNetCore.All" Version="$(PackageVersion)" ExcludeAssets="Build" />
```
JunTaoLuo: We need this to prevent trimming when building the shared framework.

PR #703, opened 2017-12-11
Ensure template tests run on CI

natemcmaster:
Templates tests aren't actually running on CI. 

⌚️ on @JunTaoLuo to fixup some stuff in the shared fx. Templates tests do not pass with latest builds of aspnetcore.


JunTaoLuo:
SharedFX will be fixed once we ingest latest 2.2 sdk

JunTaoLuo:
So it looks like we have ingested the new SDK, we should give the template tests another run.

natemcmaster:
This requires some new thinking about how we test with the shared framework. Since the template tests will assume the shared fx is installed, we need to figure that bit out first.

natemcmaster:
Opened https://github.com/aspnet/Universe/issues/713 to follow up on this

PR #704, opened 2017-12-12
Update BenchmarkDotNet

pakrym:
To get `netcoreapp21` support


smitpatel:
This version upgrade involved breaking changes. It broke all the benchmarks of EF Core.
cc: @ajcvickers

natemcmaster:
Does EF use the "fast validate" tool @pakrym wrote for benchmarks? KoreBuild can be configured to build and run a fast run of benchmarks to ensure they keep working.

https://github.com/aspnet/Common/blob/ed962e9123b30dc41c940bce703a882522e5676f/shared/Microsoft.AspNetCore.BenchmarkRunner.Sources/

PR #707, opened 2017-12-12
Add Microsoft.AspNetCore.Razor.Design

rynowak:
Adding a new package for Razor MSBuild assets

/cc @natemcmaster @pranavkm


build/artifacts.props
```diff
@@ -112,6 +112,7 @@
     <PackageArtifact Include="Microsoft.AspNetCore.Proxy" Category="noship" />
     <PackageArtifact Include="Microsoft.AspNetCore.RangeHelper.Sources" Category="noship" />
     <PackageArtifact Include="Microsoft.AspNetCore.Razor" Category="ship" Metapackage="true" RuntimeStore="true"/>
+    <PackageArtifact Include="Microsoft.AspNetCore.Razor.Design" Category="ship" Metapackage="true" RuntimeStore="true"/>
```
muratg: @JunTaoLuo @natemcmaster FYI for potential ASP.NET platform and/or ProdCon impact

natemcmaster: Shouldn't affect prodcon unless this adds some kind of new build dependency (only builds in VS, or something)

pranavkm:
FYI - we used to have a package with this name - https://www.nuget.org/packages/Microsoft.AspNetCore.Razor.Design/1.1.0-preview4-final. Need to make sure we version this package to avoid conflicts.

natemcmaster:
Where was the change to add this package? CI is broken because the current submodule Universe is pulling from does not produce this package.

rynowak:
Did this change??? Do I need to create the package before adding it here? 

I'm used to the requirement to add a package here before checking in any new code.

natemcmaster:
It changed a few weeks ago. Universe now asserts that all packages in `artifacts.props` and all packages in `artifacts/build/` are an exact match. 

FYI - this might change soon if we can get #686 done before you finish building this new package

PR #708, opened 2017-12-12
Set DotNetAdditionalRestoreSources on all ExternalDependency items

natemcmaster:
cc @dagood 

Other assets, not just those from dotnet-core/roslyn, may come from DotNetAdditionalRestoreSources  in an orchestrated build

### DIFF
--- a/build/external-dependencies.props
+++ b/build/external-dependencies.props
@@ -384,7 +384,7 @@
 
   <ItemGroup Condition="'$(DotNetAdditionalRestoreSources)'!=''">
     <ExternalDependency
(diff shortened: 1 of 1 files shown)


build/external-dependencies.props
```diff
@@ -384,7 +384,7 @@
 
   <ItemGroup Condition="'$(DotNetAdditionalRestoreSources)'!=''">
     <ExternalDependency
-      Update="@(ExternalDependency->WithMetadataValue('Source', '$(DotNetCoreFeed)'));@(ExternalDependency->WithMetadataValue('Source', '$(RoslynFeed)'))"
+      Update="@(ExternalDependency)"
```
pranavkm: Out of curiosity, what packages did this miss?

natemcmaster: Packages that we are currently pulling from nuget.org, like Microsoft.Build.Framework. This one is now building in PCP

PR #711, opened 2017-12-13
Don't import the external package props file into AdditionalImports

natemcmaster:
This shields us against the external PackagePropsPath from adding package versions that we don't want overriden. For example, in the last run of the orchestrated build, the external props file included versions such as `MicrosoftAspNetCoreMvcRazorPackageVersion`, which broke the version consistency of inter-repo dependencies.

This additional import isn't really needed. We map the versions from the external file to the our own generated versions props file.

cc @dagood


dagood:
Is `MicrosoftAspNetCoreMvcRazorPackageVersion` built during an ASP.NET build? This may have been my fault because of where the feed I'm using came from: it did have an ASP.NET build run pointing to it, and the build just didn't complete successfully due to the multiple build chain issue. (I think there was also an error in one of them.) This will get better when we get proper retry support in orchestration.

That all said, shielding against this is fine, I'm just pointing out this shouldn't happen normally in case there's any risk in making this change.

pranavkm:
This change looks fine. I changed how this works but forgot to backout this change. :shipit:

natemcmaster:
> Is MicrosoftAspNetCoreMvcRazorPackageVersion built during an ASP.NET build

Yes, this comes from aspnet. I figured it was caused by the way we started the orchestrated build run, but it exposed a problem we should fix anyways. So, glad we hit it now

PR #717, opened 2017-12-16
Update package dependencies

ahsonkhan:
Required to get new MemoryMarshal APIs

See: https://github.com/aspnet/KestrelHttpServer/pull/2222

cc @pakrym, @davidfowl, @BrennanConroy, @KrzysztofCwalina, @natemcmaster, @jkotas, @joshfree


BrennanConroy:
Is a new runtime also required?

ahsonkhan:
> Is a new runtime also required?

I don't think so. As far as I can tell, Kestrel relies on slow span and OOB System.Memory, which contains the new APIs. If it relies on fast span and the inbox version of System.Memory, then we would need to update the runtime.

Why do you ask?

pakrym:
Please update `MicrosoftNETCorePlatformsPackageVersion`. Also we don't build universe in PR so you have to verify updates locally.

pakrym:
Closing in favor of https://github.com/aspnet/Universe/pull/718

PR #718, opened 2017-12-18
Update pacakge dependencies 

pakrym:
Modified version of https://github.com/aspnet/Universe/pull/717


pakrym:
It passed up until EF, which failed with the database already exists errors which always happens on my machine.

ryanbrandenburg:
I imagine it will be fine then, EF is ~halfway down the chain so it probably would have failed by then if it was going to.

Separately, file an issue about the database exists error. Tests should be resilient to this kind of check-able and resolvable machine state.

PR #723, opened 2017-12-21
Crossgen shared framework and create corresponding symbol files

JunTaoLuo:
Assumptions to verify with coreclr once they are back from vacation:
1. We do not need to crossgen and create symbols for resources dlls. This makes sense since these assemblies not contain any code and only contains strings.
2. The difference between -app_paths and -platform_assemblies_paths. I don't know what the difference is but I'm passing the dotnet runtime binaries via -platform_assemblies_paths and aspnetcore binaries via -app_paths. I have verified that I can run an mvc app with the crossgen'ed assemblies but we should understand what distinguishes these two command line switches


JunTaoLuo:
Verified that the shared framework builds and crossgens on win, osx and linux. Also verified that published apps using the aspnetcore shared framework runs on all three platforms.

JunTaoLuo:
So far the only person who seems knowledgeable about crossgen seems to be @javiercn. 

As for testing, I installed the crossgen'ed shared framework and published an app that depends on the corresponding .All metapackage. I verified that the published output is trimmed against both shared frameworks and running the published output loads the assemblies from /shared/Microsoft.NETCore.App and /shared/Microsoft.AspNetCore.All.

I agree we should add tests for these scenarios. I'd like to discuss how we should do this infrastructure-wise when I'm back. I don't think that should block this PR though.

build/SharedFx.targets
```diff
@@ -121,6 +127,137 @@
...
+      <LibExtension Condition="$([MSBuild]::IsOSPlatform('OSX'))">.dylib</LibExtension>
+    </PropertyGroup>
+
+    <!-- Determine runtime location -->
+    <MSBuild Projects="$(SharedFxWorkDirectory)SharedFx.csproj" Targets="GetPackageDefinitions" >
+      <Output TaskParameter="TargetOutputs" ItemName="PackageDefinitions" />
+    </MSBuild>
+
+    <ItemGroup>
+      <RuntimePackage Include="@(PackageDefinitions)" Condition="$([System.String]::new('%(PackageDefinitions.Name)').Contains('runtime')) AND $([System.String]::new('%(PackageDefinitions.Name)').Contains('$(RuntimePackageName)'))" />
```
natemcmaster: Instead of batching, would this work? 
```xml
<RuntimePackage Include="@(PackageDefinitions->WithMetadataValue('Name', 'runtime.$(RuntimePackageName)'))" />
```

build/SharedFx.targets
```diff
@@ -121,6 +127,137 @@
...
+    </ItemGroup>
+
+    <!-- Compute the intersection of crossgen candidates and native/resources assemblies as the set of assemblies to skip crossgen -->
+    <CreateItem Include="@(_AssembliesToCrossgen)" Condition="'%(Filename)' != ''and '@(IgnoredAssemblies)' != ''">
+      <Output TaskParameter="Include" ItemName="AssembliesToRemove"/>
+    </CreateItem>
+
+    <!-- Resolve list of assemblies to crossgen -->
+    <ItemGroup>
+      <AssembliesToCrossgen Include="@(_AssembliesToCrossgen)">
```
natemcmaster: Could be condensed to 
```xml
<AssembliesToCrossgen Include="@(_AssembliesToCrossgen)" Exclude="@(AssembliesToRemove)">
```

build/SharedFx.targets
```diff
@@ -121,6 +127,137 @@
...
+
+  <Target Name="CrossGenAssemblies"
+    DependsOnTargets="PrepareForCrossGen"
+    Inputs="@(AssembliesToCrossgen)"
+    Outputs="%(AssembliesToCrossgen.Native)">
+    <ItemGroup>
+      <CrossGenArgs Include="-nologo" />
+      <CrossGenArgs Include="-readytorun" />
+      <CrossGenArgs Include="-in %(AssembliesToCrossgen.Source)" />
+      <CrossGenArgs Include="-out %(AssembliesToCrossgen.Native)" />
```
natemcmaster: Nit naming: Source/Output or Source/Destination

build/SharedFx.targets
```diff
@@ -133,4 +270,6 @@
       WorkingDirectory="$(SharedFxOutputDirectory)"
       Overwrite="true"/>
   </Target>
+
+  <Target Name="GenerateSharedFx" DependsOnTargets="BuildSharedFx;CrossGenAssemblies;CrossGenSymbols;PackSharedFx"/>
```
natemcmaster: Requires a CI config change, right? Could you keep BuildSharedFx as the top-level target and rename the existing BuildSharedFx to something like "CollectSharedFxFiles"

javiercn:
I dug out a rsp file that I used to crossgen the runtime cache for our 1.0.3
/platform_assemblies_paths needs to contain mscorlib and all the stuff that "makes up the platform". I believe this is the stuff that is part of the shared framework. In our case I would imagine everything goes here?
/app_paths needs to contain all the assemblies that are not part of the platform like Microsoft.Extensions.Primitives, etc. in the old cache.

I do believe though that there is not a meaningful difference between the two arguments so long as /platform_assemblies_paths contains a reference to System.Runtime.

See my rsp file https://gist.github.com/javiercn/8c77af29b5f9eb99e1558ffbaaf528ec for details

build/tools/templates/SharedFx/SharedFx.csproj
```diff
@@ -23,4 +23,33 @@
    <PackageReference Include="Microsoft.AspNetCore.All" Version="$(PackageVersion)" ExcludeAssets="Build" />
    <PackageReference Include="Microsoft.NETCore.App" Version="$(RuntimeFrameworkVersion)"/>
   </ItemGroup>
+
+  <Target Name="GetPackageDefinitions" Returns="@(_PackageDefinitions)">
+    <ItemGroup>
+      <ProjectAssetsJson Include="$(MSBuildThisFileDirectory)**\project.assets.json" />
```
natemcmaster: Why search? The property `ProjectAssetsFile` should already be set to the right file path.

https://github.com/dotnet/sdk/blob/b5be57155df09c2f76afb6050397d1a85a745f45/src/Tasks/Microsoft.NET.Build.Tasks/build/Microsoft.PackageDependencyResolution.targets#L28

build/SharedFx.targets
JunTaoLuo: I tried that, for some reason this did not work and I had to use Remove. I'll test it out a little more to see why not.

build/SharedFx.targets
JunTaoLuo: it's a little harder to get this to work since the resolved package actually looks like runtime.win-x64.Microsoft.NETCore.App. It's also a little tricky to determine the actual runtime since we use the RID win7-x64 which resolves to win-x64 of the runtime package. IIRC we needed to use win7-x64 instead of win-x64 because some of our dependencies had specific native assets that needed to be picked.

build/SharedFx.targets
JunTaoLuo: Sounds good

build/SharedFx.targets
JunTaoLuo: Sure

build/tools/templates/SharedFx/SharedFx.csproj
JunTaoLuo: Didn't know this existed. I'll update https://github.com/aspnet/Universe/blob/e860022e1a7b1ea825c1d185d1084feec24faf47/build/tools/templates/RS.Manifest/RS.Manifest.csproj#L25 as well.

build/SharedFx.targets
natemcmaster: MSBuild globbing matches can be finicky. For example if you include "C:\myfile.txt" and exclude "C:\dir\..\myfile.txt" it doesn't work, even though the normalized file path is the same

build/SharedFx.targets
natemcmaster: i see. Seems like it should be hard to establish this mapping using conditional.

My concern is that if the CoreCLR team introduces a new RID, we could end up pickup something we may not have meant to use.

JunTaoLuo: I think in this case, it should be okay. Since as far as I can tell, the package definitions I'm searching through only contain packages that are restored. Restore takes care of the RID resolution.

JunTaoLuo:
This is probably going to conflict with the .App/.All split that I'm currently working on but I think this should go in first. Let's get this reviewed and merged soon @natemcmaster.

build/SharedFx.targets
```diff
@@ -121,6 +127,137 @@
...
+      <CrossGenArgs Include="-app_paths $(SharedFxPublishDirectory)" />
+      <CrossGenArgs Include="-platform_assemblies_paths $(SharedFxCrossGenToolDirectory)" />
+      <CrossGenArgs Include="-JITPath %(ClrJitAssembly.FullPath)" />
+    </ItemGroup>
+
+    <MakeDir Directories="$([System.IO.Path]::GetDirectoryName('%(AssembliesToCrossgen.Rsp)'))" />
+    <MakeDir Directories="$([System.IO.Path]::GetDirectoryName('%(AssembliesToCrossgen.Destination)'))" />
+    <WriteLinesToFile File="%(AssembliesToCrossgen.Rsp)" Lines="@(CrossGenArgs)" Overwrite="true" />
+
+    <Exec Command="$(SharedFxCrossGenToolDirectory)$(CrossGenTool) @%(AssembliesToCrossgen.Rsp)" EnvironmentVariables="COMPlus_PartialNGen=0" />
```
natemcmaster: Consider also adding `StandardOutputImportance="Normal"`. Makes `-verbose:minimal` actually minimal. You can still see all the crossgen output for `-verbose:normal`

natemcmaster:
DTN and :shipit:

JunTaoLuo:
Rebased and fixed conflicts. I will look at this tomorrow with a fresh pair of eyes before merging.

PR #729, opened 2017-12-27
Change output of sharedfx build to match feed layout

natemcmaster:
Output of sharedfx build now matches expected blob store layout

```
<root>/Runtime/$(PackageVersion)/aspnetcore-runtime-$(PackageVersion)-$(Rid).zip
<root>/Runtime/$(PackageVersion)/aspnetcore-runtime-$(Rid)-version-badge.svg
<root>/Runtime/$(PackageVersion)/latest.aspnetcore.version
```

Other changes
- Upgrade build tools
- Produce an SVG badge for the sharedfx
- Produce the 'latest.aspnetcore.version' file for the dotnet-install.ps1/sh downloader script
- Change layout of output for sharefx build
- Use KoreBuild's bundled package folder for internal build tools packages

cc @javiercn 

cc @jkotalik - includes https://github.com/aspnet/BuildTools/commit/c390edb6f73e3ab992402835c004414c1fadfd4a

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+    "files.trimTrailingWhitespace": true,
+    "files.associations": {
+        "*.*proj": "xml",
+        "*.props": "xml",
+        "*.targets": "xml"
+    }
+}
--- a/build/SharedFx.targets
+++ b/build/SharedFx.targets
@@ -48,8 +48,8 @@
 
     <PropertyGroup>
       <SharedFxWorkDirectory>$(_WorkRoot)SharedFx\</SharedFxWorkDirectory>
-      <SharedFxOutputDirectory>$(_WorkRoot)Publish\</SharedFxOutputDirectory>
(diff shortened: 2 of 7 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
```
javiercn: nit: Remove.

We should add this to all our .gitignore files across repos

natemcmaster: Actually, this was very intentional.  We already have some VS code settings files committed to source, both in Universe https://github.com/aspnet/Universe/tree/dev/.vscode and in some runtime repos too https://github.com/aspnet/KestrelHttpServer/tree/dev/.vscode

javiercn: I think those got there as an accident. See https://github.com/aspnet/Identity/issues/1428
That said, I don't really care about it. The only problem I see is that those are User specific environment files, so they will change if a user changes something

PR #730, opened 2017-12-28
Update Razor.Design to noship

rynowak:
I'm having lotsa problems getting this package to flow through all of our packaging checks. I don't think we're going to ship this in it's current form, so I'm changing to noship as an unblocker


rynowak:
FYI only - merging this to unblock coherence signed

Issue #721, opened 2017-12-19
Produce Microsoft.AspNetCore.App package and shared framework


JunTaoLuo:
This package will include first party packages and a few sanctioned third party packages. The .All package will take a dependency on the .App package and exclude its build assets. The .All shared framework will be dependent on the .App shared framework. Transitive references to third party references in both packages will be flattened.

Packages in the .All that will be removed from .App:
```
SQLitePCLRaw.*

    <PackageReference Include="Microsoft.Data.Sqlite" version="2.0.0" />
    <PackageReference Include="Microsoft.Data.Sqlite.Core" version="2.0.0" />
    <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" version="2.0.1" />
    <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite.Core" version="2.0.1" />

StackExchange.Redis

    <PackageReference Include="Microsoft.Extensions.Caching.Redis" version="2.0.0" />

WindowsAzure.Storage

    <PackageReference Include="Microsoft.AspNetCore.DataProtection.AzureStorage" version="2.0.1" />

Microsoft.Azure.KeyVault

    <PackageReference Include="Microsoft.Extensions.Configuration.AzureKeyVault" version="2.0.0" />
 ```

Third party packages remaining in .App:
```
Libuv
Remotion.Linq
Newtonsoft.Json
Microsoft.VisualStudio.Web.BrowserLink
System.Interactive.Async
```

cc @anurse @DamianEdwards

**Comment by natemcmaster** [comment]
FYI - BrowserLink isn't third-party. We build it. https://github.com/aspnet/browserlink

**Comment by JunTaoLuo** [comment]
Fair, though I believe there's some discussions on whether this package should remain in .App in the future, which is why I called it out explicitly here.

**Comment by Eilon** [comment]
The list above is missing Ix-Async. @ajcvickers can you follow up with @DamianEdwards to confirm that this stays in the list? And if not, what is the plan?

**Comment by DamianEdwards** [comment]
It should be there, along with EF Core in general.

**Comment by JunTaoLuo** [comment]
Updated. I think I missed it because the package name is System.Interactive.Async so I missed it in my previous list.

**Comment by JunTaoLuo** [comment]
Seems like I was generating the list of packages based on 2.0 packages for the original list. In 2.1 we also need to exclude these packages:
```
Microsoft.Azure.KeyVault
[...]

PR #731, opened 2017-12-28
Produce the Windows sharedfx as win-x64/x64 for consistency with Microsoft.NETCore.App and dotnet-sdk

natemcmaster:
All download scripts and infrastructure is currently written to look for files named `aspnetcore-runtime-{version}-win-{arch}.zip`, but we are currently producing files as `win7-{arch}`.

IIUC the only technical reason for this is that SQLitePCLRaw uses win7-x64 in its native nupkg. This binary works just fine on all .NET Core 2.0 win-x64/x86 compatible platforms (i.e.
 we don't support Vista and Windows ...https://github.com/dotnet/core/blob/master/release-notes/2.0/2.0-supported-os.md), so there shouldn't be any reason to keep the `win7` identifier in the final product name.

Plus, we aren't currently planning to put SQLitePCLRaw in Microsoft.AspNetCore.App. cref #721 

cc @javiercn


build/SharedFx.targets
```diff
@@ -38,7 +38,7 @@
 
   <Target Name="BuildSharedFx" DependsOnTargets="ResolveCommitHash">
     <ItemGroup>
-      <_AcceptableSharedFxRIDs Include="win7-x64;win7-x86;osx-x64;linux-x64"/>
+      <_AcceptableSharedFxRIDs Include="win-x64;win-x86;osx-x64;linux-x64"/>
```
JunTaoLuo: This is going to require a change in the build configuration on the CI.

natemcmaster:
@JunTaoLuo is there any other reason we need to keep win7? I looked over the project.assets.json file, and I didn't see any other win7-specific assets.

build/SharedFx.targets
natemcmaster: That's fine. I can update it.

natemcmaster:
> So this change does not affect any functionality and is mostly just a rename of the zips?

Correct.

JunTaoLuo:
I kept the RID as win7-* since I wanted to use the same set of restored artifacts as the previous runtime store. I was planning to check which dependency forced us to use win7-* instead of win-* when I got back and verify apps can still run. If it's a rename, I think it's fine, but if we plan to change the restore behaviour I'd like to do some more verification.

natemcmaster:
Ok. I'll make sure to follow thru with all downstream dependencies in our build chain.

natemcmaster:
>  I was planning to check which dependency forced us to use win7-* instead of win-* when I got back and verify apps can still run.

From what I could see, the only difference was e_sqlite3.dll. When restoring SharedFx for win-x64, not win7-x64, the shared framework only has 226 files, not 227. The missing file was e_sqlite3.dll

PR #746, opened 2018-01-03
Update build tools

ryanbrandenburg:
This should let us re-enable the UpdateBuildTools step (or at least find out more about its failure).


scripts/UpdateBuildTools.ps1
```diff
@@ -48,49 +48,57 @@ function Get-KoreBuildVersion {
...
     }
+    finally {
+        Pop-Location
+    }
 }
-finally {
-    Pop-Location
+
+$submodules = Get-Submodules $RepoRoot
+foreach ($submodule in $submodules) {
```
natemcmaster: How do we expect this change to be persisted? Doesn't this already happen as a part of UpdateRepos?

ryanbrandenburg: So it is!

PR #751, opened 2018-01-05
Add a script for more resilient aspnetci-bot pushes

natemcmaster:
Should reduce the number of times Update Universe fails on Ci due to race conditions on manual changes to universe.


natemcmaster:
Talked in person @pakrym. I think we are okay with instability in the auto-update submodules step in favor of more stable builds on commits made to universe.

PR #754, opened 2018-01-05
Produce .App metapackage and shared framework

JunTaoLuo:
Addresses https://github.com/aspnet/Universe/issues/721.

Potentially significant refactoring needed since it conflicts heavily with https://github.com/aspnet/Universe/issues/721.

### DIFF
--- a/build/SharedFx.targets
+++ b/build/SharedFx.targets
@@ -6,85 +6,139 @@
     <SharedFxOutputPath>$([MSBuild]::NormalizeDirectory($(ArtifactsDir)))assets\Runtime\$(PackageVersion)\</SharedFxOutputPath>
   </PropertyGroup>
 
-  <Target Name="BuildMetapackage" DependsOnTargets="ResolveRepoInfo">
-    <!-- Clear working directory -->
-    <RemoveDir Directories="$(_WorkRoot)" />
-
+  <Target Name="_BuildMetapackage" DependsOnTargets="ResolveRepoInfo">
     <PropertyGroup>
-      <MetapackageWorkDirectory>$(_WorkRoot)Microsoft.AspNetCore.All\</MetapackageWorkDirectory>
-      <_SharedFxProps>Configuration=$(Configuration);DotNetRestoreSourcePropsPath=$(GeneratedRestoreSourcesPropsPath);AspNetUniverseBuildOffline=true</_SharedFxProps>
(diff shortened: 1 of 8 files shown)


build/artifacts.props
```diff
@@ -2,7 +2,8 @@
 
   <ItemDefinitionGroup>
     <PackageArtifact>
-      <Metapackage>false</Metapackage>
+      <AppMetapackage>false</AppMetapackage>
```
natemcmaster: Since these cannot both be true, can you add a check to ensure this? Similar to this check:
https://github.com/aspnet/Universe/blob/cedbbd83943770eaefc79e67666cb7365952b0f3/build/repo.targets#L181-L184

Or, you could make this an enum instead of two bools `Metapackage="all"`, `Metapackage="app"`

build/SharedFx.targets
```diff
@@ -253,7 +317,7 @@
     <ItemGroup>
       <CrossGenSymbolsArgs Include="-nologo" />
       <CrossGenSymbolsArgs Include="-readytorun" />
-      <CrossGenSymbolsArgs Include="-app_paths $(SharedFxPublishDirectory)" />
+      <CrossGenSymbolsArgs Include="-app_paths $(AppSharedFxPublishDirectory)%3B$(AllSharedFxPublishDirectory)" />
```
natemcmaster: Needs to vary based on platform. For linux, `:`. For windows, `;`.

```xml
    <!-- 3B = semicolon in ASCII -->
    <PathSeparator Condition="'$(PathSeparator)' == ''">:</PathSeparator>
    <PathSeparator Condition="$(RuntimeIdentifier.StartsWith('win'))">%3B</PathSeparator>
```

build/SharedFx.targets
```diff
@@ -228,7 +292,7 @@
       <CrossGenArgs Include="-readytorun" />
       <CrossGenArgs Include="-in %(AssembliesToCrossgen.Source)" />
       <CrossGenArgs Include="-out %(AssembliesToCrossgen.Destination)" />
-      <CrossGenArgs Include="-app_paths $(SharedFxPublishDirectory)" />
+      <CrossGenArgs Include="-app_paths $(AppSharedFxPublishDirectory)%3B$(AllSharedFxPublishDirectory)" />
```
natemcmaster: same. needs to be platform-specific.

build/artifacts.props
JunTaoLuo: Checks added

PR #756, opened 2018-01-06
Use latest `Microsoft.AspNet.WebApi.Client` package

dougbu:
- aspnet/Mvc#5822
- see also PR aspnet/Mvc#7200


dougbu:
aspnet/Mvc#7200 is approved but on hold for this review. I'll check the two in together.

dougbu:
cedbbd8394

Issue #747, opened 2018-01-04
Produce distro agnostic deb installers


JunTaoLuo:
We need to adopt core-setup's builds for creating distro-agnostic deb installer. See https://github.com/dotnet/core-setup/pull/3487.

PR #759, opened 2018-01-07
Distro agnostic debian installers

JunTaoLuo:
Addresses #747 

Tested and verified installers on debian and ubuntu 14.04.


natemcmaster:
What is in the final output? Do we just have one .deb file now? Or do we still have several?

build/SharedFxInstaller.targets
```diff
@@ -222,26 +222,6 @@
   </Target>
 
   <Target Name="GenerateDebs" DependsOnTargets="_EnsureInstallerPrerequisites">
-    <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="_GenerateDebOnPlatform" Properties="Version=$(Version);Image=debian.8" />
     <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="_GenerateDebOnPlatform" Properties="Version=$(Version);Image=ubuntu.14.04" />
```
natemcmaster: Curious. Why was ubuntu.14.04 kept but other variants were removed?

JunTaoLuo:
Right we now only produce one .deb file and it will be used for every distro.

muratg:
@JunTaoLuo For debian based distros only though, right? (i.e. distros with apt or apt-get).

build/SharedFxInstaller.targets
JunTaoLuo: Mainly to keep in line with what I saw in the core-setup PR https://github.com/dotnet/core-setup/pull/3487, specifically: https://github.com/rakeshsinghranchi/core-setup/blob/2bf97ce26e7c1883707a36a8106aff96cbc07cf1/src/pkg/packaging/deb/package.targets#L49-L52. We now only need to build the .deb installers once and it doesn't matter which platform we build it on. I'm checking with the core-setup team offline right now to see why they chose ubuntu 14.04 but I suspect they just picked it arbitrarily.

JunTaoLuo:
@muratg Yes we will now only have one .deb file instead of one for each of debian.8, debian.9, ubuntu.14.04, ubuntu.16.04 and ubuntu.17.04.

JunTaoLuo:
Note that .rpm installers are intended to remain as is.

build/SharedFxInstaller.targets
JunTaoLuo: Yup checked with @rakeshsinghranchi that ubuntu 14.04 was picked arbitrarily.

natemcmaster: Does this have any effect on the deb produced?

rakeshsinghranchi: @wli3 and @leecow

JunTaoLuo: From my testing on debian.8 ad ubuntu.14.04 there didn't seem to be any effect on the produced debs.

leecow: I don't know enough about the build system to answer. @rakeshsinghranchi will need to speak to how or if the targets impact downstream deb creation (dependency selection, etc).

rakeshsinghranchi: >Does this have any effect on the deb produced?
As long as produced deb ( aspnet package ) doesn't exhibit direct system package dependency, it shouldn't matter. All the distro specific system package is taken care by runtime package.

Issue #757, opened 2018-01-06
Need to build cumulative sharefx archive for all platforms


natemcmaster:
We have cumulative .tar.gz for linux, but not for macos or windows.

By cumulative, I mean it includes Microsoft.NETCore.App, AspNetCore.App, and AspNetCore.All shared frameworks.

cc @JunTaoLuo

Issue #760, opened 2018-01-08
Cumulative shared framework archives need to include 2.0.x runtime stores


JunTaoLuo:
Until the "trampoline" from runtime store to shared framework is implemented, we need to include the 2.0 runtime stores in the 2.1 shared frameworks. This is accomplished via a package dependency in the .deb and .rpm installers. The cumulative archives need to be updated to include these files.

cc @natemcmaster

PR #761, opened 2018-01-08
Generate cumulative archives for all platforms

JunTaoLuo:
Addresses: https://github.com/aspnet/Universe/issues/757

### DIFF
--- a/build/SharedFxInstaller.targets
+++ b/build/SharedFxInstaller.targets
@@ -1,5 +1,5 @@
 <Project>
-  <Target Name="BuildInstallers" DependsOnTargets="GenerateTargzs;GenerateRpms;GenerateDebs;GenerateRelabeledInstallers" />
+  <Target Name="BuildInstallers" DependsOnTargets="GenerateCumulativeArchives;GenerateRpms;GenerateDebs;GenerateRelabeledInstallers" />
 
   <Target Name="_EnsureInstallerPrerequisites">
     <MakeDir Directories="$(_InstallersOutputDir)" />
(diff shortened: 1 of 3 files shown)


build/SharedFxInstaller.targets
```diff
@@ -9,31 +9,76 @@
...
+      SourceFiles="@(SharedFxArchiveFiles)"
+      WorkingDirectory="$(_WorkRoot)"
+      Overwrite="true"
+      Condition="'$(WindowsArchive)' == 'true'"/>
+  </Target>
 
-    <!-- Create Aspnet Runtime tar.gz -->
-    <Exec Command="tar -czf $(_InstallersOutputDir)$(SharedFxArchiveName) -C $(_WorkRoot) ."/>
+  <Target Name="GenerateCumulativeArchives" DependsOnTargets="_EnsureInstallerPrerequisites;_DownloadInstallers">
+    <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="_GenerateCumulativeArchive" Properties="SharedFxPlatform=%(NonWindowsSharedFxRIDs.Identity)" />
```
JunTaoLuo: Currently these steps only work for osx and linux archives. Unzipping dotnet runtime archives creates files with backslashes in their paths and the msbuild ziparchive task fails. I did come up with an ugly workaround, see https://github.com/aspnet/Universe/commit/62e28c6c0087d1a34003732213542df4708ebce2, but I'm sure we have a better mechanism for handling this no? Another alternative would be to create a separate build configuration that runs on windows to generate these archives.

JunTaoLuo:
Updated to address #760

build/SharedFxInstaller.targets
natemcmaster: >  Unzipping dotnet runtime archives creates files with backslashes in their paths

Which file paths specifically? We should notify the core-setup team that they have malformed zip files.

Also, instead of https://github.com/aspnet/Universe/commit/62e28c6c0087d1a34003732213542df4708ebce2, we could add update `UnzipArchive` to normalize file slashes

JunTaoLuo: Looks like all the files contained in dotnet-runtime have this issue.
image

Fortunately our zips don't seem to have this problem. Who do we contact on the core-setup team for this? In the meantime should we update UnzipArchive?

natemcmaster: @weshaggard would know how to fix or who to ask to fix this. These zips are produced from https://github.com/dotnet/core-setup. 

In the meantime, I think we should change the `UnzipArchive` task to normalize slashes. I've seen zips with this issue before, but never has the backslash in the zip entry been intentional...it's always been a mistake.

weshaggard: https://github.com/dotnet/core-setup/blob/41949b226c09874a6a15baecda6bea3184dc00f9/src/pkg/packaging/dir.proj#L127 We are just calling tar to create these, so I would expect them to be valid. Is this a recent change? Did it used to work for you guys?

natemcmaster: @weshaggard  the problem is with the windows .zip files. The tarballs appear ok.

image

weshaggard: I see what are guys using to decompress? Most windows utilities including .NET APIs will accept either slash in the archive.

weshaggard: Also it isn't clear to me what problem these backslashes are causing. 

As for the issue in our zip generation I suspect we are hitting an bug in ZipArchive class which was fixed in recent versions of the framework https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/mitigation-ziparchiveentry-fullname-path-separator,

natemcmaster: We are using System.IO.Compression.ZipArchive to decompress this zip on fairly recent .NET Core 2.1 builds.

The problem is that unzipping on Linux or macOS does not create the right directory structure. Instead of folders, you get one folder with filenames like `host\fxr\2.1.0-preview1-26108-01\hostfxr.dll`

natemcmaster:
@JunTaoLuo build tools 15664 has a fix for the slash problems in UnzipArchive

JunTaoLuo:
@natemcmaster Updated to new build tools and the cumulative archives are not generated correctly on all platforms.

build/SharedFxInstaller.targets
weshaggard: Ah I see you guys are extracting windows zip files on unix. I suspect that is why no one else has complained about this. Hopefully you guys can work around this but if not please file an issue on core-setup and we will see what we can do.

natemcmaster:
> are not generated correctly on all platforms.

What's still wrong?

build/SharedFxInstaller.targets
natemcmaster: Ok. https://github.com/dotnet/core-setup/issues/3577

JunTaoLuo:
Oops that's a typo not => now. The archives are generated correctly.

natemcmaster:
haha, ok.

:shipit:

PR #764, opened 2018-01-09
Use platform assemblies paths instead of app paths for crossgen

JunTaoLuo:
From our email discussions, it seems like we would rather use platform_assemblies_paths instead of app_paths though theoretically they should be equivalent in our use case.

### DIFF
--- a/build/SharedFx.props
+++ b/build/SharedFx.props
@@ -14,6 +14,9 @@
     <_InstallerSourceDir>$(RepositoryRoot).deps\Installers\</_InstallerSourceDir>
     <_DockerRootDir>/opt/code/</_DockerRootDir>
     <_InstallersOutputDir>$(ArtifactsDir)installers\</_InstallersOutputDir>
+    <!-- 3B = semicolon in ASCII -->
+    <PathSeparator Condition="'$(PathSeparator)' == ''">:</PathSeparator>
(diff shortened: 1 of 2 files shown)


natemcmaster:
Ah, just saw it, buried in the bottom of my inbox. Adding here for posterity....

> See https://docs.microsoft.com/en-us/dotnet/core/tutorials/netcore-hosting “Step 5 - Prepare AppDomain settings”
·       APP_PATHS This is a list of paths to probe in for an assembly if it can't be found in the trusted platform assemblies (TPA) list. These paths are meant to be the locations where users' assemblies can be found. In a sandboxed AppDomain, assemblies loaded from these paths will only be granted partial trust. Common APP_PATH paths include the path the target app was loaded from or other locations where user assets are known to live.
Even the dotnet.exe host doesn’t set `app_paths` by default when running a .NET Core app.  See https://github.com/dotnet/core-setup/issues/193.

JunTaoLuo:
Also
> The only “difference” between these two args for crossgen, that I know of, would be it would search App_Paths first when looking for an assembly, then it would search Platform_Assemblies_Paths.  But you can put multiple paths on Platform_Assemblies_Paths, and it will search in the order specified.  So I’m not certain why’d you want/need to use App_Paths.

PR #768, opened 2018-01-10
Add SignalR to the meta-package

anurse:
This also fixes an upcoming build failure because I merged https://github.com/aspnet/SignalR/pull/1274 and forgot I needed to do this :)

@JunTaoLuo Is this all that's needed to add SignalR to the metapackage?

Fixes aspnet/MetaPackages#235


anurse:
Fixes https://github.com/aspnet/MetaPackages/issues/235

PR #770, opened 2018-01-11
Be verbose about missing items

ryanbrandenburg:
When we're looking to an artifact that's not in there it fails, but doesn't tell us what package we were looking for.


build/tasks/AddMetapackageReferences.cs
```diff
@@ -46,9 +46,18 @@ public override bool Execute()
...
+                string packageVersion;
+                try
+                {
+                    packageVersion = buildArtifacts
+                        .Single(p => string.Equals(p.PackageInfo.Id, packageName, StringComparison.OrdinalIgnoreCase))
+                        .PackageInfo.Version.ToString();
+                }
+                catch (InvalidOperationException)
+                {
+                    Log.LogMessage(MessageImportance.High, $"Missing Package: {packageName} from build artifacts");
```
natemcmaster: LogError?

ryanbrandenburg:
:up::date:

PR #774, opened 2018-01-11
Remove Dispatcher packages

jbagga:
As part of aspnet/Home#2741

PR #775, opened 2018-01-11
fix artifacts, external deps and update SignalR

anurse:
* Unreverts `952dbae68df4b0e49909d201e6e81df466b8316a` which adds new packages to SignalR build
* Add SignalR to metapackage and add entries for new SignalR packages to `artifacts.props`
* Update SignalR submodule to remove use of private package System.Buffers.Primitives.


natemcmaster:
Looks like you'll need @jbagga's changes too: https://github.com/aspnet/Universe/pull/774

ryanbrandenburg:
@jbagga merged, @natemcmaster he may as well not wait for appveyor since it's already broken without this anyway right?

natemcmaster:
This needs to rebased on dev. As long as either Travis or AppVeyor passes, :shipit:

Issue #662, opened 2017-11-21
Produce AspNetCore Platform


JunTaoLuo:
- [x] Produce platform by publishing framework dependent app
  - [x] win7-x64, win7-x86, osx-x64, linux-x64
  - [x] Update artifacts (naming and trimming) including deps.json, runtimeconfig.json and .version files
  - [x] Create optimized binaries using cross-gen 
  - [x] Create symbols
  - [x] Create installers
- [x]  M.A.All package
  - [x] Remove manifests
  - [x] Update publish targets to set `MicrosoftNETPlatformLibrary`
  - [x] Errors
  - [x] Enforce dependent version ranges

natemcmaster:
FYI - I'm working on https://github.com/aspnet/Coherence-Signed/issues/690 next

PR #777, opened 2018-01-11
Add TFM compatibility check to 2.1 metapackage

JunTaoLuo:
Last work item in https://github.com/aspnet/Universe/issues/662.

I'm wondering if we should backport this to the 2.0 metapackage and make it only compatible with netcoreapp2.0. 

I'm adding an initial target that should be run before build or publish. Is there a better approach?


JunTaoLuo:
I tested this approach in 2.0 to simulate tfm upgrade. When I create a netcoreapp2.1 app that uses the 2.0 .All metapackage with this error added, the following error is produced during build:

image

Note that I changed the error message but otherwise the mechanism is the same.

PR #778, opened 2018-01-11
mark System.Threading.Channels non-private and update SignalR

anurse:
SignalR update removes System.IO.Pipelines.Extensions dependency as well.


JunTaoLuo:
Is it possible to see if there are further issues by building universe locally?

Issue #773, opened 2018-01-11
Fix and start running RuntimeStore-MusicStore tests.


ryanbrandenburg:
We need to get these running again. They'll have to depend on `SetupSharedFramework` and `SharedFx_Installers` instead of `Coherence-Signed` so that we have the shared framework.

JunTaoLuo:
Dup of https://github.com/aspnet/Universe/issues/706

Issue #713, opened 2017-12-13
Templates tests don't actually run on CI


natemcmaster:
They're currently skipped because our CI splits compilation and testing into separate phases.

We need to also figure out how to enable the tests to run after the shared framework is built.

PR #739, opened 2018-01-03
Refactor Universe build into smaller, more parallel steps

natemcmaster:
Massive refactoring is still WIP, so this PR is still targeting the <kbd>feature/repo-proj</kbd> branch.

Changes:
 - Build each of the following in a separate MSBuild project.
     - Metapackage
     - RuntimeStore
     - LZMA
     - Templating
     - All "normal" aspnet repos
     - Runtime store installers
 - Add a solution file named "Microsoft.AspNetCore.sln" (not for actual use in VS). It has the following solution configurations/platforms
    - Debug/linux-x64
    - Debug/osx-x64
    - Debug/win-x64
    - Debug/win-x64
    - Release/linux-x64
    - Release/osx-x64
    - Release/win-x64
    - Release/win-x64
 - Make the metapackage and runtimestore builds faster by generating a nuspec, instead of generating an MSBuild project, calling restore, and then invoking msbuild /t:Pack to generate the nuspec.
 - Add dotnet-archive to this repo (yes, I hate myself for this. This is temporary until we fix https://github.com/aspnet/Universe/issues/595).

cc @pakrym  @JunTaoLuo @pranavkm

PR #779, opened 2018-01-12
Split compilation and tests into separate phases

### LINKED ISSUE #742: Add a switch to continue running universe even when tests fail
And have a report of failed tests when finished.

/cc @natemcmaster @ryanbrandenburg

natemcmaster:
Resolves #742, prerequisite for fixing MusicStore #773 and Template tests #713 

Before:

```
Common: build
Common: test
DI: build
DI: test
FileSystem: build
FileSystem: test   <--- fails and build.cmd halts
```

After:
```
Common: build
DI: build
FileSystem: build
Kestrel: build
...etc..

Common: test
DI: test
FileSystem: test <--- fails
Kestrel: test <-- tests should continue chugging
...etc...
```

Also, enables running: `build.cmd /t:test /p:NoBuild=true` to skip re-building altogether and to jump ahead to tests.


build/RepositoryBuild.targets
```diff
@@ -84,4 +94,26 @@
...
 
+  <Target Name="_TestRepository" DependsOnTargets="GetRepoBuildProps">
+    <PropertyGroup>
+      <BuildArguments>/t:Test /p:NoBuild=$(_NoBuildRepos) $(RepositoryBuildArguments)</BuildArguments>
+
+      <_BuildScriptToExecute Condition="'$(OS)'!='Windows_NT'">build.sh</_BuildScriptToExecute>
+      <_BuildScriptToExecute Condition="'$(OS)'=='Windows_NT'">build.cmd</_BuildScriptToExecute>
+    </PropertyGroup>
+
+    <Message Text="============ Testing $(RepositoryToBuild) ============" Importance="High" />
```
pakrym: Do we have to copy? Doesn't Korebuild have an argument which repo to build?

pakrym: + Seems to be duplicated

natemcmaster: No, KoreBuild does not have an argument that allows you to build one repo. It would if we did #739

natemcmaster: Updated to reduce some duplication between these targets

pakrym:
LGTM

PR #784, opened 2018-01-18
Add doc for common questions about working with git-submodules

natemcmaster:
Things I get asked frequently.

cc @pakrym @ryanbrandenburg


docs/Submodules.md
```diff
@@ -0,0 +1,53 @@
+Submodules
+==========
+
+This repository composes ASP.NET Core repos from a collection of many submodules.
+Working with submodules in git introduces some quirks into an otherwise workflow.
```
ryanbrandenburg: > into an otherwise workflow.

I think you accidentally a word.

docs/Submodules.md
```diff
@@ -0,0 +1,53 @@
...
+    git clone https://github.com/aspnet/Universe.git --recursive
+
+If you have already cloned, run this to initialize all submodules.
+
+    git submodule update --init
+
+## Pulling updates
+
+When you execute `git pull`, submodules do not automatically snap to the version
+used in new commits of the master repo. Update all modules to match the master repo
```
ryanbrandenburg: Is "master repo" the correct term? I thought it was "parent repo". Just to avoid confusion with "master branch".

docs/Submodules.md
```diff
@@ -0,0 +1,53 @@
...
+
+    git submodule foreach '<command>'
+
+For example, to clean and reset each submodule:
+
+    git submodule foreach 'git reset --hard; git clean -xfd'
+
+## Updating submodules
+
+Updating all submodules to newer versions can be done like this.
```
ryanbrandenburg: How is the newer version determined in this case? Does it just take the latest of the branch for the submodule?

docs/Submodules.md
natemcmaster: I'm not really sure, Parent repo sounds better so I'll update .

docs/Submodules.md
natemcmaster: Yup. it uses the branch as configured in .gitmodules

PR #786, opened 2018-01-19
Update FileSystem submodule and apply the result of the cascading version

natemcmaster:
This includes a bunch of scripted updates to repos to apply cascading versions

cc @Eilon @muratg

### DIFF
--- a/build/artifacts.props
+++ b/build/artifacts.props
@@ -11,13 +11,57 @@
 
   <ItemGroup>
     <PackageArtifact Include="Microsoft.AspNet.Identity.AspNetCoreCompat" Category="ship" />
+    <PackageArtifact Include="Microsoft.AspNetCore.Antiforgery" Category="ship" Metapackage="true" RuntimeStore="true"/>
+    <PackageArtifact Include="Microsoft.AspNetCore.ApplicationInsights.HostingStartup" Category="ship" Metapackage="true" RuntimeStore="true" HostingStartup="true" />
+    <PackageArtifact Include="Microsoft.AspNetCore.Authentication.Abstractions" Category="ship" Metapackage="true" RuntimeStore="true"/>
+    <PackageArtifact Include="Microsoft.AspNetCore.Authentication.Cookies" Category="ship" Metapackage="true" RuntimeStore="true"/>
+    <PackageArtifact Include="Microsoft.AspNetCore.Authentication.Core" Category="ship" Metapackage="true" RuntimeStore="true"/>
+    <PackageArtifact Include="Microsoft.AspNetCore.Authentication.Facebook" Category="ship" Metapackage="true" RuntimeStore="true"/>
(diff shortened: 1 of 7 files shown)


natemcmaster:
> how do we verify the packages included in a patch

Shifting repos from "Shipped" to re-build is manual, as is updating whether a package is considered an external dependency or an output of the current build. We opened https://github.com/aspnet/Coherence-Signed/issues/738 to automate this calculation, but we haven't implemented it yet.

Fortunately, we have written a bunch of tests that check we've done things right. I don't know if we've reviewed lately that we're making the right assumptions, so I've jotted down everything we take into consideration.

The rules I follow are:
1. every package that is produced by the build must be listed in build/artifacts.props
1. every PackageReference in our csproj must point to a package that is either listed in build/artifacts.props or matches exactly one of the external dependencies in build/dependencies.props
1. don't patch a package unless required
1. changes that require patching a package
   A. the transitive rule: approved updates to external dependencies or our own packages that would cause `<dependencies>` in nuspec, or cause the compilation output to change
   B. source change: approved changes to source code
1. everything in a submodule revs together. If one component in the repo ships, all other packages in that repo ship too
1. When bumping a submodule version, increment patch by one.
1. Microsoft.AspNetCore.All, runtime store, and symbols always patch. It's version matches the version of Microsoft.NETCore.App
1. Templates always update (b/c of the previous rule)

Rule 4A is the trickiest one. We have a few tests that verify this is mostly correct:

1. Coherent check. All packages produced must reference the same versions of packages.
1. Cascading repo version check. We compute the repo-to-repo graph. A "shipped" repository cannot depend on a repo that is patching.
1. Unknown dependency check - we fail the build if 1 or 2 are not satisfied

Some gaps:
 - 4A an update to a library that is "privateassets=all" could cause the compilation output to change. For example, NETStandard.Library is PrivateAssets=all so it's not in the nuspec. But if this package updated and caused a change to the references baked into an assembly, we should probably rebuild. IIRC the change between 2.0.0 and 2.0.1 was to fix .NET Framework references. Fortunately, our packages are almost all netstandard2.0 so I don't think they need an update....but we should probably verify that manually.
- We don't automatically verify if an external dependency has changed from the last version and should cause a rebuild. We manually scan for usages of external dependencies when they update.
- The manual computation is tedious. I mark everything as "shipped", and then iteratively add things in as "patching". The tests produce errors like "because A is patching, B must patch too". Repeat until the checks are happy.

natemcmaster:
> This is scripted now? Can you share/productize your script?

The part that is scriptable is in scripts/PatchVersionPrefix.ps1.

Some of the other changes were one-time fixes that I scripted. Those included:
 - Repos for which this is their first patch since 2.0.0. We built 2.0.0 using "old" korebuild, before we had things like SourceLink. So I had to add properties and settings to Directory.Build.props to update this.
 - Unifying the variable name to NETStandardLibrary20PackageVersion (we had 3 variants across our stack)
 - Fix tests in DotNetTools to use correct package versions. (Test break in private builds.)
 - Update korebuild.json to point to the "release/2.0" channel instead of the "rel/2.0.2" channel.

PR #789, opened 2018-01-19
Mirror System.Net.WebSockets.WebSocketProtocol

anurse:
We wants it. We needs it. Must have the precious.

Not sure what else needs to be updated or how to verify this... also, the version number is slightly ahead of the current CoreFX version (this package was literally uploaded an hour ago). Not sure if I need to do anything about that. My hope is we can get this mirrored ASAP so I can quickly port our WebSocket middleware to use it and slide that in to Preview 1 :)


build/dependencies.props
```diff
@@ -175,6 +175,7 @@
     <SystemThreadingTasksDataflowPackageVersion>4.9.0-preview1-26112-01</SystemThreadingTasksDataflowPackageVersion>
     <SystemThreadingTasksExtensionsPackageVersion>4.5.0-preview1-26112-01</SystemThreadingTasksExtensionsPackageVersion>
     <SystemValueTuplePackageVersion>4.5.0-preview1-26112-01</SystemValueTuplePackageVersion>
+    <SystemNetWebSocketsWebSocketProtocolPackageVersion>4.5.0-preview1-26119-06</SystemNetWebSocketsWebSocketProtocolPackageVersion>
```
natemcmaster: Any reason you chose this version? To keep it aligned with other corefx packages, would 4.5.0-preview1-26112-01 work?

anurse: There is no 4.5.0-preview1-26112-01 of this package :). It was introduced in 26119

anurse: That's what my comment about the version number above is about. Would it be best to wait for 26119 to be mirrored? Will this file be auto-updated by that? Happy to wait on that and rebase, just not clear on the best approach.

pakrym: Or you can just update other corefx packages too ;)

pakrym:
I would prefer all packages to be updated all together because threre are some scenarios where using WebSocket package might cause an unexpected upgrade of other dependencies.

anurse:
Wasn't sure if that was something that should be left to the automation. Happy to do so.

pakrym:
There is unfortunately no automation, just try building universe with all  corefx packages updated to the same version

anurse:
> just try building universe



<small>fine</small>

anurse:
I found the magic dependency version incantations. Universe is passing locally now. Ready to merge on approval

PR #791, opened 2018-01-19
Add EntityFramework.Proxies package

ajcvickers:


natemcmaster:
> error KRB4005: Undefined external dependency on Castle.Core/4.2.1 

Looks like this is a new external dependency. Contact for @Eilon for OSS approval, and add this to build/external-dependencies.props

ajcvickers:
@natemcmaster Already got approval. I'll add it.

build/external-dependencies.props
```diff
@@ -342,6 +342,7 @@
       <NoWarn>KRB2004</NoWarn>
       <VariableName>VisualStudio_NewtonsoftJsonPackageVersion</VariableName>
     </ExternalDependency>
+    <ExternalDependency Include="Castle.Core" Version="$(CastleCorePackageVersion)" Source="$(DefaultNuGetFeed)" />
```
natemcmaster: Looks like you also need to define CastleCorePackageVersion. We keep the versions list in build/dependencies.props

PR #794, opened 2018-01-23
Fix .All SharedFx WIP

JunTaoLuo:


JunTaoLuo:
Verified that a new mvc app can now run on shared frameworks with this fix.

Issue #790, opened 2018-01-19
Add ProjectCapability for ANCM In-process hosting to .App metapackage


shirhatti:
This is independent of whether we chose to make in-process the default behavior. That is tracked separately by  #766  

```xml
<Project>
  <!--
      Capability that enables Visual Studio supoort for hosting Asp.Net Core applications in the IIS process
    -->
  <ItemGroup>
    <ProjectCapability Include="AspNetInProcessHosting" />
  </ItemGroup>

</Project>
```

cc @BillHiebert since tooling is going to depend on this.
I'll make a similar issue in IISIntegration as well

PR #796, opened 2018-01-23
Add project capability for ANCM In-process hosting.

jkotalik:
For #790. Currently leaving the AspNetCoreModuleHostingModel commented as a reminder to enable it for 2.1-preview2.
cc/ @shirhatti


jkotalik:
@natemcmaster did you want it directly in the props file or in a separate file?

Issue #802, opened 2018-01-25
Set manifest info properties on PushToBlobFeed


pranavkm:
Requirement for prodcon that we need to get done for preview1. See https://github.com/dotnet/core-eng/issues/2404

pranavkm:
FYI @muratg

Issue #803, opened 2018-01-25
Building LZMA with DotnetTool packages fails 


natemcmaster:
We currently build the LZMA by generating a csproj file with PackageReference and DotNetCliToolReference, restoring it, and zipping up the cache folder. This fails with recent changes to NuGet that enforce restrictions on packageType.

```
C:\b\w\33bdfc1cae7b2a38\.w\Archive.csproj : error NU1202: Package dotnet-dev-certs 2.1.0-preview1-28172 is not compatible with netcoreapp2.1 (.NETCoreApp,Version=v2.1). Package dotnet-dev-certs 2.1.0-preview1-28172 does not support any target frameworks.
[13:51:04]C:\b\w\33bdfc1cae7b2a38\.w\Archive.csproj : error NU1212: Invalid project-package combination for dotnet-dev-certs 2.1.0-preview1-28172. DotnetToolReference project style can only contain references of the DotnetTool type 
[13:51:04]C:\b\w\33bdfc1cae7b2a38\.w\Archive.csproj : error NU1202: Package dotnet-sql-cache 2.1.0-preview1-28172 is not compatible with netcoreapp2.1 (.NETCoreApp,Version=v2.1). Package dotnet-sql-cache 2.1.0-preview1-28172 does not support any target frameworks.
[13:51:04]C:\b\w\33bdfc1cae7b2a38\.w\Archive.csproj : error NU1212: Invalid project-package combination for dotnet-sql-cache 2.1.0-preview1-28172. DotnetToolReference project style can only contain references of the DotnetTool type 
[13:51:04]C:\b\w\33bdfc1cae7b2a38\.w\Archive.csproj : error NU1202: Package dotnet-watch 2.1.0-preview1-28172 is not compatible with netcoreapp2.1 (.NETCoreApp,Version=v2.1). Package dotnet-watch 2.1.0-preview1-28172 does not support any target frameworks.
[13:51:04]C:\b\w\33bdfc1cae7b2a38\.w\Archive.csproj : error NU1212: Invalid project-package combination for dotnet-watch 2.1.0-preview1-28172. DotnetToolReference project style can only contain references of the DotnetTool type 
[...]

PR #804, opened 2018-01-25
Include metadata when publishing to blob

pranavkm:
Fixes https://github.com/aspnet/Universe/issues/802


korebuild-lock.txt
```diff
@@ -1,2 +1,2 @@
-version:2.1.0-preview2-15683
-commithash:54af97ac56dfdc495c2514c6916fe4ca79910939
+version:2.1.0-preview2-15684
```
dagood: I assume this has an upgraded version of the Feed package?

build/Publish.targets
```diff
@@ -175,15 +175,23 @@
                     AccountKey="$(PublishBlobFeedKey)"
                     ItemsToPush="@(PackagesToPublishToTransport)"
                     Overwrite="$(PushToBlobFeed_Overwrite)"
+                    ManifestBuildId="$(BuildNumber)"
+                    ManifestCommit="$(CommitHash)"
+                    ManifestName="https://github.com/aspnet/Home"
```
dagood: I think the `/`s from the url will make this not work--it's used to name a blob and I haven't plumbed through escaping. When I said repo name in [publish.md](https://github.com/dotnet/core-eng/blob/master/Documentation/Orchestrated-Build/Api/publish.md#cookbook-using-pushtoblobfeed-to-generate-a-manifest) I only meant the last bit, so `Home` in this case... which isn't very descriptive. I didn't consider repos that primarily use the organization to distinguish themselves, sorry.

I think `aspnet-Home` would be reasonable, but `aspnet-Universe` might be best. The `CommitHash` is for Universe, right?

korebuild-lock.txt
pranavkm: Yup

build/Publish.targets
pranavkm: We'll use `aspnet` for now. `CommitHash` points to the current repo viz Universe.

build/Publish.targets
```diff
@@ -175,15 +175,23 @@
...
                     MaxClients="$(PushToBlobFeed_MaxClients)"
+                    SkipCreateManifest="false"
```
natemcmaster: Why is this false?

build/Publish.targets
```diff
@@ -175,15 +175,23 @@
                     AccountKey="$(PublishBlobFeedKey)"
                     ItemsToPush="@(PackagesToPublishToTransport)"
```
natemcmaster: We should also update the ManifestArtifactData metadata on these items to `NonShipping=true` for our noship packages

build/Publish.targets
pranavkm: I looked at what they had in their other PRs e.g. https://github.com/dotnet/templating/pull/1400/files. Assumed this needed to be set.

build/Publish.targets
pranavkm: Ahh, oops

build/Publish.targets
natemcmaster: Ok. I assume we need it then.

cc @dagood should we call this out in https://github.com/dagood/core-eng/blob/orchestration/Documentation/Orchestrated-Build/Api/publish.md?

dagood: It's [optional and defaults to `false`](https://github.com/dotnet/buildtools/blob/cd0812d52c432a415fb625bca5dd790adc8136a9/src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs#L44), no need for this line. :smile: The core\* repos don't include it.

dagood: That said, I'll note it in the doc so the property isn't a mystery.

pranavkm:
🆙 📅

build/Publish.targets
```diff
@@ -112,16 +112,23 @@
         <Overwrite>true</Overwrite>
       </FilesToPublish>
 
-      <PackagesToPublish Include="$(DependencyPackagesDir)*.nupkg" Exclude="$(DependencyPackagesDir)*.symbols.nupkg" />
-      <PackagesToPublish Include="$(DependencyPackagesDir)*.symbols.nupkg" Exclude="@(PackagesToPublish)" IsSymbolsPackage="true" />
+      <_SymbolArtifactPath Include=">$(DependencyPackagesDir)%(PackageArtifact.Identity).*$(BuildNumber).nupkg" Artifact="%(PackageArtifact.Identity)" />
+      <_PackageArtifactPath Include=">$(DependencyPackagesDir)%(PackageArtifact.Identity).*$(BuildNumber).nupkg" Exclude="@(_SymbolArtifactPath)" Artifact="%(PackageArtifact.Identity)" />
```
pranavkm: @natemcmaster this would be a fair assumption to make, no?

build/Publish.targets
```diff
@@ -112,16 +112,23 @@
...
     </ItemGroup>
 
     <Message Text="@(FilesToPublish -> '%(FullPath) -> %(RelativeBlobPath)','%0A')" />
   </Target>
 
   <Target Name="_CheckFilesToPublish">
     <ItemGroup>
-      <_MissingFiles Include="%(FilesToPublish.Identity)" Condition=" ! Exists(%(FilesToPublish.Identity))" />
+      <_MissingFiles Include="%(FilesToPublish.Identity)" Condition="!Exists(%(FilesToPublish.Identity))" />
+      <_MissingFiles Include="%(PackagesToPublish.PackagePath)" Condition="!Exists(%(PackagesToPublish.PackagePath)) AND '%(PackagesToPublish.Category)' == 'ship'" />
```
pranavkm: My guess is I shouldn't check the category here. The packages dir should have all the packages, no?

build/Publish.targets
```diff
@@ -175,6 +189,9 @@
                     AccountKey="$(PublishBlobFeedKey)"
                     ItemsToPush="@(PackagesToPublishToTransport)"
                     Overwrite="$(PushToBlobFeed_Overwrite)"
+                    ManifestBuildId="$(BuildNumber)"
+                    ManifestCommit="$(CommitHash)"
+                    ManifestName="aspnet"
```
dagood: Oh, if you have a property laying around with the branch name in it, it would be good to pass that to `ManifestBranch`. It is optional, though.

pranavkm: We pass in an environment variable, but I'd rather not use it. I think the commit sha does a much better job at it and we're guaranteed to include it.

natemcmaster: +1 - I'd prefer not to include branch name.

dagood: That's ok with me.

build/Publish.targets
natemcmaster: For most builds it would be okay, but when IsFinalBuild=true, BuildNumber won't be in the file name.

To split these apart, we probably need to use the `ArtifactInfo` item group, which includes info about package ID + version. You could then use that to compute all the file paths you need `$(DependencyPackagesDir)%(ArtifactInfo.PackageId).$(ArtifactInfo.Version).nupkg`. Unfortunately, that doesn't contain package category due to #686, so you would need to join PackageArtifact and ArtifactInfo. That's hard to do in MSBuild, so I have a task for such a thing:
https://github.com/aspnet/Universe/blob/namc/graph-build/build/tasks/JoinItems.cs
Example usage:
https://github.com/aspnet/Universe/blob/6029a73327d3eb886b3c2d3957a06b8d2450c069/build/targets/Repository.Common.targets#L166-L170

I'm about to add this task to release/2.1 because I need it to solve #803

build/Publish.targets
natemcmaster: Yes, it should. We still want to publish shipoob and noship packages to the transport feed - they might be used by downstream partners, even if they aren't intended for nuget.org

build/Publish.targets
pranavkm: Gotcha. Would you like to pick this up once you've made the changes? Would be easier if one of one partied around with BuildTools + this repo changes?

natemcmaster: If you're okay waiting for me to fix #803 (almost ready to open a PR with a fix), lets do that. Otherwise, feel free to just copy https://github.com/aspnet/Universe/blob/namc/graph-build/build/tasks/JoinItems.cs to your PR. I'm going to add it verbatim to my PR to fix 803

pranavkm: Sure, I think we can wait a bit. @dagood we should be ready with this change by tomorrow morning.

dagood: Sounds good, thanks for the update!

pranavkm:
🆙 📅 

CI change I had to make to make for this - http://aspnetci/admin/settingsDiffView.html?id=buildType:Releases_21Public_Publish&versionBefore=3&versionAfter=4&actionId=47615

pranavkm:
@dagood could you kick off a preview1 build? It'd be good to verify things work E2E. I'll port this over to dev once we get a confirmation.

natemcmaster:
😬 oops, I merged release/2.1 to dev already to pick up another change I made to universe....

dagood:
I don't think I have an orchestrated build id that's parked at aspnet right now. I started a fresh 2.1 build, but it will take a while to get to aspnet and might take some retries depending on reliability right now: https://devdiv.visualstudio.com/DevDiv/Default/_build/index?buildId=1325284

pranavkm:
@natemcmaster no worries 👍

Issue #811, opened 2018-01-26
EnsureTFMCompatibility check should not run when IsCrossTargetingBuild = true


pranavkm:
Repro steps

1. dotnet new mvc
2. dotnet run
3. Edit the csproj to change `<TargetFramework>netcoreapp2.1</TargetFramework>` to `<TargetFrameworks>netcoreapp2.1</TargetFrameworks>`
4. Run dotnet restore

Expected: Things work
Actual: This version of Microsoft.AspNetCore.All is only compatible with the netcoreapp2.1 target framework.

pranavkm:
@muratg \ @JunTaoLuo it will be good to take a fix for this in preview1. The workaround involves wiping the obj directory before proceeding which is just bad experience.

JunTaoLuo:
Yes, I'm looking into fixing this for preview1. Testing fix locally right now.

Issue #810, opened 2018-01-26
Microsoft.AspNetCore.All does not include any target \ prop files from packages


pranavkm:
We explicitly exclude build items from .App and it's dependencies. This was done to avoid picking up the App's props file that sets the TFM. Since NuGet ensures targets and props are imported in dependency order, we don't need to do this - the .All package's props file will get imported after .App and should override any configured setting.

pranavkm:
cc @muratg \ @JunTaoLuo

muratg:
@JunTaoLuo Was this by design?  Anything we need to do for preview1?

pranavkm:
This essentially results in the Razor.Design \ Razor.Compilation targets not being imported. We should fix this for preview1.

JunTaoLuo:
Oops, I commented on the wrong issue, but this is the actual bug I'm working on testing.

Eilon:
We need to make sure this works in .App and .All right?

Approved for preview1.

PR #812, opened 2018-01-27
Update metapackages

JunTaoLuo:
.All will include all assets from .App
.All and .App will check TFM compatibility only when IsCrossTargetingBuild is not set
Remove unused runtime store templates

Addresses #810 #811


src/Microsoft.AspNetCore.App/build/netcoreapp2.1/Microsoft.AspNetCore.App.targets
```diff
@@ -0,0 +1,7 @@
+<Project InitialTargets="EnsureTFMCompatibility">
```
JunTaoLuo: Missed creating this check when I separated the shared framework into two.

src/Microsoft.AspNetCore.All/build/netcoreapp2.1/Microsoft.AspNetCore.All.targets
```diff
@@ -2,6 +2,6 @@
   <Target Name="EnsureTFMCompatibility">
     <Error
       Text="This version of Microsoft.AspNetCore.All is only compatible with the netcoreapp2.1 target framework."
-      Condition="'$(TargetFramework)' != 'netcoreapp2.1'"/>
+      Condition="'$(TargetFramework)' != 'netcoreapp2.1' AND '$(IsCrossTargetingBuild)' != 'true'"/>
```
JunTaoLuo: This concerns me a little. It means that you can bypass this check by using <SharedFrameworks> instead of <SharedFramework>. We might need a different mechanism to key off of.

natemcmaster: What about `Condition=" '$(TargetFramework)' != 'netcoreapp2.1' AND '$(TargetFramework)' != '' "`?

build/tools/templates/RS.References/RS.References.csproj
```diff
@@ -1,22 +0,0 @@
-<Project Sdk="Microsoft.NET.Sdk">
```
natemcmaster: Unintentional deletion? or are these not needed?

JunTaoLuo: These were used for RuntimeStore builds (hence the RS prefix) and apparently I forgot to delete them in a previous PR.

src/Microsoft.AspNetCore.All/build/netcoreapp2.1/Microsoft.AspNetCore.All.targets
JunTaoLuo: I'll give that a try.

src/Microsoft.AspNetCore.App/build/netcoreapp2.1/Microsoft.AspNetCore.App.targets
```diff
@@ -0,0 +1,7 @@
...
+  <Target Name="EnsureTFMCompatibility">
+    <Error
+      Text="This version of Microsoft.AspNetCore.App is only compatible with the netcoreapp2.1 target framework."
```
pranavkm: Have we considered expanding the error message to say how to resolve this?

src/Microsoft.AspNetCore.App/build/netcoreapp2.1/Microsoft.AspNetCore.App.targets
pranavkm: Out of curiosity, is there a reason we do this as opposed to wiring up to a specific target such as `BeforeBuild`?

src/Microsoft.AspNetCore.App/build/netcoreapp2.1/Microsoft.AspNetCore.App.targets
JunTaoLuo: I'll add one.

src/Microsoft.AspNetCore.App/build/netcoreapp2.1/Microsoft.AspNetCore.App.targets
JunTaoLuo: Not particularly, but is there anything wrong with this approach?

pranavkm: There aren't very strong reasons, but the one thing I can think of is that there really is no way for a user to turn it off (short of deleting the target) by manipulating the csproj. Just a thought.

natemcmaster: Would be good to see how this behaves in VS and VS Code. InitialTargets will run in the design-time builds that power the UI in VS and Omnisharp in VSCode

JunTaoLuo:
:up::date: @natemcmaster  and I spoke offline and it seems like InitialTargets will create pitfalls where users will not be able to update a wrong version of .All to a compatible one. We decided depending on the target
 `_CheckForInvalidConfigurationAndPlatform` would be better as this target is run before every build/rebuild and is run in the inner build which has the correct `TargetFramework` set from the list of `TargetFrameworks`

PR #814, opened 2018-01-29
Add ShipInstaller=dotnetcli to non-package blobs

pranavkm:


build/Publish.targets
```diff
@@ -69,25 +69,30 @@
...
         <ContentType>text/plain</ContentType>
       </FilesToPublish>
 
       <!-- Archive installers -->
       <FilesToPublish Include="$(DependencyAssetsDir)$(InstallerBaseFileName)-%(NativeInstaller.Identity)%(NativeInstaller.FileExt)" Condition=" '%(NativeInstaller.FileExt)' != '' ">
         <RelativeBlobPath>$(BlobBasePath)$(InstallerBaseFileName)-%(NativeInstaller.Identity)%(NativeInstaller.FileExt)</RelativeBlobPath>
+        <ManifestArtifactData>ShipInstaller=dotnetcli</ManifestArtifactData>
       </FilesToPublish>
 
       <!-- Support for README badges and dotnet-install.ps1/sh -->
```
dagood: I think these should be `ShipInstaller=dotnetcli` too, so that once they can be published, they make it where they need to be.

Another option is to remove it from `LatestRuntimeVersionFile` and `SharedFxVersionBadge` and rely on the final publish step, which is still WIP. (https://github.com/dotnet/buildtools/pull/1882 + https://github.com/dotnet/buildtools/pull/1871)

PR #819, opened 2018-01-30
Update versions to include prodcon build id

natemcmaster:
Should make PB builds on TC more obvious.
Also updates TC to display the actual product number produced

Incidentally fixes issue with Microsoft.AspNetCore.Analyzers which is producing packages as "0.1.0-alpha1" without build number.


run.sh
```diff
@@ -272,6 +280,10 @@ if [ ! -z "$access_token_suffix" ]; then
     msbuild_args[${#msbuild_args[*]}]="-p:DotNetAssetRootAccessTokenSuffix=$access_token_suffix"
 fi
 
+if [ ! -z "$product_build_id" ]; then
+    msbuild_args[${#msbuild_args[*]}]="-p:DotNetProductBuildId=pb-$product_build_id"
```
smitpatel: Is `pb-` suffix here intentional?

natemcmaster: nope, good catch

PR #826, opened 2018-01-30
Update publish to include more metadata in the manifest

### LINKED ISSUE #822: Packages aren't being published to transport feed anymore
Possibly related or due to #804. Package aren't being published to our transport feed anymore

cc @pranavkm

natemcmaster:
Depends on https://github.com/aspnet/BuildTools/pull/567

- Update the publish task
- Include metadata for the manifest - cc @dagood 
- Ensure packages are actually published. Resolves #822


build/tasks/RepoTasks.tasks
```diff
@@ -20,4 +22,15 @@
   <UsingTask TaskName="RepoTasks.ResolveHostingStartupPackages" AssemblyFile="$(_RepoTaskAssembly)" />
   <UsingTask TaskName="RepoTasks.TrimDeps" AssemblyFile="$(_RepoTaskAssembly)" />
   <UsingTask TaskName="RepoTasks.VerifyCoherentVersions" AssemblyFile="$(_RepoTaskAssembly)" />
+
+  <!-- tools from dotnet-buildtools -->
+  <PropertyGroup>
+    <NuGetPackagesPath>$(NUGET_PACKAGES)</NuGetPackagesPath>
+    <NuGetPackagesPath Condition=" '$(NuGetPackagesPath)' == '' AND '$(USERPROFILE)' != '' ">$(USERPROFILE)\.nuget\packages\</NuGetPackagesPath>
```
pranavkm: Should UserProfile \ Home be normalized as well?

build/tasks/dotnet-buildtools.props
```diff
@@ -0,0 +1,5 @@
+<Project>
+  <PropertyGroup>
+    <DotNetBuildToolsVersion>2.1.0-prerelease-02430-04</DotNetBuildToolsVersion>
```
pranavkm: Can we put this in dependencies.props as `DevDependency_DotNetBuildToolsVersion`? Same for the other package versions listed in RepoTasks.

build/tasks/RepoTasks.tasks
natemcmaster: It should be normalized by line 31 (two down)

build/Publish.targets
```diff
@@ -200,7 +201,8 @@
                     AccountKey="$(PublishBlobFeedKey)"
                     ItemsToPush="@(PackagesToPublishToTransport)"
                     Overwrite="$(PushToBlobFeed_Overwrite)"
-                    ManifestBuildId="$(BuildNumber)"
+                    ManifestBuildId="$(Version)"
+                    ManifestBuildData="ProductVersion=$(PackageVersion);UniverseCommitHash=$(CommitHash)"
```
dagood: Out of curiosity, what's `Version` vs. `PackageVersion` vs. `BuildNumber`?

If `UniverseCommitHash` is kinda to open the way to have multiple hashes listed in one Build element, originally my thinking was that each "root" repo (which may contain submodules) would be one build. But I know now there are multiple "roots" that go into one aspnetci build, so this makes sense. Unfortunately the build metadata (`ManifestBuildId`, `ManifestBuildData`, ...) only gets set on the first update, but that could be fixed to start merging it if you're looking to add more Build metadata over the course of the build.

natemcmaster: So, for our final builds, the values might look like this:
```
BuildNumber    = 1234
Version        = 2.1.0-preview1-1234
PackageVersion = 2.1.0-preview1-final
```

Re: UniverseCommitHash

I wasn't looking to add more at the moment. I was just looking to make it explicit in the manifest what "Commit" means.

dagood: Just to make sure, the `2.1.0-preview1-final` matches what you'll be publishing? That is, each final build will need to overwrite every aspnet asset on `dotnetcli`?

natemcmaster:
Ran this on my own azure storage taccount to verify it works: here is the manifest generated:

[aspnet.xml.txt](https://github.com/aspnet/Universe/files/1679478/aspnet.xml.txt)

dagood:
Thanks for posting it! Looks like what we need.

build/Publish.targets
natemcmaster: Yes, 2.1.0-preview1-final is the "product friendly version", the version of nupkgs, installers, etc. `Version` is the more like a build ID. It represents additional info that is sometimes baked in to assembly attributes and other things.

PR #827, opened 2018-01-30
Mark Logging.Analyzers as shipoob so it gets signed

pranavkm:


pranavkm:
Need this for preview1.

natemcmaster:
Is Microsoft.AspNetCore.Analyzers also shipoob?

PR #835, opened 2018-01-31
Only build in UpdateRepos script

pakrym:
Right now we guard updating repos on tests passing with intention of non-breaking tests by updating. 

But I think it has opposite effect - devs working on repositories often do not see test failures cause by their new changes in their day to day work and have a smaller chance of catching them while universe tests continue to fail.

This change disables tests in UpdateRepos configuration.

/cc @Eilon


ryanbrandenburg:
We discussed this yesterday in engineering triage. We decided instead to aggressively disable (and subsequently fix and re-enable) flaky tests instead, and leave the tests running to catch behavioral breaks.

pakrym:
But they are running in Universe already. Problem is that no one is noticing that their repo didn't update versions. But people would notice when tests would start failing.

natemcmaster:
I think we can do both this and what @ryanbrandenburg was asked to do. Our automation moves repos forward. If upgrade broke something, this helps elevate the visibility of issues

Issue #836, opened 2018-02-01
Flatten Microsoft.AspNetCore.All dependencies


JunTaoLuo:
Currently Microsoft.AspNetCore.All references Microsoft.AspNetCore.App which transitively references other aspnet components: Hosting, Kestrel, Mvc, etc. As per offline discussion yesterday, we would like to flatten the dependencies instead so that Microsoft.AspNetCore.All directly references all of its dependencies and remove its reference to Microsoft.AspNetCore.App.

cc @DamianEdwards @Eilon since this is a preview1 change

DamianEdwards:
The key here is to ensure we don't change the dependency graph of the existing meta-package (.All) in this release, as moving dependencies to the different levels of the tree can introduce behavior changes.

JunTaoLuo:
What do you mean by no change in the dependency graph of the existing .All metapackage? I did a diff of the 2.0.5 and the new 2.1.0-preview1 .All metapackages and ensured nothing was removed. However, in 2.1 we are adding references to new packages, for example Microsoft.AspNetCore.Identity.Service.*. I'm pretty sure this is fine but just wanted to check.

natemcmaster:
Related questions:
 - do we flatten third-party dependencies too?
 - do we also version-cap third-party dependencies?

If yes to both, my concern here is that we will be locking all aspnetcore users everywhere to third-party dependencies we don't control. `<dependency id="Newtonsoft.Json" version="[10.0.1]" />` seems dangerous.

DamianEdwards:
@JunTaoLuo I mean the depth of the graph should be the same, e.g. don't hoist anything higher or push anything lower. This can have semantic differences and affect how something is restored.

@natemcmaster no, in line with above, we should not be flattening anymore than we did before. Version caps **is** in *Microsoft.AspNetCore.App* but I agree we should likely not introduce them now in .All to ensure the behavior for folks using it remains the same.

JunTaoLuo:
Ah, let me remove the version caps in .All

Issue #820, opened 2018-01-30
Remove BrowserLink from Microsoft.AspNetCore.App


DamianEdwards:
Let's remove the Browser Link package from *Microsoft.AspNetCore.App* and ensure it's added to the appropriate (i.e. VS) templates directly.

JunTaoLuo:
@DamianEdwards Should BrowserLink be added to Microsoft.AspNetCore.All?

PR #837, opened 2018-02-01
Flatten dependencies of Microsoft.AspNetCore.All metapackage

JunTaoLuo:
Remove BrowserLink from Microsoft.AspNetCore.App

Addresses #836 https://github.com/aspnet/Universe/issues/820. 

Note @pranavkm this would mean that we can revert https://github.com/aspnet/Mvc/pull/7253. 

cc @muratg @Eilon

### DIFF
--- a/build/repo.targets
+++ b/build/repo.targets
@@ -202,8 +202,8 @@
     <Error Text="Invalid configuration of %(PackageArtifact.Identity). Packages marked as AllMetapackage='true' must be Category='ship'."
            Condition="'%(PackageArtifact.Category)' != 'ship' AND '%(PackageArtifact.AllMetapackage)' == 'true' " />
 
-    <Error Text="Invalid configuration of %(PackageArtifact.Identity). Packages cannot be marked as AppMetapackage='true' and AllMetapackage='true'."
-           Condition="'%(PackageArtifact.AppMetapackage)' == 'true' AND '%(PackageArtifact.AllMetapackage)' == 'true' " />
+    <Error Text="Invalid configuration of %(PackageArtifact.Identity). Packages marked as AppMetapackage='true' must also be marked as AllMetapackage='true'."
+           Condition="'%(PackageArtifact.AppMetapackage)' == 'true' AND '%(PackageArtifact.AllMetapackage)' != 'true' " />
   </Target>
 
   <Target Name="VerifyCoherentVersions" DependsOnTargets="ResolveRepoInfo">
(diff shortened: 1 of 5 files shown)


build/repo.targets
```diff
@@ -202,8 +202,8 @@
     <Error Text="Invalid configuration of %(PackageArtifact.Identity). Packages marked as AllMetapackage='true' must be Category='ship'."
            Condition="'%(PackageArtifact.Category)' != 'ship' AND '%(PackageArtifact.AllMetapackage)' == 'true' " />
 
-    <Error Text="Invalid configuration of %(PackageArtifact.Identity). Packages cannot be marked as AppMetapackage='true' and AllMetapackage='true'."
-           Condition="'%(PackageArtifact.AppMetapackage)' == 'true' AND '%(PackageArtifact.AllMetapackage)' == 'true' " />
+    <Error Text="Invalid configuration of %(PackageArtifact.Identity). Packages marked as AppMetapackage='true' must also be marked as AllMetapackage='true'."
```
natemcmaster: Seems like you could just automatically assume AllMetapackage == true if AppMetapackage== true

JunTaoLuo: I considered that but I thought it would be best to keep it as explicit as possible. This way, in artifacts props, it's obvious when a package goes into both .App and .All and there's no "magic" where AppMetapackages are also included in AllMetapackage.

build/tasks/AddMetapackageReferences.cs
```diff
@@ -18,6 +18,9 @@ public class AddMetapackageReferences : Task
         [Required]
         public string MetapackageReferenceType { get; set; }
 
+        [Required]
+        public bool VersionLock { get; set; }
```
pranavkm: `LockToExactVersions`?

build/tasks/AddMetapackageReferences.cs
```diff
@@ -61,11 +64,14 @@ public override bool Execute()
...
-                Log.LogMessage(MessageImportance.High, $" - Package: {packageName} Version: [{packageVersion}]");
+
+                var packageVersionValue = VersionLock ? $"[{packageVersion}]" : packageVersion;
+
+                Log.LogMessage(MessageImportance.High, $" - Package: {packageName} Version: {packageVersionValue}");
 
                 var packageReferenceElement = xmlDoc.CreateElement("PackageReference");
                 packageReferenceElement.SetAttribute("Include", packageName);
-                packageReferenceElement.SetAttribute("Version", $"[{packageVersion}]");
+                packageReferenceElement.SetAttribute("Version", $"{packageVersionValue}");
```
pranavkm: Don't need an interpolated string here.

PR #840, opened 2018-02-01
Include SpaServices.Extensions in .App and .All metapackages

JunTaoLuo:
This should fix the VSTS issue 555879 where SpaServices.Extensions is missing from the offline scenarios.


Eilon:
Logged this bug to track it: https://github.com/aspnet/Universe/issues/842

PR #844, opened 2018-02-02
[2.0.x] Add new WsFederation package and dependencies

Tratcher:
This is a new package that's shipping out of band. We're now merging it into the patch branch for long term servicing. If 2.0.6 re-spins then we'll include it.

I'll update the submodule once we confirm if this is going into 2.0.6.


natemcmaster:
Tagging in @Eilon to make sure we're clear about whether this goes in to 2.0.6 or 2.0.7

Eilon:
2.0.6 is the plan. It only *builds* as part of the patch so we can sim-ship. Not part of .All or anything like that.

Tratcher:
Updated the submodule.

Issue #848, opened 2018-02-02
LZMA is missing Microsoft.AspNetCore.App


natemcmaster:
2.1.0-preview1-28226 is missing Microsoft.AspNetCore.App from the offline cache.

https://github.com/dotnet/versions/tree/master/build-info/dotnet/product/cli/release/2.1

cc @JunTaoLuo @Eilon

natemcmaster:
had to revert #849.

muratg:
👀

JunTaoLuo:
Build passed. Closing

PR #849, opened 2018-02-02
Include Microsoft.AspNetCore.App in the LZMA

JunTaoLuo:
Forgot to include the .App metapackage to the LZMA after removing it as a dependency of .All in https://github.com/aspnet/Universe/pull/837. This broke offline scenarios where templates are now using .App as the default metapackage.

cc @DamianEdwards @Eilon as this is required for preview1.


JunTaoLuo:
Addresses #848

Issue #850, opened 2018-02-05
Publish symbols to prodcon feed


pranavkm:
Here are notes on what the symbol publish step will be looking for in the manifest:https://github.com/dotnet/core-eng/blob/master/Documentation/Orchestrated-Build/Api/publish.md#publish-symbol-packages-to-symbol-server. (Something that might not be clear is that it’s looking for Blob elements, not Packages.)

*Placeholder for what do with profiling (.ni.pdb) pdbs produced as part the build.*

pranavkm:
cc @natemcmaster \ @JunTaoLuo \ @dagood

ryanbrandenburg:
We confirmed that this made its way onto the Prodcon feed.

ctaggart:
@ryanbrandenburg Is the prodcon feed something can be access externally/publicly?

ryanbrandenburg:
It is not. To my knowledge there has not been a decision on where/how/if to make symbols publicly available.

ctaggart:
Is there another issue for making the publicly available? If not, can you make one and link it here?

dagood:
The ProdCon feed is available for each build, for example this is the 2.1.1 build: https://github.com/dotnet/versions/blob/726018548f536bb1176c25792a6bf18f88b5d989/build-info/dotnet/product/cli/release/2.1/build.xml. That said, it's not easy to navigate these and ProdCon feeds aren't terribly useful as a symbol source.

When we release a build, symbols should be available on `https://msdl.microsoft.com/download/symbols`. In Visual Studio, this is the default location called "Microsoft Symbol Servers".

PR #851, opened 2018-02-05
Publish symbol packages to transport feed

ryanbrandenburg:
This should publish symbols packages to the transport feed on ProdCon builds as per #850.


natemcmaster:
Requesting help from @dagood. Guidance on how to use PushToBlobFeed to push symbols.nupkg files is unclear to me.

build/Publish.targets
```diff
@@ -73,6 +74,12 @@
         <ManifestArtifactData>ShipInstaller=dotnetcli</ManifestArtifactData>
       </FilesToPublish>
 
+      <SymbolPackagesToPublish Include="$(DependencySymbolsDir)\**\*.symbols.nupkg" />
+
+      <FilesToPublishToTransport Include="%(SymbolPackagesToPublish.FullPath)" >
+        <RelativeBlobPath>symbols/%(SymbolPackagesToPublish.FileName)%(SymbolPackagesToPublish.Extension)</RelativeBlobPath>
```
dagood: This looks like it will work--if I'm reading it right, the path ends up being `assets/symbols/...` by the time it hits the publish task. But...

The way I was mentioning in email is that you could add the symbol packages to `PackagesToPublishToTransport` (or whatever you end up passing into the publish task). `PushToBlobFeed` [detects symbol packages by filename, filters them out from the "normal" packages, and puts it at the right relative blob path automatically](https://github.com/dotnet/buildtools/blob/2a9486a85cb385042e26ada50da9370f996f110d/src/Microsoft.DotNet.Build.Tasks.Feed/PushToBlobFeed.cs#L95-L100). I'd suggest going that way because it means you don't need to code in the `assets/symbols/` path here.

ryanbrandenburg: Cool, so I can just remove line 80 and `PushToBlobFeed` will do the right thing.

ryanbrandenburg:
:up::date:

build/Publish.targets
dagood: Not quite, the autodetection only works in the `PublishFlatContainer=false` (default) case, the one used to push NuGet packages (and NuGet symbol packages). This itemgroup goes to the task with `PublishFlatContainer=true`, which is for any assets you want but doesn't do detection. (In theory I don't see any reason it *shouldn't* do detection if `RelativeBlobPath` isn't defined, but it just doesn't right now.)

I think the additional change needed is changing `FilesToPublishToTransport` to `PackagesToPublishToTransport`.

dagood: Oh, I didn't refresh and Nate beat me to it. :+1:

ryanbrandenburg:
:up::date:

build/Publish.targets
```diff
@@ -73,6 +74,10 @@
         <ManifestArtifactData>ShipInstaller=dotnetcli</ManifestArtifactData>
       </FilesToPublish>
 
+      <SymbolPackagesToPublish Include="$(DependencySymbolsDir)\**\*.symbols.nupkg" />
```
natemcmaster: I think at this point, the item groups can be reduced to one since we don't need to transform metadata anymore.

ryanbrandenburg:
@natemcmaster just a double check that I understood you correctly on the latest update here.

PR #852, opened 2018-02-05
Notify which package is missing

ryanbrandenburg:
This should prevent unhelpful messages like [this](http://aspnetci/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=400608&_focus=333).

### DIFF
--- a/build/tasks/AddArchiveReferences.cs
+++ b/build/tasks/AddArchiveReferences.cs
@@ -55,23 +55,27 @@ namespace RepoTasks
 
                 string packageVersion;
 
-                if(string.Equals(packageName, "Microsoft.AspNetCore.All", StringComparison.OrdinalIgnoreCase))
+                if (string.Equals(packageName, "Microsoft.AspNetCore.All", StringComparison.OrdinalIgnoreCase)
+                   || string.Equals(packageName, "Microsoft.AspNetCore.App", StringComparison.OrdinalIgnoreCase))
                 {
                     packageVersion = MetaPackageVersion;
                 }
                 else
                 {
-                    try
-                    {
-                        packageVersion = buildArtifacts
-                            .Single(p => string.Equals(p.PackageInfo.Id, packageName, StringComparison.OrdinalIgnoreCase))
-                            .PackageInfo.Version.ToString();
-                    }
-                    catch (InvalidOperationException)
+                    var packageVersionCandidates = buildArtifacts.Where(p => string.Equals(p.PackageInfo.Id, packageName, StringComparison.OrdinalIgnoreCase));
+
+                    if (!packageVersionCandidates.Any())
                     {
                         Log.LogError($"Missing Package: {packageName} from artifacts archive.");
-                        throw;
+                        return false;
                     }
+                    else if (packageVersionCandidates.Count() > 1)
+                    {
+                        Log.LogError($"Duplicate Packages: {packageName} from artifacts archive.");
+                        return false;
+                    }
+
+                    packageVersion = packageVersionCandidates.Single().PackageInfo.Version.ToString();
                 }
                 Log.LogMessage(MessageImportance.High, $" - Package: {packageName} Version: {packageVersion}");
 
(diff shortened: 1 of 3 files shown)


build/tasks/AddArchiveReferences.cs
```diff
@@ -52,12 +52,27 @@ public override bool Execute()
...
                 var packageName = package.ItemSpec;
-                var packageVersion = string.Equals(packageName, "Microsoft.AspNetCore.All", StringComparison.OrdinalIgnoreCase) ?
-                    MetaPackageVersion :
-                    buildArtifacts
-                        .Single(p => string.Equals(p.PackageInfo.Id, packageName, StringComparison.OrdinalIgnoreCase))
-                        .PackageInfo.Version.ToString();
 
+                string packageVersion;
+
+                if(string.Equals(packageName, "Microsoft.AspNetCore.All", StringComparison.OrdinalIgnoreCase))
```
natemcmaster: @JunTaoLuo remind me again why this is special cased? And does this need to be updated to take into consideration that we have 2 metapackages?

JunTaoLuo: Yup, I think it's gonna need to include .App as well. The reason this is here is that the metapackages are not included in the build artifacts collected from modules so it wasn't possible to determine what version should be referenced when building the LZMA. To resolve this, we pass in the metapackage version directly from the msbuild invocation.

natemcmaster: Makes since. This is probably why the PackageArchive build step is currently failing.

JunTaoLuo: My bad, totally forgot about this.

ryanbrandenburg:
@natemcmaster do we want to take this in addition to your revert of the current problem?

JunTaoLuo:
I pushed a commit that should fix the archiver build but I can't test locally since my laptop lost access to the drop share so I can't access the archiver tool. However, I did verify that AddArchiveReference now succeeds.

JunTaoLuo:
This PR now addresses #848

build/tasks/AddArchiveReferences.cs
```diff
@@ -52,12 +52,28 @@ public override bool Execute()
...
                 var packageName = package.ItemSpec;
-                var packageVersion = string.Equals(packageName, "Microsoft.AspNetCore.All", StringComparison.OrdinalIgnoreCase) ?
-                    MetaPackageVersion :
-                    buildArtifacts
-                        .Single(p => string.Equals(p.PackageInfo.Id, packageName, StringComparison.OrdinalIgnoreCase))
-                        .PackageInfo.Version.ToString();
 
+                string packageVersion;
+
+                if (string.Equals(packageName, "Microsoft.AspNetCore.All", StringComparison.OrdinalIgnoreCase)
```
natemcmaster: Are there any other areas in the code that have hard-coded this package name? I'm concerned now that we're probably missing other places where we need to add .App too.

build/tasks/AddArchiveReferences.cs
```diff
@@ -52,12 +52,28 @@ public override bool Execute()
...
+                {
+                    try
+                    {
+                        packageVersion = buildArtifacts
+                            .Single(p => string.Equals(p.PackageInfo.Id, packageName, StringComparison.OrdinalIgnoreCase))
+                            .PackageInfo.Version.ToString();
+                    }
+                    catch (InvalidOperationException)
+                    {
+                        Log.LogError($"Missing Package: {packageName} from artifacts archive.");
```
natemcmaster: I have a slight preference for changing the call from `.Single` to a `.Where()`. InvalidOperationException could thrown in two conditions: 0 packages or more than 1 package with that ID.

JunTaoLuo: So that we can differentiate between package not found and too many packages found?

build/tasks/AddArchiveReferences.cs
JunTaoLuo: I did a search in the Universe repo and i found another in lineups. Not sure what else might be affected. I'm not sure if there are any other repos affected. I also took a look at Coherence-Signed but didn't find anything.

build/tasks/AddArchiveReferences.cs
natemcmaster: yeah, that's where i was going with that. Update the error to either say "Missing package" or "Multiple packages" found. I think missing package is the most likely cause of the error right now, but that could always change.

JunTaoLuo:
:up::date:

PR #855, opened 2018-02-07
Fix #785 - publish nuget packages and the shared runtime at the same time

### LINKED ISSUE #785: Lineups and Runtime should publish at the same time
Currently if Universe-Coherence passes but Publish fails we can get into a state where someone who does "./run.ps1 upgrade deps" will have new lineup packages, but fail when the Runtime can't be found. We should publish the Lineup at the same time as we do the Runtime/SDK (which is currently Publish) so that we always have a working coherent set of artifacts published.

@natemcmaster @muratg  and @pranavkm have apparently discussed this before.

natemcmaster:
When builds of package succeed, but the runtime fails to build, we can end up in state where the lineup package on myget is newer than the available shared runtime.

This delays publishing until we have _everything_ built. It also aligns closer with how we publish for prodcon builds.

Corresponding change to coherence-signed:  https://github.com/aspnet/Coherence-Signed/pull/769

depends on https://github.com/aspnet/SignalR/pull/1416

### DIFF
--- a/build/Publish.targets
+++ b/build/Publish.targets
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <DependencyAssetsDir>$(RepositoryRoot).deps\assets\</DependencyAssetsDir>
     <DependencyPackagesDir>$(RepositoryRoot).deps\packages\</DependencyPackagesDir>
-    <DependencySymbolsDir>$(RepositoryRoot).deps\symbols\</DependencySymbolsDir>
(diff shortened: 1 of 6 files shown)


natemcmaster:
Yes, I needed https://github.com/aspnet/SignalR/pull/1416, which allows Universe to acquire the ArtifactInfo items for npm packages without building signalR.

PR #859, opened 2018-02-08
Removes the token server packages from preview1

javiercn:


javiercn:
@pranavkm Is there anything else to do here?

Issue #853, opened 2018-02-06
Produce symbol packages for Microsoft.AspNetCore.All and Microsoft.AspNetCore.App


JunTaoLuo:
https://github.com/aspnet/Universe/issues/850

This specifically addresses the symbols requirements for the metapackages

PR #870, opened 2018-02-09
Create aspnetcore shared framework symbols packages

JunTaoLuo:
Addresses https://github.com/aspnet/Universe/issues/853

Checking effects on downstream builds.

### DIFF
--- a/build/SharedFx.targets
+++ b/build/SharedFx.targets
@@ -284,8 +284,17 @@
 
     <ItemGroup>
       <IgnoredAssemblies Include="@(AppPublishAssemblies);@(AllPublishAssemblies)" Condition="'%(AssetType)' == 'native' OR '%(AssetType)' == 'resources'" />
+      <AppRuntimeAssemblies Include="@(AppPublishAssemblies)" Condition="'%(AssetType)' == 'runtime'">
+        <PortablePDB>%(RootDir)%(Directory)%(Filename).pdb</PortablePDB>
+      </AppRuntimeAssemblies>
+      <AllRuntimeAssemblies Include="@(AllPublishAssemblies)" Exclude="@(AppRuntimeAssemblies)" Condition="'%(AssetType)' == 'runtime'">
+        <PortablePDB>%(RootDir)%(Directory)%(Filename).pdb</PortablePDB>
+      </AllRuntimeAssemblies>
+      <OtherAssemblies Include="@(AppPublishAssemblies);@(AllPublishAssemblies)" Exclude="@(IgnoredAssemblies);@(AppRuntimeAssemblies);@(AllRuntimeAssemblies)" />
       <_AssembliesToCrossgen Include="$(SharedFxIntermediateOutputPath)**\*.dll" />
     </ItemGroup>
+
+    <Error Text="Unaccounted shared framework assemblies found: @(OtherAssemblies). Assemblies must be included as runtime assemblies or marked as ignored." Condition="'@(OtherAssemblies)' != ''" />
 
     <!-- Compute the intersection of crossgen candidates and native/resources assemblies as the set of assemblies to skip crossgen -->
     <CreateItem Include="@(_AssembliesToCrossgen)" Condition="'%(Filename)' != ''and '@(IgnoredAssemblies)' != ''">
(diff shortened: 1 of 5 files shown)


build/SharedFx.targets
```diff
@@ -383,4 +422,45 @@
   </Target>
 
   <Target Name="BuildSharedFx" DependsOnTargets="ResolveSharedFxFiles;CrossGenAssemblies;CrossGenSymbols;PackSharedFx"/>
+
+  <Target Name="_BuildSharedFxSymbols">
+    <!-- Clear working directory -->
+    <RemoveDir Directories="$(_WorkRoot)" />
+
+    <ItemGroup>
+      <RuntimeSymbolsArchive Include="win-x86;win-x64;osx-x64;linux-x64"/>
```
natemcmaster: This is already defined in repo.props.

build/SharedFx.targets
```diff
@@ -383,4 +422,45 @@
...
+    </ItemGroup>
+
+    <!-- Extract Symbols archives -->
+    <UnzipArchive
+      File="$(_SymbolsSourceDir)$(SymbolsSourceArchiveName)-%(RuntimeSymbolsArchive.Identity).zip"
```
natemcmaster: Can you clarify what's going on here? Why are we producing a zip of symbols, unzipping it, and then packing it into a nuspec?

build/tools/templates/SharedFxSymbols/Microsoft.AspNetCore.All.symbols.nuspec
```diff
@@ -0,0 +1,22 @@
...
+    <id>Microsoft.AspNetCore.All.symbols</id>
+    <version>$VERSION$</version>
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <licenseUrl>http://www.microsoft.com/web/webpi/eula/net_library_eula_ENU.htm</licenseUrl>
+    <projectUrl>http://www.asp.net/</projectUrl>
+    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <description>Symbol packages for Microsoft.AspNetCore.All shared framework</description>
+    <copyright>© .NET Foundation. All rights reserved.</copyright>
```
natemcmaster: Can you update copyright and licenseUrl to match our runtime packages? 

https://github.com/aspnet/BuildTools/blob/dev/src/Internal.AspNetCore.Sdk/build/Common.props#L13-L22

I'd actually prefer to flow these automatically from Internal.AspNetCore.Sdk so we don't have to remember to update more places.

build/tools/templates/SharedFxSymbols/Microsoft.AspNetCore.All.symbols.nuspec
```diff
@@ -0,0 +1,22 @@
...
+    <tags>aspnetcore</tags>
+    <serviceable>true</serviceable>
+    <repository type="git" url="https://github.com/aspnet/Universe" />
+  </metadata>
+  <files>
+    <file src="*.pdb" target="" />
```
natemcmaster: Should these be nested according to RID? Are there symbols for multiple RIDs in the package?

build/SharedFx.targets
JunTaoLuo: The symbols are built on each platform win-x64/win-x86/linux-x64/osx-x64. Then we need another configuration that takes the symbols produced on each platform and pack them into a symbols nupkg for .App and .All.

natemcmaster: Maybe we should rid split our symbols packages... I'm thinking ahead to our goal to build aspnetcore on a single machine. If we can avoid introducing more build outputs that require building on 3 machines and still satisfy the symbols package requirement, I think we should do that.

build/tools/templates/SharedFxSymbols/Microsoft.AspNetCore.All.symbols.nuspec
JunTaoLuo: How do we automatically flow these from Internal.AspNetCore.Sdk? Add them all to properties passed into PackNuspec?

natemcmaster: Build the symbols nupkg from a MSBuild project that references Internal.AspNetCore.Sdk.

JunTaoLuo:
:up::date:

PR #874, opened 2018-02-12
Re-enable templates tests on CI

natemcmaster:
Resolves https://github.com/aspnet/Universe/issues/713

CI changes required:

1. Tests will need an artifact dependency on the "Publish" step which gathers the shared runtime installers into a folder that matches the dotnetcli.blob.core.windows.net/dotnet feed.
2. Set `/p:InstallSharedRuntimeFromPreviousBuild=true`
3. Set `/p:AspNetCoreFxFeed=%teamcity.build.workingDir%\.deps`


build/Templating.targets
```diff
@@ -43,4 +45,35 @@
...
+    <!-- To enable this test, either publish the shared runtime to https://dotnetcli.blob.core.windows.net/dotnet, or override the install location by setting AspNetCoreFxFeed. -->
+    <Warning Text="Skipping template tests because InstallSharedRuntimeFromPreviousBuild != 'true'." Condition="'$(InstallSharedRuntimeFromPreviousBuild)' != 'true'"/>
+
+    <PropertyGroup>
+      <_BuildTemplateProjProperties>
+        $(TemplateProjProperties);
+        SkipAspNetCoreRuntimeInstall=true;
+        DotNetRestoreSourcesPropsPath=$(GeneratedRestoreSourcesPropsPath);
+        DotNetPackageVersionPropsPath=$(GeneratedPackageVersionPropsPath);
+        SkipTests=false;
```
pranavkm: Does this need to be explicitly set to get the tests to run?

build/Publish.targets
```diff
@@ -180,9 +180,9 @@
     <Error Text="Missing expected packages from PackageToPublish. These were defined as expected PackageArtifact's in artifacts.props: %0A - @(_MissingPackageArtifact, '%0A - ')" Condition="@(_MissingPackageArtifact->Count()) != 0" />
   </Target>
 
-  <Target Name="CopyToPublishArtifacts" DependsOnTargets="_CheckFilesToPublish">
-    <MakeDir Directories="$(ArtifactsDir)files\" />
-    <Copy SourceFiles="@(FilesToPublish)" DestinationFolder="$(ArtifactsDir)files\" />
+  <Target Name="CopyToPublishArtifacts" DependsOnTargets="">
```
pranavkm: Remove `DependsOnTargets`?

build/Publish.targets
```diff
@@ -180,9 +180,9 @@
...
+    <MakeDir Directories="$(ArtifactsDir)$([System.IO.Path]::GetDirectoryName($([System.String]::new('%(FilesToPublish.RelativeBlobPath)'))))" Condition="'%(FilesToPublish.RelativeBlobPath)' != ''" />
```
pranavkm: Wouldn't `Copy` take care of creating the dir structure?

natemcmaster: Yup, i always forget it does that. Updated

PR #876, opened 2018-02-13
Fix typo in property name - DotNetRestoreSourcePropsPath

natemcmaster:
Restore fails on CI because of typo in property name. This is the same bug @pakrym noticed in release/2.1. 

DotNetRestoreSourcePropsPath, not DotNetRestoreSourcesPropsPath.


pakrym:
Why are we making a separate change for this? Merging rel/2.1 would fix this.

PR #879, opened 2018-02-13
Split dependency versions into pinned vs automatically updated/overridden

natemcmaster:
For ProdCon: helps make it clearer in code which versions can be automatically set via the orchestrated build and not.

@pranavkm @ryanbrandenburg  - no versions are changed, although I did notice that we can prune EF's pinned version of system.data.sqlclient now that they don't build benchmarks in Universe.

cc @mmitche


natemcmaster:
Related issues:

https://github.com/aspnet/Universe/issues/577
https://github.com/aspnet/BuildTools/issues/463

natemcmaster:
:bell:

Issue #873, opened 2018-02-12
Ship Microsoft.AspNetCore.AspNetCoreModule


shirhatti:
It is currently marked as noship.
https://github.com/aspnet/Universe/blob/release/2.1/build/artifacts.props#L37

cc @muratg

Tratcher:
This is only for preview. File an additional issue to un-ship it for RTM.

natemcmaster:
Not sure I agree. This seems to violate the no-turds rule: we are not planning to ship this in RTM, so don't push it to NuGet.org. Since this package isn't really meant to be a "package" (i.e. you wouldn't use it as a PackageReference), have you considered other shipping mechanisms?

shirhatti:
The options we have
1. Ship an msi. We refrained from doing this since shipping an MSI that's part of a ref-counted exe bundle causes a whole slew of SxS installation failures down the road. We've made that mistake once with ANCM already
2. Ship a NuGet package on an existing feed. We can't use the https://dotnet.myget.org/gallery/aspnetcore-release feed since it's volatile preview2 bits will land up on that feed as well.
3. Ship a NuGet package on a new feed.
4. Ship a NuGet package on NuGet.org
5. Ship a NuGet package/zip archive via dist.asp.net

Of all these options we landed on 3. I don't feel too strongly about and if you think one of these options is head and shoulders above the rest, I'm happy to reconsider

Tratcher:
Number four you mean...

natemcmaster:
I don't understand MSI's at all so not sure I understand why this is off the table, but isn't installing ANCM _without_ an MSI also going to cause issues? Manually copying aspnetcore.dll into a global location seems potentially more dangerous.

You might consider this feed instead: https://dotnet.myget.org/gallery/aspnetcore-master

natemcmaster:
aspnetcore-master is stable. We haven't cleared it in years.

muratg:
Approved for preview1.

@Eilon / @DamianEdwards FYI this is regarding the ANCM package that we plan to ship to nuget, only for preview1. This is our distribution mechanism for ANCM in-proc features in preview1.

Eilon:
I'm ok with whatever. We have a plan to remove this after preview1 and I fear it'll be more complex to go some other route.

natemcmaster:
Ok. This is in for preview1. Should be in tomorrows prodcon build.

PR #880, opened 2018-02-13
[2.1.0-preview1] Mark Microsoft.AspNetCore.AspNetCoreModule as shipping

natemcmaster:
Resolves https://github.com/aspnet/Universe/issues/873


natemcmaster:
See also https://github.com/aspnet/IISIntegration/pull/585

PR #882, opened 2018-02-14
Add Polly.Signed to lineup

rynowak:


rynowak:
@Eilon - I believe I emailed you about this like 6 months ago? I'll wait for your OK enough though I think I already have it 👍

Issue #892, opened 2018-02-15
[Temporary for preview1]  Push corefxlab packages to the "ship" location


muratg:
These packages are moved to CoreFX repo for preview2, but for preview1, we'll need to make sure we push these packages.

Let's make the change temporarily for preview1.

muratg:
cc @Eilon FYI

mikeharder:
I believe this is the list of packages:
* System.Buffers.Primitives
* System.IO.Pipelines
* System.Text.Encodings.Web.Utf8
* System.Text.Primitives

natemcmaster:
Done in #894

Issue #893, opened 2018-02-15
Undo #892 for Preview2


muratg:
Remove the temporary measure from #892 in preview2.  By preview2, the impacted packages would be in corefx properly and build with their system.

muratg:
cc @Eilon FYI

PR #894, opened 2018-02-15
[2.1-preview1] Publish corefxlab packages to the transport feed

natemcmaster:
Resolves https://github.com/aspnet/Universe/issues/892

This will push the following packages from corefxlab into prodcon output, and eventually, to nuget.org.

 - System.Buffers.Primitives
 -  System.IO.Pipelines
 - System.Text.Encodings.Web.Utf8
 - System.Text.Primitives

For preview1 only! This should be reverted for preview2 as these packages should be coming from dotnet/corefx. Using #893 to ensure this is reverted.

cc @muratg @ahsonkhan @pakrym @dagood @mmitche @joshfree

### DIFF
--- a/build/Publish.targets
+++ b/build/Publish.targets
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <DependencyAssetsDir>$(RepositoryRoot).deps\assets\</DependencyAssetsDir>
     <DependencyPackagesDir>$(RepositoryRoot).deps\packages\</DependencyPackagesDir>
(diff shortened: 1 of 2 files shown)


joshfree:
> Can you do a no-op merge in to dev, so we're guaranteed not to have this checked in for preview2?

Yes, we don't want a dependency on System.Text.Encodings.Web.Utf8 remaining accidently - as that assembly isn't actually shipping in 2.1, this is just a 2.1-preview1 temporary dependency.

cc: @GrabYourPitchforks

natemcmaster:
Yup, I'll make sure this doesn't make it back to preview2

PR #906, opened 2018-02-17
Don't ship global CLI tool packages to nuget.org or put into LZMA

natemcmaster:
Mark dotnet-watch and dotnet-dev-certs as OOB and remove them from the LZMA.

These tools are now bundled in the .NET Core CLI. Since they cannot be overridden by dotnet-install-tool, there isn't any point in shipping them via NuGet.org.

See https://github.com/dotnet/cli/pull/8606

FYI @muratg @javiercn @danroth27 @glennc @wli3

Part of https://github.com/aspnet/DotNetTools/issues/396


javiercn:
Sounds good to me, but can we chat about this on Tuesday? I've been a bit disconnected and want to make sure I understand what the plan is here.

natemcmaster:
I'm not 100% sure we are clear on our plan. I've started an email thread internally to make sure we clear that up. I wasn't expecting https://github.com/dotnet/cli/pull/8606 to merge so quickly and without clear approval from anyone on our team.

natemcmaster:
:up: :date: per discussion, we are going to make all ASP.NET DotnetTool packages built-in commands in the CLI. So, these packages no longer need to ship to nuget.org or end up in the LZMA. (cc @JunTaoLuo )

PR #910, opened 2018-02-20
Remove MsgPack from .All metapackage

JunTaoLuo:
Spoke to @anurse online and he believes msgpack support should not be in the .All metapackage.

FYI @DamianEdwards


JunTaoLuo:
Merging immediately to fix the build.

PR #911, opened 2018-02-21
Adapt update dependencies script to Maestro

ryanbrandenburg:


scripts/UpdateDependencies.ps1
```diff
@@ -62,6 +66,21 @@ foreach ($varName in ($variables.Keys | sort)) {
 if ($count -gt 0) {
     Write-Host -f Cyan "Updating $count version variables in $depsPath"
     SaveXml $dependencies $depsPath
+
+    # Ensure dotnet is installed
+    Invoke-Expression -Command "$PSScriptRoot\..\run.ps1 install-tools"
```
natemcmaster: It doesn't appear to me that you actually need `Invoke-Expression`.
```ps1
& "$PSScriptRoot\..\run.ps1" install-tools
```

scripts/update-dependencies/Config.cs
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
natemcmaster: nit: aspnetcore source code typically uses the apache license.
https://github.com/aspnet/Universe/blob/c213def07af73590c5f3df0b68f60fc8b66682ea/build/tasks/AddArchiveReferences.cs#L1-L2

scripts/update-dependencies/Program.cs
```diff
@@ -0,0 +1,68 @@
...
+                int idx = arg.IndexOf('=');
+                if (idx < 0)
+                {
+                    Console.Error.WriteLine($"Unrecognized argument '{arg}'");
+                    return false;
+                }
+
+                string name = arg.Substring(0, idx);
+                string value = arg.Substring(idx + 1);
+                Environment.SetEnvironmentVariable(name, value);
```
natemcmaster: Seems kind of messy to use Environment.Set/GetEnvironmentVariable to pass config values around within the app. It seems to me like we could probably use our config builders for this: Microsoft.Extensions.Configuration.{CommandLine,Environment}

scripts/update-dependencies/Program.cs
```diff
@@ -0,0 +1,68 @@
...
+            return true;
+        }
+
+        private static void CreatePullRequest()
+        {
+            GitHubAuth gitHubAuth = new GitHubAuth(config.Password, config.UserName, config.Email);
+            GitHubProject origin = new GitHubProject(config.GitHubProject, config.UserName);
+            GitHubBranch upstreamBranch = new GitHubBranch(config.GitHubUpstreamBranch, new GitHubProject(config.GitHubProject, config.GitHubUpstreamOwner));
+
+            string commitMessage = "Updating external dependencies";
```
natemcmaster: Can we include more information in the commit message and pull request description?

scripts/update-dependencies/update-dependencies.csproj
```diff
@@ -0,0 +1,18 @@
...
+    <RestoreSources>
+      https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json;
+      https://api.nuget.org/v3/index.json;
+    </RestoreSources>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.DotNet.VersionTools" Version="1.0.27-prerelease-01723-01" />
+    <PackageReference Include="System.Diagnostics.TextWriterTraceListener" Version="4.0.0" />
+    <PackageReference Include="NuGet.Packaging" Version="4.3.0-preview2-4095" />
```
natemcmaster: What's this used for?

scripts/UpdateDependencies.ps1
```diff
@@ -62,6 +66,21 @@ foreach ($varName in ($variables.Keys | sort)) {
...
+
+    $ProjectPath = "$PSScriptRoot\update-dependencies\update-dependencies.csproj"
+
+    # Restore the app
+    Write-Host "Restoring App $ProjectPath..."
+    Invoke-Expression "dotnet restore `"$ProjectPath`""
```
natemcmaster: dotnet-run runs dotnet restore, so you can condense this do just `& dotnet run -p $ProjectPath @EnvVars`

scripts/update-dependencies/Program.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            Trace.Listeners.Add(new TextWriterTraceListener(Console.Out));
+
+            if (!ParseArgs(args))
+                return;
+
+            CreatePullRequest();
+        }
+
+        private static bool ParseArgs(string[] args)
```
pranavkm: Can this use CommandLineUtils (the one from NuGet.org would suffice)? Or is the input format dictated by something outside our control?

scripts/update-dependencies/Program.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    public static class Program
+    {
+        private static readonly Config config = Config.Instance;
+
+        public static void Main(string[] args)
+        {
+            Trace.Listeners.Add(new TextWriterTraceListener(Console.Out));
+
+            if (!ParseArgs(args))
```
pranavkm: formatting - braces

scripts/update-dependencies/Program.cs
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.DotNet.VersionTools;
```
pranavkm: Formatting - sort usings with System first

scripts/update-dependencies/Program.cs
```diff
@@ -0,0 +1,68 @@
...
+using Microsoft.DotNet.VersionTools.Automation;
+using Microsoft.DotNet.VersionTools.Dependencies;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
```
pranavkm: Is this used?

scripts/update-dependencies/Program.cs
pranavkm: Why do we do this instead of just setting properties on a `Config` instance?

scripts/update-dependencies/Program.cs
```diff
@@ -0,0 +1,68 @@
...
+                string value = arg.Substring(idx + 1);
+                Environment.SetEnvironmentVariable(name, value);
+            }
+
+            return true;
+        }
+
+        private static void CreatePullRequest()
+        {
+            GitHubAuth gitHubAuth = new GitHubAuth(config.Password, config.UserName, config.Email);
```
pranavkm: `var` all day every day

scripts/update-dependencies/Program.cs
```diff
@@ -0,0 +1,68 @@
...
+            string commitMessage = "Updating external dependencies";
+            string body = string.Empty;
+            if (config.GitHubPullRequestNotifications.Any())
+            {
+                body += PullRequestCreator.NotificationString(config.GitHubPullRequestNotifications);
+            }
+
+            new PullRequestCreator(gitHubAuth, origin, upstreamBranch)
+                .CreateOrUpdateAsync(commitMessage, commitMessage + $" ({upstreamBranch.Name})", body)
+                .Wait();
```
pranavkm: You can target C# 7.1 and make `Main` async.

scripts/update-dependencies/Program.cs
ryanbrandenburg: Bah, my settings didn't carry over when I re-installed VS.

scripts/update-dependencies/Program.cs
ryanbrandenburg: What info would you like? Only thing I can come up with is the versions that were changed. That could go in the description, but not really the PR title or commit message.

natemcmaster: As mentioned in person, I think having the orchestrated build ID as found here would be sufficient:

https://github.com/dotnet/versions/blob/master/build-info/dotnet/product/cli/release/2.1/build.xml#L1. Example contents:
```xml
<OrchestratedBuild Name="cli" BuildId="20180218-02-1397458">
```

scripts/update-dependencies/Program.cs
```diff
@@ -1,68 +1,76 @@
...
 
-            CreatePullRequest();
+            await CreatePullRequest();
+        }
+
+        private static void ParseArgs(string[] args)
+        {
+            var builder = new ConfigurationBuilder().AddCommandLine(args);
+            var configRoot = builder.Build();
+            Config.Instance.Load(configRoot);
```
pranavkm: Remove `Config.Instance` and just have a local `Config` field. Also the only advantage of using `Configuration` here over `CommandLineUtils` is so you can bind it. Consider using that. I'd still prefer using CommandLineUtils here, just seems less unusual as a way to parse command line args

scripts/update-dependencies/Program.cs
```diff
@@ -1,68 +1,76 @@
...
-            if (config.GitHubPullRequestNotifications.Any())
+            using (var client = new HttpClient())
             {
-                body += PullRequestCreator.NotificationString(config.GitHubPullRequestNotifications);
+                var response = await client.GetAsync(xmlUrl);
+                using (var bodyStream = await response.Content.ReadAsStreamAsync())
+                {
+                    var xmlDoc = new XmlDocument();
+                    xmlDoc.Load(bodyStream);
+                    var orcBuild = xmlDoc.GetElementsByTagName("OrchestratedBuild")[0];
```
pranavkm: We should do some error checking here. This will fail with an out of bound index if the tag does not exist which wouldn't be helpful. Have this print the contents of the xml if the tag is empty or missing BuildId.

scripts/update-dependencies/update-dependencies.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.0</TargetFramework>
+    <LangVersion>7.1</LangVersion>
```
pranavkm: 💯

ryanbrandenburg:
:up::date:

scripts/update-dependencies/Config.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Dotnet.Scripts
+{
+    public class Config
+    {
+
```
pranavkm: Extra line

scripts/update-dependencies/Config.cs
```diff
@@ -0,0 +1,26 @@
...
+        public string BuildXml { get; set; }
+        public string GithubUsername {get; set;}
+        public string GithubEmail {get; set;}
+        public string GithubToken {get; set;}
+        public string GithubUpstreamOwner {get; set;} = "aspnet";
+        public string GithubProject {get; set;} = "Universe";
+        public string GithubUpstreamBranch {get; set;} = "dev";
+        public string[] GitHubPullRequestNotifications { get; set; } = new string[] { };
+
+        public string[] UpdatedVersionsList {
```
pranavkm: Brance on nexe line

PR #915, opened 2018-02-22
Drop Pipelines.Testing dependency that we don't use

pakrym:


natemcmaster:
Needs to target dev instead. https://github.com/aspnet/KestrelHttpServer/pull/2343#issuecomment-367766167

pakrym:
Rebased

natemcmaster:
Assuming this needs https://github.com/aspnet/KestrelHttpServer/commit/d996f6b7fcc805b749a6dfaf3c3dd88dff691f85 too. CI automation is ~15 min away from updating the kestrel submodule. Once that's in, :shipit:

ahsonkhan:
When do we plan to get rid of the corefxlab dependencies and use the latest pipelines package?

https://github.com/aspnet/Universe/blob/dev/build/dependencies.props#L96-L99

ahsonkhan:
Same question regarding Kestrel - https://github.com/aspnet/KestrelHttpServer/blob/dev/build/dependencies.props#L34-L35

davidfowl:
@pakrym is working on it.

PR #918, opened 2018-02-24
Add dotnet-aspnet-codegenerator to artifacts.props

pranavkm:
Build error on the CI:

```
Undeclared package artifacts. Add these to artifacts.props: 
- dotnet-aspnet-codegenerator
```


pranavkm:
FYI @natemcmaster

PR #927, opened 2018-02-28
Add new WsFederation package

Tratcher:
See https://github.com/aspnet/Security/pull/1639


build/artifacts.props
```diff
@@ -46,6 +46,7 @@
     <PackageArtifact Include="Microsoft.AspNetCore.Authentication.OAuth" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
     <PackageArtifact Include="Microsoft.AspNetCore.Authentication.OpenIdConnect" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
     <PackageArtifact Include="Microsoft.AspNetCore.Authentication.Twitter" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
+    <PackageArtifact Include="Microsoft.AspNetCore.Authentication.WsFederation" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
```
Eilon: @DamianEdwards - just checking: this brings in some additional Microsoft-built dependencies, such as Microsoft.IdentityModel.Protocols.WsFederation and others. Is that OK for the .App package?

Tratcher: They're largely the same dependencies as OpenIdConnect.

natemcmaster:
Actually, this will most likely break Universe. You also need to update the submodule that contains this package.

build/artifacts.props
Eilon: Ah ok then I'm not concerned. Thanks!

PR #930, opened 2018-02-28
Add new package artifacts

pakrym:


pakrym:
Merging to unblock universe.

Issue #843, opened 2018-02-01
Remove workaround for SQLitePCLRaw 1.1.8


natemcmaster:
Remove this:

https://github.com/aspnet/Universe/blob/1c31731d89e757ce02478fa5af4f124a59a65985/build/SharedFx.targets#L122-L124

Blocked on https://github.com/aspnet/EntityFrameworkCore/issues/10724 and https://github.com/ericsink/SQLitePCL.raw/issues/194

Issue #833, opened 2018-01-31
Support Alpine Linux


natemcmaster:
All of our native dependencies need to also be updated to support musl instead of glibc.

 - https://github.com/aspnet/libuv-package/issues/23
 - https://github.com/ericsink/SQLitePCL.raw/issues/192

We also need to generate an aspnetcore-runtime tarball for Alpine.

- [x] Upgrade SQLitePclRaw
- [ ] ~~Upgrade libuv~~ Deferred until 2.2. See https://github.com/aspnet/libuv-package/issues/23
- [ ]  Add functional tests for Alpine
- [x]  Update CI to produce an Alpine configuration
- [x]  Update dotnet/CLI to consume

muratg:
SQLLitePCL.raw seems to be ready.  libuv-package task is assigned to @natemcmaster.

@glennc You had some concerns about this. Could you elaborate?  Tentatively putting this in the milestone so that we don't lose track of it.

natemcmaster:
I believe @bricelam is currently assigned to pull in a new SQLitePCL.raw package for a separate reason. We can use https://github.com/aspnet/EntityFrameworkCore/issues/10724 to track that update.

natemcmaster:
I've been poking at this for the last few days. The biggest challenge to implementing this at the moment is crossgen-ing the shared framework. This is difficult because

1. Installing dotnet via dotnet-install.sh into Alpine doesn't work yet. https://github.com/dotnet/cli/issues/8738 We can workaround for now by using curl/tar. 
1. MSBuild cannot compile a .csproj inside an Alpine container. https://github.com/Microsoft/msbuild/issues/3066
1. You can't run crossgen for Alpine on an Ubuntu machine because Ubuntu does not support the required libraries for musl-libc.

So, options:
1. Skip crossgen for now on Alpine Linux. (@DamianEdwards is this acceptable for a preview?)
2. Run crossgen without MSBuild. We could write MSBuild targets that generate shell scripts, but this must execute on an Ubuntu or Windows agent. Then, copy those scripts and files so crossgen can run inside a Docker container. (cc @JunTaoLuo )

DamianEdwards:
How are they doing it for MS.NETCore.App?

natemcmaster:
[...]

PR #931, opened 2018-02-28
Produce an Alpine Linux version of the ASP.NET Core shared framework

natemcmaster:
Part of #833. Support for Alpine is still going to be experimental. This will unblock being able to produce a shared runtime installer for alpine so we can start figuring out if there are other dependencies we need to upgrade. For sure we will need to update libuv, but it may be possible to rely on the libuv baked into Alpine Linux for now.

Changes in this PR:
 - Update to SQLitePCLRaw 1.1.10-pre20180223200113 (cc @Eilon for OSS approval)
 - Add alpine.3.6-x64 as a new RID
 - Add new external dependencies

Incidentally resolves https://github.com/aspnet/Universe/issues/843 (FYI @bricelam )
See also: https://github.com/aspnet/Microsoft.Data.Sqlite/pull/488


build/dependencies.props
```diff
@@ -179,9 +184,9 @@
     <RemotionLinqPackageVersion>2.2.0</RemotionLinqPackageVersion>
     <SerilogExtensionsLoggingPackageVersion>1.4.0</SerilogExtensionsLoggingPackageVersion>
     <SerilogSinksFilePackageVersion>3.2.0</SerilogSinksFilePackageVersion>
-    <SQLitePCLRawBundleGreenPackageVersion>1.1.8</SQLitePCLRawBundleGreenPackageVersion>
-    <SQLitePCLRawBundleSqlcipherPackageVersion>1.1.8</SQLitePCLRawBundleSqlcipherPackageVersion>
-    <SQLitePCLRawCorePackageVersion>1.1.8</SQLitePCLRawCorePackageVersion>
+    <SQLitePCLRawBundleGreenPackageVersion>1.1.10-pre20180223200113</SQLitePCLRawBundleGreenPackageVersion>
```
JunTaoLuo: Where is this version coming from? I don't see it on nuget.

natemcmaster: https://www.nuget.org/packages/SQLitePCLRaw.bundle_green/1.1.10-pre20180223200113 ?

JunTaoLuo: huh, not sure why i didn't see it before

JunTaoLuo: Still seems like a very strange version number. I'm assuming there will be a 1.1.10 stable at some point?

natemcmaster: Yes, Eric said he was planning to have 1.1.10 ready sometime in the next few weeks.

natemcmaster:
We can use https://github.com/aspnet/Universe/issues/933 to track follow-up to move to SQLitePCLRaw 1.1.10 when it becomes available.

PR #932, opened 2018-03-01
Add EFCore.Analyzers package

smitpatel:
@anpete @ajcvickers


build/artifacts.props
```diff
@@ -205,6 +205,7 @@
     <PackageArtifact Include="Microsoft.DotNet.Web.ProjectTemplates.2.1" Category="shipoob" />
     <PackageArtifact Include="Microsoft.DotNet.Web.Spa.ProjectTemplates" Category="shipoob" />
     <PackageArtifact Include="Microsoft.EntityFrameworkCore" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
+    <PackageArtifact Include="Microsoft.EntityFrameworkCore.Analyzers" Category="shipoob" />
```
pranavkm: `Analyzer="true"`

pranavkm: The plan for preview1 was to ship 0.0.1-alpha versions of analyzers. I haven't heard from @DamianEdwards what the plans are for preview2, but it doesn't seem like the EF analyzer package has the version set correctly

smitpatel: >  it doesn't seem like the EF analyzer package has the version set correctly

Elaborate please..

smitpatel: @pranavkm - Right now we just want this to be built and pushed to nightly feed. It should not be included in any sdk/meta package or nuget.

PR #940, opened 2018-03-05
Remove HostingStartup infrastructure from Universe

JunTaoLuo:
I don't think we are using any of this infrastructure for building hosting startup deps files anymore in 2.1 so this should be safe to remove.


JunTaoLuo:
Needed to rebase on the submodule updates in Universe.

PR #943, opened 2018-03-06
Add Microsoft.AspNet.WebApi.Client to .App and .All metapackages

JunTaoLuo:
Addresses https://github.com/aspnet/MetaPackages/issues/260.

I tested building the metapackage and building the shared framework locally on macOS.

This adds System.Net.Http.Formatting.dll to the .App shared framework.

cc @DamianEdwards @glennc @rynowak @davidfowl


natemcmaster:
Where does System.Net.Http.Formatting.dll come from?

natemcmaster:
Ah, nevermind. Dumb question. Its the binary that ships in  Microsoft.AspNet.WebApi.Client. It just doesn't match the package name.

PR #947, opened 2018-03-09
Add Oracle.ManagedDataAccess.Core package to external dependencies 

smitpatel:


build/dependencies.props
```diff
@@ -178,7 +178,7 @@
     <NETStandardLibrary20PackageVersion>2.0.1</NETStandardLibrary20PackageVersion>
     <NewtonsoftJsonBsonPackageVersion>1.0.1</NewtonsoftJsonBsonPackageVersion>
     <NewtonsoftJsonPackageVersion>10.0.1</NewtonsoftJsonPackageVersion>
-    <OracleManagedDataAccessPackageVersion>12.2.1100</OracleManagedDataAccessPackageVersion>
+    <OracleManagedDataAccessCorePackageVersion>2.12.0-beta2</OracleManagedDataAccessCorePackageVersion>
```
smitpatel: @Eilon - 3rd party dependency update. Used in sample provider only. Non-shipping package.

smitpatel:
Do we need OSS approval? It is not used in product code.
cc: @ajcvickers

ajcvickers:
@smitpatel Yep, we do.

build/dependencies.props
Eilon: Working on it

Eilon: Chatted with @smitpatel and @pakrym . This is a proprietary license so I'll have to review more closely when I'm back at the office on Monday.

Eilon:
Chatted with @ajcvickers and @smitpatel and we won't add the Oracle dependency to the build, but we'll try it as a one-off to see that everything works.

PR #949, opened 2018-03-09
UpdateRepos.ps1 get latest of branch

### LINKED ISSUE #938: UpdateRepos.ps1 should use dev HEAD instead of submodule commit.
Since UpdateRepos pushes to the repo branch instead of to Universe we should do `git checkout $($submodule.branch)`. This should make UpdateRepos pass more often, but still let Update Universe gate any bad changes.

ryanbrandenburg:
Fixes #938.


scripts/UpdateRepos.ps1
```diff
@@ -56,11 +56,12 @@ Push-Location $ModuleDirectory
 try {
 
     $build_errors = @()
-    # Get-Submodules also update --init's them
```
ryanbrandenburg: I didn't see any evidence that this is true so I removed it.

scripts/UpdateRepos.ps1
```diff
@@ -56,11 +56,12 @@ Push-Location $ModuleDirectory
...
     $submodules = Get-Submodules $RepoRoot
     foreach ($submodule in $submodules) {
         Push-Location $submodule.path
         try {
+            Invoke-Block { & git fetch }
+            Invoke-Block { & git checkout origin/$($submodule.branch) }
```
natemcmaster: This creates a branch named "origin/dev" locally, right? Might be better to do this:
```
git fetch origin dev:dev
git checkout dev
```

ryanbrandenburg: It does not (as long as your remote is named origin). It puts you in a detached head state at HEAD of origin/dev. I prefer this method because it doesn't pollute the sub-module with a dev branch, so it's "cleaner" for when we have to run this locally. I've previously gotten into weird states with branches on sub-modules of Universe which have been annoying.

natemcmaster: ok, makes sense

PR #956, opened 2018-03-13
Add publishing step to debian installers

JunTaoLuo:
Addresses https://github.com/aspnet/MetaPackages/issues/231.

The task is copied from https://github.com/dotnet/cli/pull/7460. The idea is to publish our debian installer (which is platform agnostic) to the three debian feeds for ubuntu 16.04, 14.04 and debian.8 on the nightly server, azure-apt-cat.cloudapp.net. Checked with @wli3 and made sure this is where the cli team publishes their debian installers and I received the credentials offline so I will update our build configs. 

Note that the cli team currently has paused their debian uploads and we may need to update the feeds we publish to later on (for example when ubuntu 17.10/18.04 need to be added).


JunTaoLuo:
I think a lot of this code can go into project arcade.

natemcmaster:
Won't have time to take a look today. I'll try to get to reviewing this tomorrow.

build/SharedFxInstaller.targets
```diff
@@ -254,13 +263,51 @@
...
+    </MSBuild>
+  </Target>
+
+  <Target Name="_PublishDeb">
+    <PropertyGroup>
+      <DebianRepoServer>azure-apt-cat.cloudapp.net</DebianRepoServer>
+    </PropertyGroup>
+    <ItemGroup>
+      <!-- ubuntu1604x64 -->
+      <DebianRepoId Include="575f40f3797ef7280505232f" />
```
natemcmaster: Just curious, where does this ID come from? How can we find the IDs for new distros, such as Ubuntu 18?

build/SharedFxInstaller.targets
```diff
@@ -254,13 +263,51 @@
...
   </Target>
 
   <Target Name="GenerateDebs" DependsOnTargets="_EnsureInstallerPrerequisites">
-    <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="_GenerateDebOnPlatform" Properties="PackageVersion=$(PackageVersion);Image=ubuntu.14.04" />
+    <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="_GenerateDebOnPlatform" Properties="PackageVersion=$(PackageVersion);Image=ubuntu.14.04" >
+      <Output TaskParameter="TargetOutputs" ItemName="DebsGenerated" />
+    </MSBuild>
+  </Target>
+
+  <Target Name="_PublishDeb">
```
natemcmaster: Can you move this to `Publish.targets` instead?

build/SharedFxInstaller.targets
```diff
@@ -254,13 +263,51 @@
...
+    <PropertyGroup>
+      <DebianRepoServer>azure-apt-cat.cloudapp.net</DebianRepoServer>
```
natemcmaster: Remove. We don't want to be hard-coding publish-paths anymore. Instead, add checks like this: 
https://github.com/aspnet/Universe/blob/734b3cefe8caea04b722e2886d2c64d1548e3b43/build/Publish.targets#L186-L188

Deployment settings should be in CI only.

build/tasks/LinuxPackageRepository/AddPackageStrategy.cs
```diff
@@ -0,0 +1,62 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
natemcmaster: I'm assuming this was copied from somewhere else? @Eilon  - license stuff

build/tasks/LinuxPackageRepository/LinuxPackageRepositoryHttpPrepare.cs
```diff
@@ -0,0 +1,46 @@
...
+        public async Task<string> RemoteCall()
+        {
+            using (var handler = new HttpClientHandler())
+            {
+                using (var client = new HttpClient(handler))
+                {
+                    var authHeader =
+                        Convert.ToBase64String(Encoding.UTF8.GetBytes((string) _linuxPackageRepositoryDestiny.GetSimpleAuth()));
+                    client.DefaultRequestHeaders.Authorization =
+                        new AuthenticationHeaderValue("Basic", authHeader);
```
natemcmaster: 😬 are we asserting that we only use HTTPS?

build/tasks/LinuxPackageRepository/AddPackageStrategy.cs
Eilon: From where was it copied?

JunTaoLuo: As per my original comment, this is copied from https://github.com/dotnet/cli/pull/7460. I didn't bother with updating the license headers since I assume this is only temporary and we will be moving the tool to dotnet/arcade soon?

Eilon: This is ok in this circumstance. Please consult me whenever copying code from any repo outside of our org (or product).

JunTaoLuo: Will do.

build/SharedFxInstaller.targets
JunTaoLuo: I got these from @wli3 in an email. I assume when a new repo is created for Ubuntu18 then we'll need to update our build configuration to include the new ids.

JunTaoLuo:
:up::date: Note that I will be merging to the preview2 branch before merging back to dev.

build/Publish.targets
```diff
@@ -106,6 +107,13 @@
         <IsDuplicateUpload>true</IsDuplicateUpload>
       </FilesToPublish>
 
+      <!-- Debian installers to be published to nightly feed -->
+      <FilesToPublish Include="$(DependencyAssetsDir)%(InstallerArtifacts.Identity)" Condition=" '%(InstallerArtifacts.InstallerType)' == 'deb' ">
```
natemcmaster: This adds duplicate entries into the FilesToPublish group for the debian installers. This will cause issues for PublishToTransportFeed and PublishToAzureFeed. 

Other ways to do this: 
 - add the metadata currently in `InstallerArtifacts` to `NativeInstaller` instead.
 - Create a new group `FilesToPublishToDebianFeed` and add this to the `_CheckFilesToPublish` target to ensure these files exist

build/Publish.targets
```diff
@@ -288,4 +296,31 @@
...
+      Server="$(PublishDebianInstallerServer)"
+      RepositoryId="%(RepoIds.Identity)"
+      Username="$(PublishDebianInstallerUser)"
+      Password="$(PublishDebianInstallerPassword)"
+      PathOfPackageToUpload="$(DebFile)"
+      PackageNameInLinuxPackageRepository="$(DebName)"
+      PackageVersionInLinuxPackageRepository="$(DebVersion)" />
+  </Target>
+
+  <Target Name="PublishDebs" DependsOnTargets="GetFilesToPublish">
```
natemcmaster: Add a condition here: `Condition="'$(PublishToDebianFeed)' == 'true'"` and use `$(PublishType.Contains('debian'))` to determine when this should happen. e.g. this would be skipped for prodcon builds.

https://github.com/aspnet/Universe/blob/5d670bf6b1b5fbe06942d00b0e27e1709396ee1e/build/repo.props#L47-L56

JunTaoLuo:
:up::date:

natemcmaster:
@JunTaoLuo - it looks like we may not want to do this anyways. Per that internal email thread, it looks like core-setup and cli are not publishing to the debian feed either, so it doesn't make sense for us to do it.

JunTaoLuo:
Confirmed not needed.

PR #957, opened 2018-03-13
Ship Microsoft.EntityFrameworkCore.Analyzers

anpete:
Update artifacts.props

### DIFF
--- a/build/artifacts.props
+++ b/build/artifacts.props
@@ -202,7 +202,7 @@
     <PackageArtifact Include="Microsoft.DotNet.Web.ProjectTemplates.2.1" Category="shipoob" />
     <PackageArtifact Include="Microsoft.DotNet.Web.Spa.ProjectTemplates" Category="shipoob" />
     <PackageArtifact Include="Microsoft.EntityFrameworkCore" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
-    <PackageArtifact Include="Microsoft.EntityFrameworkCore.Analyzers" Category="noship" />
+    <PackageArtifact Include="Microsoft.EntityFrameworkCore.Analyzers" Category="ship" Analyzer="true" AppMetapackage="true" AllMetapackage="true" />
     <PackageArtifact Include="Microsoft.EntityFrameworkCore.Abstractions" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
     <PackageArtifact Include="Microsoft.EntityFrameworkCore.Design" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
     <PackageArtifact Include="Microsoft.EntityFrameworkCore.DocumentDb" Category="noship"/>


build/artifacts.props
```diff
@@ -202,7 +202,7 @@
     <PackageArtifact Include="Microsoft.DotNet.Web.ProjectTemplates.2.1" Category="shipoob" />
     <PackageArtifact Include="Microsoft.DotNet.Web.Spa.ProjectTemplates" Category="shipoob" />
     <PackageArtifact Include="Microsoft.EntityFrameworkCore" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
-    <PackageArtifact Include="Microsoft.EntityFrameworkCore.Analyzers" Category="noship" />
+    <PackageArtifact Include="Microsoft.EntityFrameworkCore.Analyzers" Category="ship" Analyzer="true" AppMetapackage="true" AllMetapackage="true" />
```
natemcmaster: I was under the impression we didn't want to add analyzers to the Microsoft.AspNetCore.App/.All metapackages yet.
cc @DamianEdwards @divega  to confirm

Eilon: @natemcmaster that was true for the ASP.NET Core ones for sure, because I think they are only pre-rel, but for EF ones I'm uncertain.

pranavkm: Likely don't want it in the Analayzers metapackage (which is what `Analyzer="true"`) and the regular metapackage.

ajcvickers: Based on feedback from Roslyn team (I can re-forward the email if needed) analyzers should be shipped as development dependencies of the runtime package to which they apply. So what we're trying to do here is fit that guidance/best practice into the ASP.NET build system.

For EF, how to ship the analyzer was the only open question for shipping in 2.1, such that we decided to ship once we got the unambiguous answer on shipping best practice.

natemcmaster: So, if I understand what you're saying, you're also going to make Microsoft.EntityFrameworkCore depend on Microsoft.EntityFrameworkCore.Analyzers, right?

ajcvickers: Yes, it should be a "development dependency" which, based on my understanding, means that it will not get deployed with the app. I will forward you the email on this.

anpete: @ajcvickers @natemcmaster Just let me know if anything needs to change..

natemcmaster: Thanks @ajcvickers. Changes look fine here. 

I'm assuming there will be a similar change in the EFCore repo to add a dependency from EFCore -> Analyzers too. When you do that, make sure to douple check the final .nuspec generated for EFCore. By default, NuGet generates a nuspec with an "exclude" flag on it.
```
<dependency id="blah" version="blah" exclude="Build,Analyzers" />
```
You will probably want to change this default so analyzers are always included when EfCore is referenced directly. Use `PrivateAssets="None"` on your ProjectReference if that's how you will generate this nuspec.

PR #961, opened 2018-03-15
Remove Microsoft.AspNetCore.Sockets

davidfowl:
This changed deleted this package https://github.com/aspnet/SignalR/commit/6a8ede07701c82bbebb91d55a6d1dc5a42b129ee


davidfowl:
@JunTaoLuo why is the build failing?

JunTaoLuo:
You need to update the SignalR submodules as well, otherwise it will find the package and complain that you didn't include it in the artifacts file.

PR #965, opened 2018-03-15
Update dependencies to unblock ProdCon

ryanbrandenburg:
We're making this change to unblock ProdCon, once ProdCon passes we should take new version which align more correctly.


pakrym:
This would fail universe, versions are not aligned between core-setup and corefx.

Microsoft.NETCore.App 2.1.0-preview2-26308-01 coresponds to 2.1.0-preview2-26308-02 of corefx packages.

ryanbrandenburg:
:up::date:

pakrym:
Just don't merge submodule from Davids branch, there were breaking changes after he branched.

PR #969, opened 2018-03-16
Add new HostFiltering package

Tratcher:


pranavkm:
Can we hold off on this until after we branch? I want to snapshot the dev branch and we'll have to wait until a successful build if this goes in which will be several hours

ryanbrandenburg:
@pranavkm if we're going to make that a requirement of adding a new Package ID we should add it to the new package id creation wiki page in Coherence-Signed.

PR #973, opened 2018-03-19
Make dotnet-ef a global tool

bricelam:
Part of aspnet/EntityFrameworkCore#10554

### DIFF
--- a/build/artifacts.props
+++ b/build/artifacts.props
@@ -22,6 +22,7 @@
   <ItemGroup>
     <PackageArtifact Include="dotnet-aspnet-codegenerator" Category="shipoob" />
     <PackageArtifact Include="dotnet-dev-certs" Category="shipoob" />
+    <PackageArtifact Include="dotnet-ef" Category="shipoob" />
     <PackageArtifact Include="dotnet-sql-cache" Category="shipoob" />
     <PackageArtifact Include="dotnet-user-secrets" Category="shipoob" />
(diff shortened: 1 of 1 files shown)


natemcmaster:
Btw, to get this into the CLI, make a PR to update this file: https://github.com/dotnet/cli/blob/master/build/BundledDotnetTools.props

build/artifacts.props
```diff
@@ -215,7 +216,6 @@
     <PackageArtifact Include="Microsoft.EntityFrameworkCore.Sqlite.Core" Category="ship" AllMetapackage="true"/>
     <PackageArtifact Include="Microsoft.EntityFrameworkCore.SqlServer" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
     <PackageArtifact Include="Microsoft.EntityFrameworkCore.Tools" Category="ship" AppMetapackage="true" AllMetapackage="true" />
-    <PackageArtifact Include="Microsoft.EntityFrameworkCore.Tools.DotNet" Category="ship" LZMA="true" PackageType="DotnetCliTool" />
```
bricelam: @ryanbrandenburg Do I need to revert this?

ryanbrandenburg: Ah, I see, the problem wasn't that you removed the tool as a Package, it was that the reaction to that hadn't gotten through yet. You're other reaction's should be sufficient.

PR #977, opened 2018-03-20
Adjust docker build to use current user ID

natemcmaster:
Fixes the issue with undeleteable files.


natemcmaster:
The build worked just fine, but left behind files in .w/ and obj/ that were only deletable by root.

PR #985, opened 2018-03-22
Add Azure AD B2C integration package

javiercn:
/cc: @Eilon


javiercn:
FYI. Merging right away

natemcmaster:
@javiercn this broke the build. PR checks were red for valid reasons. If you can't fix this soon (~next 10 min or so), we'll need to revert to unblock ProdCon.

PR #986, opened 2018-03-22
Include symbols in published output

natemcmaster:
Latest ProdCon build was not publishing symbols for the sharedfx or packages.

Changes:
 - update build tools to ensure symbols packages from repos are included
 - add package info for sharedfx symbols nupkgs

Here is a complete list of all files expected to exist during publishing. 

https://gist.github.com/natemcmaster/5ae280dd41efe44efa7b603fb4198433


natemcmaster:
:up; :date: filters osx and add npm packages to prodcon blob publishing

PR #996, opened 2018-03-26
Fix UpdateDependencies script for maestro

ryanbrandenburg:


scripts/UpdateDependencies.ps1
```diff
@@ -81,13 +81,13 @@ if ($count -gt 0) {
     $ConfigVars += "--UpdatedVersions"
     $varString = ""
     foreach ($updatedVar in $updatedVars.GetEnumerator()) {
-        $varString += "$($updatedVar.Name)=$($updatedVar.Value);"
+        $varString += "$($updatedVar.Name)=$($updatedVar.Value)+"
```
ryanbrandenburg: ';' and ' ' don't work for the separator here, is there any danger in using '+'? If so what's a safer option?

natemcmaster: seems okay for now

PR #997, opened 2018-03-26
Generate branding.g.props for use in downstream builds that need to align version information

natemcmaster:
Reduce the number of places we have to manually update to keep our installers consistent with the runtime.

### DIFF
--- a/version.props
+++ b/version.props
@@ -1,12 +1,19 @@
 <Project>
   <PropertyGroup>
-    <VersionPrefix>2.0.7</VersionPrefix>
-    <VersionSuffix>rtm</VersionSuffix>
+    <AspNetCoreMajorVersion>2</AspNetCoreMajorVersion>
+    <AspNetCoreMinorVersion>0</AspNetCoreMinorVersion>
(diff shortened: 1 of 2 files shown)


version.props
```diff
@@ -1,13 +1,20 @@
 <Project>
   <PropertyGroup>
-    <VersionPrefix>2.0.7</VersionPrefix>
-    <VersionSuffix>rtm</VersionSuffix>
+    <AspNetCoreMajorVersion>2</AspNetCoreMajorVersion>
+    <AspNetCoreMinorVersion>0</AspNetCoreMinorVersion>
+    <AspNetCorePatchVersion>7</AspNetCorePatchVersion>
```
joeloff: Does this track the .NET Core version too? Reason is that for the hosting bundle, the branding is always based off the runtime version we redist and not any ASP.NET versions

natemcmaster: For 2.x, these should always align. We have to re-release an aspnetcore runtime installer for each .NET Core release, and we are planning to always align them

PR #999, opened 2018-03-27
[2.0.x] Add ANCM version to generate list of deps versions

natemcmaster:
Generate additional dependency information for downstream consumption by Windows installer builds.


natemcmaster:
Merging b/c we need to get the 2.0.x patch rolling asap

PR #1001, opened 2018-03-27
Add MicrosoftDiaSymReaderNativePackageVersion

pranavkm:


pranavkm:
Required (?) for https://github.com/aspnet/Mvc/pull/7552/files

natemcmaster:
You need to add an entry into build/external-dependencies.props as well if you want unification and update automation to work

pranavkm:
I don't think they produce this as part of ProdCon \ CoreFx \ CoreCLR. Don't see a version of it here - https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.DiaSymReader.Native.  @tmat, is Microsoft.DiaSymReader.Native produced outside of these repos?

tmat:
DSRN is built from internal repo. It's not open source.

pranavkm:
What does the file do?

natemcmaster:
It's used to generate the info that goes into the lineup package for automating updates across our repos.

pranavkm:
🆙 📅

build/external-dependencies.props
```diff
@@ -68,6 +68,7 @@
     <ExternalDependency Include="Microsoft.CSharp" Version="$(MicrosoftCSharpPackageVersion)" />
     <ExternalDependency Include="Microsoft.DotNet.PlatformAbstractions" Version="$(MicrosoftDotNetPlatformAbstractionsPackageVersion)" />
     <ExternalDependency Include="Microsoft.DotNet.ProjectModel" Version="$(MicrosoftDotNetProjectModelPackageVersion)" Private="true" />
+    <ExternalDependency Include="Microsoft.DiaSymReader.Native" Version="$(MicrosoftDiaSymReaderNativePackageVersion)" Private="true" />
```
natemcmaster: Don't use the Private=true flag. This ends up as an external dependency of MVC, so it will fail validation. You should remove Private=true.

>  This dependency is 'Private', aka. it should not end up as a public-facing external dependency. This is validated by checking the nuspec on generated packages.

pranavkm: Done

PR #1007, opened 2018-03-27
Merge release/2.1 -> dev

HaoK:


HaoK:
What about the submodule update for Config, how does that get merged over?

natemcmaster:
Good point. did you merge to dev in aspnet/Config? If so, `git submodule update --remote modules/Configuration && git add modules/Configuration`

HaoK:
Cool this looks like the matching set of change snow...  would a cherrypick from release have worked better here?

natemcmaster:
No, merging is the right thing. It just takes some manual effort since there isn't a good submodule merge strategy

PR #1008, opened 2018-03-27
Add .App metapackage to the LZMA

JunTaoLuo:
Let's take a closer look at https://github.com/aspnet/Universe/commit/d8f190561432ef9dce5b2fef16349d8ff1a321fd.

Currently Microsoft.AspNetCore.App is missing from the LZMA in preview2 but was there in preview1. There were a few reverts and merges around these changes and we ended up with the wrong set of changes in release/2.1. This is currently breaking VS new template tests run by the vendors.


natemcmaster:
I manually scanned through https://github.com/aspnet/Universe/commit/d8f190561432ef9dce5b2fef16349d8ff1a321fd and didn't see anything obvious.

PR #1011, opened 2018-03-28
Add option to update dependencies from core-setup build

ryanbrandenburg:
This script allows us to update the dependencies in Universe based off a core-setup pass, rather than having to wait for a final build.xml.


scripts/UpdateDependencies.ps1
```diff
@@ -17,36 +18,125 @@ Set-StrictMode -Version 1
...
-    Write-Verbose "Downloading from $url to $BuildXml"
-    Invoke-WebRequest -OutFile $localXml $url
-}
+$variables = @{}
 
-[xml] $remoteDeps = LoadXml $localXml
-$count = 0
+if ($UseCoreFx) {
+    if ($BuildXml) {
+        Write-Warning "$BuildXml parameter will not be used when the UseCoreFx switch is passed."
```
natemcmaster: Seems like a code smell. Might be cleaner to have separate script files, and extract the shared code into functions that you import via a .psm1 file

ryanbrandenburg:
:up::date: I changed us to start using hub instead of the PR creating infrastructure dotnet uses, I think it's easier to follow and it's less to maintain.

pakrym:
🎉

PR #1015, opened 2018-03-30
Add Microsoft.AspNetCore.Analyzers.Experimental package

pranavkm:


build/artifacts.props
```diff
@@ -110,7 +110,8 @@
     <PackageArtifact Include="Microsoft.AspNetCore.MiddlewareAnalysis" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
     <PackageArtifact Include="Microsoft.AspNetCore.Mvc" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
     <PackageArtifact Include="Microsoft.AspNetCore.Mvc.Abstractions" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
-    <PackageArtifact Include="Microsoft.AspNetCore.Mvc.Analyzers" Category="shipoob" Analyzer="true" />
+    <PackageArtifact Include="Microsoft.AspNetCore.Mvc.Analyzers" Category="ship" Analyzer="true" AppMetapackage="true" AllMetapackage="true" />
+    <PackageArtifact Include="Microsoft.AspNetCore.Mvc.Analyzers.Experimental" Category="shipoob" Analyzer="true" />
```
natemcmaster: So, just to be clear, both `Mvc.Analyzers` and `.Experimental` are meant to be dependencies of the `Microsoft.AspNetCore.Analyzers` metapackage?

pranavkm: I guess so. I think it makes sense for the meta-analyzer package to reference all analyzers we produce. Additionally, the EF analyzer works the same way

PR #1018, opened 2018-03-30
noship Microsoft.AspNetCore.Server.IIS

jkotalik:
@JunTaoLuo will this remove Microsoft.AspNetCore.Server.IIS from the meta packages?


build/artifacts.props
```diff
@@ -147,7 +147,7 @@
     <PackageArtifact Include="Microsoft.AspNetCore.Routing.Abstractions" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
     <PackageArtifact Include="Microsoft.AspNetCore.Routing.DecisionTree.Sources" Category="noship" />
     <PackageArtifact Include="Microsoft.AspNetCore.Server.HttpSys" Category="ship" AppMetapackage="true" AllMetapackage="true"/>
-    <PackageArtifact Include="Microsoft.AspNetCore.Server.IIS" Category="ship" WinAppMetapackage="true" WinAllMetapackage="true" />
+    <PackageArtifact Include="Microsoft.AspNetCore.Server.IIS" Category="noship" WinAppMetapackage="true" WinAllMetapackage="true" />
```
natemcmaster: We'll need to cut out more than just this to unship it. You'll need to revert https://github.com/aspnet/Universe/pull/989 too.

jkotalik: Hey, let's think about the positives. There will be no native dependencies in the meta packages now!

JunTaoLuo:
What's going on with this? Are we pushing it out to rc or beyond 2.1? If we are just moving it to rc, I'd just mark https://github.com/aspnet/Universe/blob/dev/build/artifacts.props#L312-L313 as noship. If we are committed to not shipping it for 2.1 then revert the change Nate mentioned.

Eilon:
@JunTaoLuo out of 2.1 was my understanding.

jkotalik:
@natemcmaster I added the revert in this PR too.

PR #1021, opened 2018-03-30
Merge into dev.

jkotalik:


jkotalik:
This is still a non-mergeable request. I'll investigate why.

PR #1025, opened 2018-04-02
update artifacts.props in dev

anurse:
No clue if this will work... but it's what people told me to do ;)


anurse:
The failures appear to be because Update Universe hasn't run and @pakrym said that it should pick up the changes with this merge, so I'm going to merge.

natemcmaster:
Hmm, this actually broke the build. 

http://aspnetci/viewLog.html?buildId=438404&tab=buildResultsDiv&buildTypeId=UniverseCoherence. 

We cannot compile https://github.com/aspnet/Universe/commit/bbc9f56eceba2810071e9b1b4da7b99a4e529c58 because the submodule of SignalR it points to does not contain these renamed packages.

Automation hasn't fixed universe yet because there is problem updating aspnet/Razor and aspnet/Scaffolding, so you'll probably have to resolve this manually.

anurse:
Done

Issue #1014, opened 2018-03-29
MSBuild on macOS is running out of memory


natemcmaster:
Something in the way we are using MSBuild, some recent change to our projects or targets, or some recent change to .NET Core 2.1 is causing MSBuild to die with out-of-memory failures. It's consistently reproducing on our CI machines.

natemcmaster:
Captured sample before the crash:
[dotnet-sample.txt](https://github.com/aspnet/Universe/files/1862165/dotnet-sample.txt)

natemcmaster:
[dotnet-sample2.txt](https://github.com/aspnet/Universe/files/1862217/dotnet-sample2.txt)

natemcmaster:
Status update: I have not been able to figure out what is going wrong yet. Could be something in MSBuild's nodeReuse refactoring, a bug in the JIT, or something else. I have a memory dump but know way to inspect what's in it. I'm currently considering other options. If we did #922, we wouldn't need the expensive analysis in AnalyzeBuildGraph task (which is currently the source of this issue). Unless I can get traction on debugging MSBuild for macOS, I may pursue that instead

@ryanbrandenburg @Eilon

natemcmaster:
Turns out it's a lot of work to setup a machine to debug a macOS dump. https://github.com/dotnet/coreclr/issues/12301 you have to recompile libcoreclr yourself. The memory dump I have was gathered using a different libcoreclr, so the dump I have is useless unless we run MSBuild on a custom private of coreclr.

Let's just workaround this by hard-coding for now (#1028) and readdress this when we get to #922.

Eilon:
No plans to continue the investigation here. Will evaluate https://github.com/aspnet/Universe/issues/922 instead.

PR #1028, opened 2018-04-02
[Workaround] Hard-code the repo-to-repo build order on macOS and Linux

natemcmaster:
Temporary workaround for #1014. The repo to repo graph rarely changes, so let's just hard-code this for now while we try to figure out what is happening in MSBuild that is causing OutOfMemoryException on macOS/Linux.


build/repo.targets
```diff
@@ -154,14 +157,19 @@
...
+    <RepoTasks.AnalyzeBuildGraph Condition=" ! $([MSBuild]::IsOSUnixLike())"
       Solutions="@(Solution)"
       Artifacts="@(ArtifactInfo)"
       Dependencies="@(ExternalDependency)"
       StartGraphAt="$(BuildGraphOf)"
       Properties="Configuration=$(Configuration);BuildNumber=$(BuildNumber);DotNetPackageVersionPropsPath=$(GeneratedPackageVersionPropsPath);DotNetRestoreSourcePropsPath=$(GeneratedRestoreSourcesPropsPath)">
       <Output TaskParameter="RepositoryBuildOrder" ItemName="RepositoryBuildOrder" />
     </RepoTasks.AnalyzeBuildGraph>
+
+    <Message Condition="$([MSBuild]::IsOSUnixLike())"
```
ryanbrandenburg: Nitpick, let's add a comment noting that this is part of #1014 too here so that no parts get missed in the revert.

natemcmaster: updated

build/repo.targets
```diff
@@ -167,9 +167,8 @@
       <Output TaskParameter="RepositoryBuildOrder" ItemName="RepositoryBuildOrder" />
     </RepoTasks.AnalyzeBuildGraph>
 
-    <Message Condition="$([MSBuild]::IsOSUnixLike())"
-             Text="%(RepositoryBuildOrder.Order). @(RepositoryBuildOrder, ', ')"
-             Importance="high" />
+    <Message Text="Repository build order:" Importance="high" />
```
ryanbrandenburg: Even better solution than my suggestion!

PR #1029, opened 2018-04-02
Reaction to *Memory changes (preview3-26331-02 corefx)

pakrym:


build/dependencies.props
```diff
@@ -1,90 +1,90 @@
 ﻿<Project>
   <!-- These package version may be overridden or updated by automation. -->
   <PropertyGroup Label="Package Versions: Auto" Condition=" '$(DotNetPackageVersionPropsPath)' == '' ">
-    <MicrosoftCSharpPackageVersion>4.5.0-preview2-26330-05</MicrosoftCSharpPackageVersion>
-    <MicrosoftDotNetPlatformAbstractionsPackageVersion>2.1.0-preview2-26330-03</MicrosoftDotNetPlatformAbstractionsPackageVersion>
-    <MicrosoftExtensionsDependencyModelPackageVersion>2.1.0-preview2-26330-03</MicrosoftExtensionsDependencyModelPackageVersion>
+    <MicrosoftCSharpPackageVersion>4.5.0-preview3-26331-02</MicrosoftCSharpPackageVersion>
```
natemcmaster: Should we wait until we have a preview2 build of these changes? Seems like we shouldn't mix preview3 into the preview2 stream.

pakrym: It's to painfull to change all reaction PR's to preview2 at this point, we'll downgrade as soon as we get a package. Change wise preview2 and 3 are the same.

Issue #856, opened 2018-02-07
Remove the need to mirror packages + transition to dotnet.myget.org/dotnet-core


pranavkm:
Our repositories currently reference aspnetcore-[dev|release] feeds and NuGet.org while relying on Universe to mirror packages for us from `dotnet-core`, `dotnet-corefxlab`, `roslyn`, `vssdk` and `vsformac` feeds. 
CoreFxLab will transition to dotnet-core and `dotnet-corefxlab` will no longer need to be referenced. The remainder are meant to support tooling and are not consumed by runtime packages. 

Here's my suggestion, once `dotnet-corefxlab` goes away,

* [x] Update repos to reference `dotnet-core` as a source
* [ ] Have Razor explicitly reference `roslyn`, `vssdk`, and `vsformac` (Are there other repos that reference packages from an external feed?)
* [x] Tell customers to use `dotnet-core` to get their ASP.NET Core nightlies. We should be ready to make an announcement to this effect and stop using aspnetcore-dev. - https://github.com/aspnet/Announcements/issues/293

pranavkm:
cc @natemcmaster \ @pakrym \ @Eilon

Eilon:
This does look overall correct to me.

natemcmaster:
+1. I'm in favor of this.

pranavkm:
@pakrym could I assign this to you so we can track whenever you're done with the corefxlab changes?

pakrym:
Sure,

natemcmaster:
How would we handle situations like this? #869  - we need a new version of xunit from yet-another myget feed. Should we update the feeds in each repo? Or continue to mirror?

pranavkm:
Not entirely sure. A test-only package is easy to deal with since the repo could reference the package and we'd be done. Packages consumed by runtime would be far more problematic (for instance if Kestrel needs a nightly build of Libuv?). Maybe we'll still need to keep the feature around, just not use it.

pakrym:
Unblocked now.

Eilon:
We decided that for now we should stop mirroring dependencies and we'll see if anything breaks.

PR #1043, opened 2018-04-07
Remove unused external dependencies

natemcmaster:
Follow-up to #856  - many of these dependencies were transitive dependencies of our explicit dependencies, so they were listed because we needed to mirror them. We don't need to mirror anymore, so we can remove these items from our list of external dependencies.


build/tasks/AnalyzeBuildGraph.cs
```diff
@@ -174,6 +181,14 @@ private void EnsureConsistentGraph(IEnumerable<ArtifactInfo.Package> packages, I
                 Log.LogError("Package versions are inconsistent. See build log for details.");
             }
 
+            foreach (var versions in dependencyMap.Values)
+            {
+                foreach (var item in versions.Where(i => !i.IsReferenced))
+                {
+                    Log.LogMessage(MessageImportance.Normal, $"Potentially unused external dependency: {item.PackageId}/{item.Version}");
```
Eilon: Why "potentially"? Also, what should someone do if they see this message?

build/tasks/AnalyzeBuildGraph.cs
```diff
@@ -150,7 +157,7 @@ private void EnsureConsistentGraph(IEnumerable<ArtifactInfo.Package> packages, I
                     ActualVersion = dependency.Value.Version,
                     ExpectedVersion = package.PackageInfo.Version,
                 });
-            }
+                }
```
Eilon: ` ` ` ` ` ` ` `

build/tasks/AnalyzeBuildGraph.cs
```diff
@@ -115,15 +119,18 @@ private void EnsureConsistentGraph(IEnumerable<ArtifactInfo.Package> packages, I
             {
                 if (!buildPackageMap.TryGetValue(dependency.Key, out var package))
                 {
+                    int idx = -1;
```
Eilon: `var`

build/dependencies.props
```diff
@@ -2,85 +2,30 @@
   <!-- These package version may be overridden or updated by automation. -->
   <PropertyGroup Label="Package Versions: Auto" Condition=" '$(DotNetPackageVersionPropsPath)' == '' ">
     <MicrosoftCSharpPackageVersion>4.5.0-preview2-26403-05</MicrosoftCSharpPackageVersion>
-    <MicrosoftDotNetPlatformAbstractionsPackageVersion>2.1.0-preview2-26403-06</MicrosoftDotNetPlatformAbstractionsPackageVersion>
```
Eilon: Nice to get rid of all of these! Do we have a "safe" way of knowing whether this will pass? Can we make sure a local build of Universe is done to mitigate potentially breaking the CI?

build/tasks/AnalyzeBuildGraph.cs
natemcmaster: I put this message as "normal" level so it won't normally be seen unless someone is intentionally working on pruning our dependency list.

It says potentially because it's possible, though rare, that a package may be used but not as a PackageReference. This analysis only detects PackageReference. There are a few ways this can happen:
 - implicit package references (like NETStandard.Library and Microsoft.NETCore.App)
 - packages that are nuspec only dependencies, like MVC's dependency on Microsoft.DiaSymReader.Native
 - packages that were just barely added but are not yet referenced in a submodule, like Polly.Extensions.Http-Signed

build/dependencies.props
natemcmaster: We could enable PR checks that build Universe, but otherwise, not really. I feel pretty okay about this though because I did a search over all source code to make sure packages I've removed are not referenced. If this fails, it's easy to revert.

build/tasks/AnalyzeBuildGraph.cs
Eilon: Can we create a wiki page that has this exact info on it, and link to it from the message? Otherwise I'm afraid that whoever sees that message is unlikely to know exactly what it means and how to act on it.

build/dependencies.props
ryanbrandenburg: Yeah, this could really save us from thoughtlessly taking dependencies in new places just because  we think it's already in use.

Eilon: Yeah that's fine.

build/tasks/AnalyzeBuildGraph.cs
natemcmaster: https://github.com/aspnet/Universe/wiki/Build-warning-and-error-codes#potentially-unused-external-dependency

PR #1062, opened 2018-04-12
Update CoreFx script

ryanbrandenburg:
We need to make some changes to this script to make it work on the CI. Namely, we must allow the use of a GitHub access token.


scripts/common.psm1
```diff
@@ -143,9 +143,10 @@ function CommitUpdatedVersions([hashtable]$updatedVars, [xml]$dependencies, [str
         $remote = "origin"
         $baseBranch = "dev"
 
+        Invoke-Block { & git stash }
```
natemcmaster: What's this about?

ryanbrandenburg: We've already made changes to build/dependencies.props, if we don't stash our changes before we checkout checkout will fail.

pakrym:
https://github.com/aspnet/Universe/issues/1063

scripts/common.psm1
natemcmaster: Is there any reason we cannot but the branch checkout to the top of the script? stash pop is prone to merge conflicts.

scripts/UpdateDependencies.ps1
```diff
@@ -15,6 +15,12 @@ Import-Module -Scope Local -Force "$PSScriptRoot/common.psm1"
 Set-StrictMode -Version 1
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
 
+$GitHubEmail = $ConfigVars["GithubEmail"]
+$GitHubUsername = $ConfigVars["GithubUsername"]
+$GitHubPassword = $ConfigVars["GithubToken"]
+
+Set-GitHubInfo $GitHubPassword $GitHubUsername $GitHubEmail
```
natemcmaster: So, if I want to run this locally without creating a PR, how would I skip this? Might be useful to add a `-NoPush` and/or `-NoCommit` switch to skip this

ryanbrandenburg:
:up::date:. @pakrym this should fix #1063.

PR #1065, opened 2018-04-13
Minor tweaks to update scripts

natemcmaster:
Resolves https://github.com/aspnet/BuildTools/issues/637 and an error in generating tags


natemcmaster:
TC updated.

PR #1072, opened 2018-04-14
Update artifacts.props

rynowak:
We need to set this as shipping, but we don't want it to be in either metapackage


rynowak:
@javiercn signed off in person 😆

PR #1080, opened 2018-04-18
Resolve conflicts in additionalDeps when building RS by selecting the latest deps.json files

natemcmaster:
When we produce a new hosting startup additionalDeps file without a new version of Microsoft.NETCore.App, build currently fails due to a conflict in file path.

> The file '/mnt/work/71ae3284d35ddee4/.w/additionalDeps/Microsoft.AspNetCore.ApplicationInsights.HostingStartup/shared/Microsoft.NETCore.App/2.0.7/Microsoft.AspNetCore.ApplicationInsights.HostingStartup.deps.json' already exists. 

This resolves the conflict by overwriting conflicts with the files from the current build. This means some releases may require an in-place upgrade for some .deps.json files.

Possible issues we need to investigate:

 - MSI. Since .deps.json files are not versioned, will MSI update in-place existing .deps.json files?
 - File locking. If a server is already running, will extracting a new RS in place and overwriting .deps.json files in additionalDeps fail because of file locking?

cc @JunTaoLuo @pakrym @muratg


JunTaoLuo:
I think the idea is not to overwrite but to rather to not include deps files in patches where there is no new core runtime.

natemcmaster:
Discussed in person. Overwriting and in-place updates are too complicated for installers, so we're not going to include additionalDeps updates in this release.

PR #1086, opened 2018-04-19
Add Microsoft.AspNetCore.AzureAppServices.SiteExtension to normal builds

natemcmaster:
React to https://github.com/aspnet/AzureIntegration/pull/188

FYI @muratg  - this is an infrastructure-only change to get this package into official ProdCon output


natemcmaster:
Merging to unblock CI.

natemcmaster:
Reverted. This breaks on CI because you can't have a PackageReference internally, but we can't produce a site extension using ProjectReference yet.

PR #1087, opened 2018-04-19
Rename EFCore.DocumentDb package to EFCore.Cosmos.Sql

smitpatel:
Reaction PR for https://github.com/aspnet/EntityFrameworkCore/pull/11742

I will update submodule once PR on EF merged.


smitpatel:
@ryanbrandenburg - Nope. The package does not have any source as of now so there cannot be any dependency.

PR #1097, opened 2018-04-20
Mark KeyVault.HostingStartup as shipoob

natemcmaster:
This package is still in preview, so don't ship it from our regular patch system yet.


natemcmaster:
Merging per in-person discussion.

PR #1099, opened 2018-04-23
Update to NETStandard.Library 2.0.3

natemcmaster:
This version does not flow automagically via ProdCon, so we need to manually update it.

cc @Petermarcu


natemcmaster:
Talked to @muratg  in person. Merging this to ensure we align versions for the RC1 build.

PR #1100, opened 2018-04-23
UpdateDeps to match maestro

ryanbrandenburg:


ryanbrandenburg:
I also realized that we were targeting the wrong branch for our PR. We can only do this for master/dev so, release/2.1 was wrong.

natemcmaster:
Let's retarget this PR to dev

scripts/UpdateDependenciesCoreFx.ps1
```diff
@@ -103,7 +103,7 @@ Write-Host "Loading deps from $depsPath"
 [xml] $dependencies = LoadXml $depsPath
 
 $remote = "origin"
-$baseBranch = "release/2.1"
+$baseBranch = "dev"
```
natemcmaster: Any chance we can detect this by running a git command to find current branch?

ryanbrandenburg: I'm not a fan of that behavior because I usually have a branch like `rybrande/*` checked out when I test stuff locally, which shouldn't be the base.  Also, any changeability of this value should accompany a change to `$coreSetupVersions` (dev->master, release/2.1->release/2.1). I specifically didn't want to have a branch name variable because it would mislead people into thinking the script currently supported release/2.1. I should probably add a comment to that effect at the top.

ryanbrandenburg:
Closing this to re-target dev.

scripts/UpdateDependenciesCoreFx.ps1
natemcmaster: Fair enough.

PR #1104, opened 2018-04-23
Add linux-musl-x64 shared runtime

natemcmaster:
React to https://github.com/dotnet/corefx/pull/28560 and https://github.com/dotnet/coreclr/pull/17623 - the base sharedfx is moving to 'linux-musl'.

Note: we will still keep `alpine.3.6-x64` for now. It's not clear we need to drop it in order to ship 2.1, and there is less risk if we simply add linux-musl side by side for now.


Petermarcu:
Change looks good to me but I don't know your infra as well as others.

natemcmaster:
Talked to @muratg. Merging as we are considering this a continuation of the linux-musl work done by the core-setup team.

PR #1105, opened 2018-04-23
UpdateDeps to match maestro

ryanbrandenburg:


natemcmaster:
Merged so we can unblock the failing bot runs.

PR #1128, opened 2018-04-30
Set version to 2.1.0-rtm

natemcmaster:
Bump the framework version to 2.1.0-rtm

TODO before merging:
 - [x] bump the version in all repos to 2.1.0-rtm as well

### DIFF
--- a/version.props
+++ b/version.props
@@ -1,10 +1,10 @@
 <Project>
   <PropertyGroup>
     <VersionPrefix>2.1.0</VersionPrefix>
-    <VersionSuffix>rc1</VersionSuffix>
+    <VersionSuffix>rtm</VersionSuffix>
 
     <ExperimentalVersionPrefix>0.1.0</ExperimentalVersionPrefix>
-    <ExperimentalVersionSuffix>alpha2</ExperimentalVersionSuffix>
+    <ExperimentalVersionSuffix>rtm</ExperimentalVersionSuffix>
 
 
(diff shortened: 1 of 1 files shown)


version.props
```diff
@@ -1,10 +1,10 @@
 <Project>
   <PropertyGroup>
     <VersionPrefix>2.1.0</VersionPrefix>
-    <VersionSuffix>rc1</VersionSuffix>
+    <VersionSuffix>rtm</VersionSuffix>
 
     <ExperimentalVersionPrefix>0.1.0</ExperimentalVersionPrefix>
-    <ExperimentalVersionSuffix>alpha2</ExperimentalVersionSuffix>
+    <ExperimentalVersionSuffix>rtm</ExperimentalVersionSuffix>
```
pranavkm: This might be incorrect or might need fixup in dev: https://github.com/aspnet/Mvc/blob/dev/version.props#L7

pranavkm: I guess? Might want to run it by a PM for confirmation since it's a problem that affects other repos too - https://github.com/aspnet/Logging/blob/dev/version.props#L6, https://github.com/aspnet/Hosting/blob/dev/version.props#L7 (release/2.1 is 0.5.0-preview2 in this case)

natemcmaster: It probably doesn't matter much. These are just placeholders until we get ready to ship experiments. I took a look at our build output, and we only have one 0.\* version packaged in the "ship" category, but Murat and I are going to move that to noship before rtm.

That said, which PM should I ping for the Mvc change?

Issue #1125, opened 2018-04-27
Provide way to disable trimming against shared framework for source-build installations


natemcmaster:
To address https://github.com/dotnet/source-build/issues/456, we want to make it easier to opt-out of the aspnet shared framework.

### Changes to make in our code

- [x] Make the override for MicrosoftNETPlatformLibrary in our .App and .All packages conditional on a property.
- [x] Set this property in source-build installations of the SDK

### Implications
* Default templates will build and run on source-build installations without requiring changes to the file or special properties passed as command line flags. 
* Apps will run from the NuGet cache, not the optimized shared framework. This means slower startup and larger application size.
* Users who publish an app with a non-source-build SDK (e.g. publish on Windows) and then try to run on a source-build-only server will be missing the shared framework. This situation is similar to several other scenarios where servers may not be configured with the aspnet shared framework, so we believe this will be address with documentation on how to configure a server with the aspnet shared runtime.

cc @omajid @tmds @muratg

natemcmaster:
Done. Changes are in to both dotnet/CLI and aspnetcore.

https://github.com/aspnet/Universe/pull/1130
https://github.com/dotnet/cli/pull/9178

PR #1130, opened 2018-04-30
Add property to disable shared framework trimming

natemcmaster:
Resolves https://github.com/aspnet/Universe/issues/1125


natemcmaster:
FYI - I tested this manually against a build of the CLI with https://github.com/dotnet/cli/pull/9178, and verified the behavior.

Microsoft.AspNetCore.App/.All are trimmed by default. An SDK build with IncludeAspNetCoreRuntime=false will use Microsoft.NETCore.App instead.

Also, referencing both Microsoft.AspNetCore.App/.All works, and should be deterministic now, regardless of the order in which the package props/targets are imported.

natemcmaster:
This was approved in shiproom. @JunTaoLuo can you take a look?

PR #1135, opened 2018-05-02
[Design]Include sdk in lineup

ryanbrandenburg:


build/dependencies.props
```diff
@@ -31,6 +31,7 @@
     <SystemThreadingTasksDataflowPackageVersion>4.9.0-preview3-26423-04</SystemThreadingTasksDataflowPackageVersion>
     <SystemThreadingTasksExtensionsPackageVersion>4.5.0-preview3-26423-04</SystemThreadingTasksExtensionsPackageVersion>
     <SystemValueTuplePackageVersion>4.5.0-preview3-26423-04</SystemValueTuplePackageVersion>
+    <SdkPackageVersion>2.1.300-preview2-008530</SdkPackageVersion>
```
natemcmaster: I don't believe we can put this here since ProdCon doesn't flow an SDK version to us. We build before the CLI, so this wouldn't be available anyways.

scripts/UpdateDependencies.ps1
```diff
@@ -55,6 +55,8 @@ foreach ($package in $remoteDeps.SelectNodes('//Package')) {
     }
 }
 
+$globalObj = Get-Content "$PSScriptRoot/../global.json" -Raw | ConvertFrom-Json
+$variables["SdkPackageVersion"] = @($globalObj.sdk.version)
```
natemcmaster: This isn't actually updating to a new version. To make this work, we'll need to figure out which ProdCon version variable to use, and then update the global.json file.

scripts/UpdateDependencies.ps1
```diff
@@ -55,14 +55,26 @@ foreach ($package in $remoteDeps.SelectNodes('//Package')) {
     }
 }
 
+$tools = @{}
+
+$cliNode = $remoteDeps.SelectSingleNode("//Build[@Name='cli']")
+$cliVersion = $cliNode.BuildId
```
natemcmaster: Use ProductVersion. Not sure if BuildId will always align.

build/dependencies.props
ryanbrandenburg: Fair duce.

build/tools.props
```diff
@@ -0,0 +1,5 @@
+<Project>
```
ryanbrandenburg: We shouldn't check this file in, it should just be generated based off the global.json in the dependencies.props format.

ryanbrandenburg:
This has been de-prioritized, so I'm closing it for now.

PR #1150, opened 2018-05-07
fix build break. These packages got renamed and it was out of sync

rynowak:
/cc @natemcmaster


natemcmaster:
Update this too https://github.com/aspnet/Universe/blob/b520569f23f67d238a101107b39ba380827865a9/build/external-dependencies.props#L113-L114

rynowak:
updated - will merge when travis and appveyor are green

PR #1161, opened 2018-05-09
Add Internal.AspNetCore.Analyzers

pakrym:


pranavkm:
Need to update submodule too

Issue #962, opened 2018-03-15
Put the universe commit sha in the Microsoft.AspNetCore.App metapackage


davidfowl:
I'd like to backtrack from a meta package version to the universe commit that produced it. It seems like that information only exists on the build server today and I'd like to stamp it into the package. 

/cc @natemcmaster

PR #1168, opened 2018-05-18
Append commit hash to metapackages

JunTaoLuo:
Addresses https://github.com/aspnet/Universe/issues/962. Not importing the internal SDK since we don't want the reference to show up as the dependency of the metapackage.


natemcmaster:
>  Not importing the internal SDK since we don't want the reference to show up as the dependency of the metapackage.

Can we set PrivateAssets = All?

Also, let's get signoff from managers first. I see you self-triaged #962. We haven't agreed to do this yet because NuGet is working on a built-in mechanism for this, so this may be unnecessary.

JunTaoLuo:
Where is the work for the built-in mechanism being tracked?

natemcmaster:
https://github.com/NuGet/NuGetGallery/issues/4941

natemcmaster:
We already embed repository commit into the package metadata (by setting the [RepositoryCommit ](https://github.com/aspnet/BuildTools/pull/586) property), but the NuGet gallery and UI doesn't show it yet.

JunTaoLuo:
Ah yes, @ajaybhargavb and I were wondering why that wasn't being displayed yesterday. Didn't know there's a plan on making that available. @davidfowl would it be sufficient if we push nuget to expose RepositoryCommit instead of us injecting it into the description or would we want both?

natemcmaster:
I wonder if we could make a PR to NuGet to add this. I poked through their code, and it seems like it shouldn't be hard to improve the nuget gallery to show this. And it would light up for our 2.1 packages, which already have the commit metadata.

davidfowl:
Do both? If @DamianEdwards tells me that he's super had that its in the description then maybe. But I can view it on the website without having to download or unzip anything and thats really what I want.

Do we also have it in assembly metadata?

DamianEdwards:
I don't mind either way

davidfowl:
@JunTaoLuo add the repository url metadata and do this.

JunTaoLuo:
Moving the implementation to https://github.com/aspnet/BuildTools/pull/678. Initially thought we don't use the internal aspnetcore sdk for building metapackage but we do. There's no point in copying the implementation in multiple locations.

PR #1169, opened 2018-05-18
Update artifact

ryanbrandenburg:


ryanbrandenburg:
Merging directly to unblock builds.

PR #1174, opened 2018-05-23
Antares fallback archive zips

JunTaoLuo:
This adds the target to build the incremental fallback zips for antares. The target can be turned on to build an incremental zip every build and can also be configured via InputLzmaUrl/File and PreviousLzmaUrl/File to produce one off antares zips for any release. Since I had the code working, I'm opening this PR. However, there is further discussion whether this is something we will actually need.


version.props
```diff
@@ -28,6 +28,8 @@
     <ExperimentalVersionSuffix Condition=" '$(ExperimentalVersionSuffix)' != '' AND '$(BuildNumber)' != '' ">$(ExperimentalVersionSuffix)-$(BuildNumber)</ExperimentalVersionSuffix>
 
     <SharedFxCliBlobChannel>release/2.1</SharedFxCliBlobChannel>
+
+    <PreviousLzmaUrl>https://dotnetfeed.blob.core.windows.net/orchestrated-release-2-1/20180515-07/final/assets/aspnetcore/Runtime/2.1.0/nuGetPackagesArchive-2.1.0.lzma</PreviousLzmaUrl>
```
natemcmaster: This makes me nervous. It makes our builds un-repeatable if this remote file were to go missing for any reason. Although I hope it doesn't, things like this have happened in the recent past, so I'd rather avoid this.

Since the contents of the LZMA aren't actually important, what do you think about taking a slightly different approach? Instead of downloading the last LZMA on each build, what if we just generated a .txt file in aspnet/Universe representing all file entries from the last archive. We can use this to produce a diff.

JunTaoLuo: I'm not sure if we are going to be automating this step since we should theoretically never need to build "delta" lzma archives for antares between different patch versions. We should never update the default versions again and therefore restore will always look for the 2.1.1 versions in the fallback folder. Even if we did, I think it will be rare enought that we will probably be running this task manually as needed so I think specifying urls and filenames should be okay. I put the logic inside build/ since it was the most convenient but maybe it's better to move this elsewhere?

Note that the first time we will be producing this antares zip will be 2.1.1. It'll be good to check the filtering logic and once we have a final 2.1.1 build, I'll test it on a private stamp before handoff.

natemcmaster: What do you think about the idea of capturing the data as a .txt file instead of downloading a LKG .lzma file?

JunTaoLuo: I'm not sure the added complexity and layer of indirection is worth it, though I could certainly change the mechanism if you feel strongly about it.

natemcmaster: Yes, I feel strongly about having a way to produce new builds if this blob were deleted or (perhaps worse) its contents changed.

JunTaoLuo:
:up::date: I'm saving a manifest of the contents in the lzma as a props file.

natemcmaster:
Looks good. Thanks for updating with my feedback.

PR #1200, opened 2018-06-06
Gather and merge sign requests from repo builds

natemcmaster:
Part of https://github.com/aspnet/Coherence-Signed/issues/784

Depends on https://github.com/aspnet/BuildTools/pull/688

### DIFF
--- a/build/RepositoryBuild.targets
+++ b/build/RepositoryBuild.targets
@@ -7,6 +7,7 @@
     <_NoBuildRepos>$(NoBuild)</_NoBuildRepos>
     <_BuildScriptToExecute Condition="'$(OS)'!='Windows_NT'">build.sh</_BuildScriptToExecute>
     <_BuildScriptToExecute Condition="'$(OS)'=='Windows_NT'">build.cmd</_BuildScriptToExecute>
+    <_SignRequestIntermediateOutputPath>$(IntermediateDir)signrequests\</_SignRequestIntermediateOutputPath>
   </PropertyGroup>
 
   <Target Name="GetRepoBatches" DependsOnTargets="GeneratePropsFiles;ComputeGraph">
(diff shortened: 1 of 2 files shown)


build/RepositoryBuild.targets
```diff
@@ -23,13 +24,22 @@
   </Target>
 
   <Target Name="_BuildRepositories" DependsOnTargets="GetRepoBatches" Condition=" @(RepositoryBuildOrder->Count()) != 0 ">
+    <RemoveDir Directories="$(_SignRequestIntermediateOutputPath)" />
```
ryanbrandenburg: I assume you're just ensuring clean directories here?

build/RepositoryBuild.targets
```diff
@@ -23,13 +24,22 @@
...
       BuildInParallel="true"
       StopOnFirstFailure="true"
       Targets="_BuildRepository"
       Properties="BuildGroup=%(BatchedRepository.BuildGroup);BuildNumber=$(BuildNumber);IsFinalBuild=$(IsFinalBuild);Configuration=$(Configuration)" />
 
+    <ItemGroup>
+      <SignRequests Include="$(_SignRequestIntermediateOutputPath)*.xml" />
+    </ItemGroup>
+
+    <MergeXmlFiles Files="@(SignRequests)" OutputPath="$(ArtifactsDir)signrequest.xml" />
```
ryanbrandenburg: Neat! Is doing this faster than sending multiple sign requests?

build/RepositoryBuild.targets
natemcmaster: yes.

build/RepositoryBuild.targets
natemcmaster: Yes. There is a fixed cost each time you have to round-trip to code-signing which is pretty high. So the more files you can send per batch, the better.

Issue #706, opened 2017-12-12
Test generated shared framework


JunTaoLuo:
Add tests to test end to end functions of the shared framework. We can test using templates, e.g. dotnet new mvc, and/or using something like MusicStore.

Eilon:
Can you come up with a specific list of tests and we can triage again?

JunTaoLuo:
# Variations

## Platforms:
Each of the follwing os and bitness must be tested
- win-x64
- win-x86
- osx-x64
- linux-x64

## Installers:
Each of the following types of installer must be tested
### Installers: 
- Windows: Windows Hosting Bundle (.exe) SDK (.exe)
- macOS: SDK (.pkg)
- Linux: SDK (.deb, .rpm)
### Archives:
- Windows: SDK (.zip)
- macOS: SDK (.tar.gz)
- Linux: SDK (.tar.gz)

## Metapackages
While the default templates will use .App as its metapackage, we need to ensure an app using either of the following metapackage can execute
- Microsoft.AspNetCore.App
- Microsoft.AspNetCore.All

## App templates
We want to test as many different types project templates. At a minimum we should verify using mvc and razor templates
- mvc aka ASP.NET Core Web App (Model-View-Controller)
- razor aka ASP.NET Core Web App
Each template should be tested with as many variations as possible. For example mvc and razor templates can be configured with individual auth, to use local db, or to use BrowserLink, etc

## Tooling Experience
Complete the test using VS tooling and CLI
- Visual Studio
- dotnet cli
- VSCode (optional?)
- VS for mac (optional?)

# Scenarios

## dotnet new and offline restore
Create a new app using the dotnet new command and ensure restore works offline. This ensures our packages are included in the LZMA/offline restore cache.

## dotnet build and dotnet run
Builds successfully and can run the app. The build output should not include any binaries from the shared frameworks (i.e. should only contain the app's main dll)

## dotnet publish portable and run the published outupt
Publishes successfully and can run the app. The publish output should also only inclulde the app's main dll

## dotnet publish portable and disable publish trimming
Publishes successfully and can run the app. Add ExcludeAssets="Build" to the reference of the metapackage and the publish output should now include the app's main dll and all the aspnetcore binaries referenced.

## dotnet publish standalone
Publish the app and specify a runtime. The published output should include the app's dll, aspnetcore binaries and dotnet core binaries. The published output should be able to run on a fresh machine with no dotnet installed.

## debugging
[...]

PR #1201, opened 2018-06-07
Begin to add unit tests for the shared framework

natemcmaster:
I'm preparing some work soon that will affect the way we generate the shared framework. Adding tests now ahead of time to ensure refactoring preserves expected output.

TODO (in follow-up PRs)
~~* Automate these tests to run against actual build output~~
* Add more coverage

cref https://github.com/aspnet/Universe/issues/706

cc @mikeharder @JunTaoLuo

### DIFF
--- a/build/SharedFx.targets
+++ b/build/SharedFx.targets
@@ -197,7 +197,7 @@
       <VersionLines Include="$(PackageVersion)" />
     </ItemGroup>
 
-     <!-- Publish -->
+    <!-- Publish -->
     <MSBuild Projects="$(SharedFxWorkDirectory)SharedFx.csproj"
       Targets="Publish"
       Properties="$(CommonSharedFxProps);GenerateRuntimeConfigurationFiles=true;SelfContained=false;PublishDir=$(SharedFxPublishDirectory)" />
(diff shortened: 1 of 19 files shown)


natemcmaster:
Found a bug in the sharedfx. (Fix proposed here https://github.com/aspnet/Universe/pull/1203). I've also added a test to catch this in the future.

natemcmaster:
:up: :date: this is now wired up to run on automatically when the sharedfx is built

build/SharedFx.targets
```diff
@@ -528,6 +528,21 @@
       Overwrite="true"/>
   </Target>
 
-  <Target Name="BuildSharedFx" DependsOnTargets="GeneratePropsFiles;ResolveSharedFxFiles;CrossGenAssemblies;CrossGenSymbols;PackSharedFx"/>
+  <Target Name="BuildSharedFx" DependsOnTargets="GeneratePropsFiles;ResolveSharedFxFiles;CrossGenAssemblies;CrossGenSymbols;PackSharedFx;TestSharedFx"/>
+
+  <Target Name="TestSharedFx" DependsOnTargets="DefineSharedFxPrerequisites;ResolveCommitHash">
+    <PropertyGroup>
+      <UnitTestFxProject>$(MSBuildThisFileDirectory)..\test\SharedFx.UnitTests\SharedFx.UnitTests.csproj</UnitTestFxProject>
```
JunTaoLuo: nit: $(RepositoryRoot) instead of $(MSBuildThisFileDirectory)..

build/tools/templates/Archive/Archive.csproj
```diff
@@ -1,4 +1,10 @@
-<Project Sdk="Microsoft.NET.Sdk">
+<Project>
+  <PropertyGroup>
+    <ImportDirectoryBuildProps>false</ImportDirectoryBuildProps>
```
JunTaoLuo: Curious what this change is for?

natemcmaster: By moving Directory.Build.props up to the root of the project, it had a wide impact on the .csproj files we use. I just updated with some cleanup that I think makes it easier to keep our .csproj files minimal.

build/SharedFx.targets
natemcmaster: Updated.

PR #1210, opened 2018-06-13
[2.0] Add script used to deploy blobs to Azure storage

natemcmaster:
Let's stop doing this by hand.

TODO:
How can we reliably also fetch the assets for the preview 2.0 build? At the moment, I've been manually copying them. Thoughts @JunTaoLuo ?

### DIFF
--- a/scripts/UploadBlobs.ps1
+++ b/scripts/UploadBlobs.ps1
@@ -1,0 +1,142 @@
+<#
+.SYNOPSIS
+Deploys a build to an Azure blob store
+
+.PARAMETER AccountName
+The account name for the Azure account
+
+.PARAMETER AccountKey
+The account key for the Azure account
+
+.PARAMETER BuildNumber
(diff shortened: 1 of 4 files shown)


natemcmaster:
Talked in person with @JunTaoLuo. Let's not both automating manually copying previous releases. Instead, let's cleanup the way we produce drop shares.

natemcmaster:
FYI @muratg - merging as this is something we would have done by hand anyways. It doesn't impact the build yet as the automation to run this script isn't wired up.

scripts/UploadBlobs.ps1
```diff
@@ -0,0 +1,142 @@
+<#
```
JunTaoLuo: Where does this copy over the artifacts from previous release?

natemcmaster: It doesn't. I'm still doing that manually.

PR #1221, opened 2018-06-19
Pass KoreBuildVersion to repo builds

ryanbrandenburg:
I believe this should force Repo builds to use Universe's version of Internal.Sdk, which they don't seem to do ATM.


natemcmaster:
This shouldn't be necessary. Are you seeing issues in this area? The _UpdateRepoLockFile target should be overriding the korebuild-lock.txt to match Universe's korebuild version, which in turn should cause [/p:KoreBuildVersion](https://github.com/aspnet/BuildTools/blob/19de84393e58d955b63078025b0ac9bc57dd2df9/files/KoreBuild/scripts/KoreBuild.psm1#L94) to be consistent.

ryanbrandenburg:
I don't remember the specifics now, but that wasn't happening. I THINK we were getting a different version of the Internal.SDK in repo builds which hadn't updated to the new KoreBuild yet.

natemcmaster:
Hmm, if you show me error logs or steps to reproduce, that would help. I don't think setting KoreBuildVersion wouldn't fix issues with Internal.Sdk versions. The version of Internal.Sdk is set via dependencies.props and overridden via DotNetPackageVersionPropsPath. aspnet/Universe is the only place where Internal.Sdk is set to `$(KoreBuildVersion)`

PR #1225, opened 2018-06-23
Give BenchmarksOnly* `<ExternalDependency>`s `VariableName` metadata

dougbu:
- should improve `UpgradeDependencies` at least in MVC repo
- hopefully will also correct build break in Update Universe builds


dougbu:
Mostly FYI. May need to check this in to determine if it fixes Update Universe build.

PR #1226, opened 2018-06-23
Support packages both built in Universe and used as external dependencies

dougbu:
- `AnalyzeBuildGraph` fix should correct Update Universe build break
- `VerifyCoherentVersions` fix should correct UniverseCoherence build break

nits:
- add and rename a couple of variables for clarity and reduced repetition
- remove and sort `using`s


dougbu:
Mostly FYI. Likely need to check this in to determine if it fixes UniverseCoherence and Update Universe builds.

natemcmaster:
Can you give me more context on this issue? Which package produced in aspnet needs to be treated as an external dependency?

Also, while I applaud your effort to unblock broken builds, it would be good to have someone review changes which could have large impact on the way we manage version consistency and coherence within aspnet. This is area of the product is prone to bugs and we've added lots of checks to ensure we are doing things right. I'm hesitant to loosen the checks and tests without understanding the need for it, and uncomfortable with the fact that this went in without discussion or review. cc @mkArtakMSFT @muratg

dougbu:
The context is aspnet/Mvc#7944. In discussions leading up to that PR, @ryanbrandenburg was concerned about adding EF dependencies to the MVC repo. I believe his reasoning was (a) it might or would perturb the Universe build graph and (b) dependencies tend to get used, leading to product code dependencies we'd like to avoid. I set things up to depend on EF from Nuget.org and not the current build (with fairly clear property names). We thought that would **Just Work:tm:** but I hit the issues described in this PR and un-broke the build.

If you're in the office and have additional questions or suggestions, suggest we chat in person.

dougbu:
BTW if you have specific comments on this PR, I can follow up with another commit.

Separately, I was surprised to find MVC already builds two groups _later_ than EF. Perhaps (a) wasn't a valid concern? If so, we could use distinct variables but stick to the latest (Universe) EF versions.

natemcmaster:
Thanks for sharing more context. I see now there were related PRs reviewed and approved by others. I was more concerned about the process -- not the changes themselves -- when I saw a PR opened and merged without discussion or review.

I have a few more questions about this change. It your desire to use EF Core 2.1 when benchmarking MVC 2.2? Or was EF 2.1 used just to workaround KoreBuild/Universe restrictions? Do you also want to benchmark EF Core 2.2 with MVC 2.2?

dougbu:
> It your desire to use EF Core 2.1 when benchmarking MVC 2.2?

No. When benchmarking, the new apps use the All meta-package and thus pick up the latest EF packages.

> Or was EF 2.1 used just to workaround KoreBuild/Universe restrictions?

Yes. The approach we landed on _should_ have been simpler than e.g. creating a unique build configuration for Universe and prodcon builds that excludes the benchmarks apps and the related functional tests. We also wanted these apps to build and test most of the time to make sure they keep working i.e. to avoid finding out about a break 5 days later when we notice something bad in the aspnet/benchmarks graphs.

> Do you also want to benchmark EF Core 2.2 with MVC 2.2?

Yes, this will be what happens automatically and repeatedly as soon as aspnet/benchmarks#495 goes in. (So far, I've only done manual benchmarking of the new apps. That also picked up the latest of everything.)

PR #1245, opened 2018-07-06
Refactor LZMA generation to support Docker and Antares usage

natemcmaster:
Changes:

* List explicitly as .csproj files the scenarios for which the offline package cache is important
* Explicitly define the artifacts for various usage, such as:
  * CLI - LZMA, first run experience
  * Docker (where xml doc files are not needed)
  * Antares (where 1.x SDKs must still be supported, but xml docs are not needed)
  * Antares  (where 1.x SDKs do not need to be supported)

Output:
image


natemcmaster:
FYI this is necessary to react to https://github.com/aspnet/Home/issues/3292. Since the baseline is going to rev each package, we'll need new antares archives and updated packages in the lzma for reach runtime update.

src/PackageArchive/Scenario.ConsoleApp/Scenario.ConsoleApp.csproj
```diff
@@ -0,0 +1,15 @@
...
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <MaxImplicitVersion>2.1.0</MaxImplicitVersion>
+    <!-- Use pre-release versions up until 2.1.0, then don't lift higher than "2.1.0" -->
+    <MicrosoftNETCoreAppPackageVersion Condition="$(MicrosoftNETCoreApp21PackageVersion.StartsWith('$(MaxImplicitVersion)-'))">$(MicrosoftNETCoreApp21PackageVersion)</MicrosoftNETCoreAppPackageVersion>
+    <MicrosoftNETCoreAppPackageVersion>$([MSbuild]::ValueOrDefault('$(MicrosoftNETCoreAppPackageVersion)','$(MaxImplicitVersion)'))</MicrosoftNETCoreAppPackageVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETCore.App" Version="$(MicrosoftNETCoreAppPackageVersion)" />
```
JunTaoLuo: Is it possible to combine this with the webapp?

build/tasks/GetDocXmlFiles.cs
```diff
@@ -0,0 +1,64 @@
...
+            {
+                try
+                {
+                    using (var file = File.OpenRead(f.ItemSpec))
+                    using (var reader = new StreamReader(file))
+                    {
+                        string line;
+                        for (var i = 0; i < 2; i++)
+                        {
+                            line = reader.ReadLine();
```
JunTaoLuo: instead of essentially writing your custom XML parser, is there an xml parse we can use to ensure we have a top level <doc> element instead?

natemcmaster: I tried that. Unfortunately, some xml docsfiles contain corrupted xml, so XDocument.Load fails to parse. :-/. It seems VS and roslyn are more forgiving on XML correctness.

By the way, this detected that _all_ .xml files in the restore folder are xmldocs. I didn't find any .xml files in packages which were not xmldocs. Is that your expectation too?

src/PackageArchive/Scenario.ConsoleApp/Scenario.ConsoleApp.csproj
natemcmaster: Maybe, but i wanted this refactoring to be explicitly clear about the types of projects and the scenarios that the offline cache supports. I think the code has been unclear about this in the past because we were trying too hard to optimize.

build/tasks/GetDocXmlFiles.cs
JunTaoLuo: That's what I found as well.

JunTaoLuo: FYI if there are corrupted xmls, we should probably at least notify the relevant teams that it's not proper XML, even though it may not matter for VS

natemcmaster: That's a really good point. Did a quick check...looks like it was only one file out of 3505.
```
/Users/namc/dev/aspnet/Universe21/obj/pkgs/netstandard.library/2.0.3/build/netstandard2.0/ref/netstandard.xml	The 'p' start tag on line 654 position 2 does not match the end tag of 'th'. Line 654, position 86.
```

natemcmaster: https://github.com/dotnet/standard/issues/794

src/PackageArchive/Scenario.ConsoleApp/Scenario.ConsoleApp.csproj
JunTaoLuo: I guess the reason why I'd want to combine this with WebApp is that the web app will also contain an implicit reference to NETCore.App as well. Class library is completely different so I can see why that's a separate scenario.

PR #1247, opened 2018-07-10
Pin to the latest stable 2.1.x corefx packages

natemcmaster:
cc @weshaggard @mmitche  - should unblock prodcon for 2.2 since corefx isn't producing anything new.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -1,36 +1,10 @@
-<Project>
+﻿<Project>
   <!-- These package versions may be overridden or updated by automation. -->
   <PropertyGroup Label="Package Versions: Auto" Condition=" '$(DotNetPackageVersionPropsPath)' == '' ">
-    <MicrosoftCSharpPackageVersion>4.6.0-preview1-26617-01</MicrosoftCSharpPackageVersion>
     <MicrosoftExtensionsDependencyModelPackageVersion>2.2.0-preview1-26618-02</MicrosoftExtensionsDependencyModelPackageVersion>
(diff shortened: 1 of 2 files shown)


build/dependencies.props
```diff
@@ -119,13 +94,15 @@
...
     <MicrosoftWebXdtPackageVersion>1.4.0</MicrosoftWebXdtPackageVersion>
+    <MicrosoftWin32RegistryPackageVersion>4.5.0</MicrosoftWin32RegistryPackageVersion>
     <MonoAddinsPackageVersion>1.3.8</MonoAddinsPackageVersion>
     <MonoDevelopSdkPackageVersion>1.0.1</MonoDevelopSdkPackageVersion>
     <MoqPackageVersion>4.7.49</MoqPackageVersion>
     <NETStandard16PackageVersion>1.6.1</NETStandard16PackageVersion>
     <NETStandardLibrary20PackageVersion>2.0.3</NETStandardLibrary20PackageVersion>
     <NewtonsoftJsonBsonPackageVersion>1.0.1</NewtonsoftJsonBsonPackageVersion>
     <NewtonsoftJsonPackageVersion>11.0.2</NewtonsoftJsonPackageVersion>
+    <NuGetFrameworksPackageVersion>4.7.0</NuGetFrameworksPackageVersion>
```
weshaggard: This isn't part of corefx so I'm not sure if needs pinned or not.

natemcmaster: Good point. This one was incidentally included. Even if NuGet produces new things in 2.2, i'm not sure we really need updates. This one is only used in our scaffolding tools, and the one or two APIs they use haven't changed since 3.x.

PR #1257, opened 2018-07-12
Add configurable patch policies for servicing updates

natemcmaster:
This implements a per-repo patch policy, which allows us some more flexibility in how we decide to cascade versions (or not). 

Changes:

* remove coherent version check. Will replace in the future with a more robust test based on the policy each repo. (TODO open an issue)
* include the templating submodule as a normal part of the repo build targets
* A bunch of updates to submodules.
  * Removed some repos from the patch which don't need to be serviced according to the new policy.
  * Move variables to 'pinned' when that repo is following the "ProductChangesOnly" patch policy. Example: https://github.com/aspnet/AADIntegration/compare/bb49eeebe64905ce705f0e70f59869085adc9dce...6f02048c3c5c0ee6684c241721422c2b8eebddbc
  * Updated to the 2.1.2 baseline so there should be no downgrades in 2.1.3 from 2.1.2

Resolves https://github.com/aspnet/Home/issues/3316

cc @DamianEdwards @davidfowl


build/artifacts.props
```diff
@@ -216,6 +219,7 @@
     <PackageArtifact Include="Microsoft.Extensions.DependencyInjection.Abstractions" AllMetapackage="true" AppMetapackage="true" Category="ship" />
     <PackageArtifact Include="Microsoft.Extensions.DependencyInjection.Specification.Tests" Category="ship" />
     <PackageArtifact Include="Microsoft.Extensions.DependencyInjection" AllMetapackage="true" AppMetapackage="true" Category="ship" />
+    <PackageArtifact Include="Microsoft.Extensions.DiagnosticAdapter" AllMetapackage="true" AppMetapackage="true" Category="ship" />
```
JunTaoLuo: We are adding packages to the metapackages in a patch?

JunTaoLuo: Nevermind, these are moved from external-dependencies.

build/submodules.props
```diff
@@ -2,59 +2,89 @@
...
-    <Repository Include="Security" />
-    <Repository Include="ServerTests" />
-    <Repository Include="Session" />
     <Repository Include="SignalR" />
-    <Repository Include="StaticFiles" />
-    <Repository Include="WebSockets" />
-    <!-- <Repository Include="Templating" /> -->
+    <Repository Include="Scaffolding" PatchPolicy="AlwaysUpdate" />
+    <Repository Include="MetaPackages" PatchPolicy="CascadeVersions" />
+    <Repository Include="Templating" PatchPolicy="AlwaysUpdateAndCascadeVersions" />
```
JunTaoLuo: Care to explain why `AlwaysUpdateAndCascadeVersions` exists? Are `CascadeVersions` and `AlwaysUpdate` two independent concepts?

build/tasks/ProjectModel/PatchPolicy.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace RepoTasks.ProjectModel
+{
+    [Flags]
+    internal enum PatchPolicy
+    {
+        /// <summary>
+        ///     Only produce new package versions if there were changes to product code.
+        /// </summary>
+        ProductChangesOnly = 1 << 0,
```
JunTaoLuo: Sorry if this is obvious but I didn't see how this was used. Where are you checking to ensure that if this flag is set, we will check whether the repo contains new changes to the product code and therefore requires an update?

natemcmaster: Not in this update, but hopefully soon when we finally address https://github.com/aspnet/Coherence-Signed/issues/858

build/submodules.props
natemcmaster: Scaffolding = AlwaysUpdate. Due to the way they've design VS, this needs to ship whenever aspnetcore.app does, *BUT* it doesn't need to update its dependencies to latest

MetaPackages = CascadeVersions. If there are changes to its dependencies, it should reship because it should help nesure transitive versions are latest

Templating = both because should be true. Use latest, and always update.

They could be independent vars, but I got tired of having dozens of boolean flags

natemcmaster:
I've opened https://github.com/aspnet/Universe/issues/1258 to follow up and add more tests to assert correctness

Issue #1256, opened 2018-07-12
Add a package dependency from Microsoft.AspNetCore.{App,All} => NETCore.App


natemcmaster:
Part of implementing https://github.com/aspnet/Home/issues/3307 we need to resolve is how to make sure both AspNetCore.App and NETCore.App end up in the restore graph so you get all the right ref assemblies.

@JunTaoLuo looked into this in the past but ran into issues. We should take a closer look at those again to inform our strategy about how we should move forward with https://github.com/aspnet/Home/issues/3307

JunTaoLuo:
I forget the issue we ran into originally, but it may have been resolved. A few things we would want to check are:
1. The correct set of build/compile targets are brought in correctly
2. The deps files generated are compatible with MVC features
3. The generated shared frameworks in our build contains the full set of binaries required for layered frameworks
4. No restore graphs and publish output regression

dsplaisted:
Instead of adding a package dependency, why wouldn't we just have two separate implicit package references?

natemcmaster:
It looks like there may be one side effect. The .deps.json generation produces a slightly difference result when NETCore.App is transitive.

When you have NETCore.App and AspNetCore.App as direct deps, you get.
```
      "Microsoft.NETCore.App/2.1.0": {
        "dependencies": {
          "Microsoft.NETCore.DotNetHostPolicy": "2.1.0",
          "Microsoft.NETCore.Platforms": "2.1.0",
          "Microsoft.NETCore.Targets": "2.1.0",
          "NETStandard.Library": "2.0.3"
        },
        "compile": {
          "ref/netcoreapp2.1/[BUNCH OF ENTRIES FOR System*.dll HERE]": {},
        }
      },
```

But with only AspNetCore.App as the top-level dep and NETCore.App is transitive, you get,


```diff
      "Microsoft.NETCore.App/2.1.3-servicing-26708-02": {
        "dependencies": {
          "Microsoft.NETCore.DotNetHostPolicy": "2.1.3-servicing-26708-02",
          "Microsoft.NETCore.Platforms": "2.1.1-servicing-26708-02",
          "Microsoft.NETCore.Targets": "2.0.0",
          "NETStandard.Library": "2.0.3"
        },
        "compile": {
          "ref/netcoreapp2.1/[BUNCH OF ENTRIES FOR System*.dll HERE]": {},
        }
+       "compileOnly": true
      },
```

[...]

PR #1261, opened 2018-07-13
wip: Add a package dependency from Microsoft.AspNetCore.App/All to NETCore.App

natemcmaster:
Resolves https://github.com/aspnet/Universe/issues/1256

This adds a package dependency from AspNetCore.App to NETCore.App. This change would need to be added along with https://github.com/dotnet/sdk/pull/2394.

cc @nguerrera @dsplaisted  @DamianEdwards


src/Microsoft.AspNetCore.App/Microsoft.AspNetCore.App.csproj
```diff
@@ -15,4 +16,8 @@
     <Content Include="lib\$(TargetFramework)\_._" PackagePath="%(Identity)" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETCore.App" Version="[$(RuntimeFrameworkVersion), 2.2.0)" PrivateAsset="None" />
```
JunTaoLuo: Which version would we need to target? The bundled version (latest patch) or the pinned version (major.minor.0)?

natemcmaster: If we add this at all, it should be latest patch. Otherwise, adding this only creates problems and doesn't solve any.

natemcmaster:
Per discussion, this is not approved for 2.1.

PR #1287, opened 2018-07-30
Replace AppVeyor and Travis with VSTS for PR validation builds

natemcmaster:
Part of moving to VSTS. This runs a quick check on Universe for PR validation. It's not doing a full build which currently takes too long.


.vsts/builds/fast-pr-validation.yml
```diff
@@ -0,0 +1,24 @@
...
+phases:
+- template: .vsts-pipelines/templates/phases/default-build.yml@buildtools
+  parameters:
+    agentOs: Windows
+    buildArgs: "/t:CheckUniverse"
+
+phases:
+- template: .vsts-pipelines/templates/phases/default-build.yml@buildtools
+  parameters:
+    agentOs: Linux
```
JunTaoLuo: We don't have a macos agent?

natemcmaster: Good point, I'll add that. We had excluded macOS from PR validation in Travis in the past because their agents were too slow. The VSTS agents are much better.

PR #1290, opened 2018-08-01
Use OSS package versions consistent with aspnet/benchmarks and Microsoft.AspNetcore.All 2.1.2

dougbu:
- update our own NuGet packages to align lower-level dependencies


dougbu:
165c712201

PR #1298, opened 2018-08-07
Prepare 2.0.10 patch

natemcmaster:


version.props
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <AspNetCoreMajorVersion>2</AspNetCoreMajorVersion>
     <AspNetCoreMinorVersion>0</AspNetCoreMinorVersion>
-    <AspNetCorePatchVersion>9</AspNetCorePatchVersion>
+    <AspNetCorePatchVersion>10</AspNetCorePatchVersion>
```
JunTaoLuo: I think you need to also update `PreviousRuntimeStoreArchiveVersion`'s build number so that it ends up as 2.0.9-10031.

natemcmaster: Good catch. Updated.

PR #1310, opened 2018-08-10
Update package versions that are important to new completion APIs.

NTaylorMullen:
aspnet/Razor#2227


NTaylorMullen:
This is only for 3.0

PR #1317, opened 2018-08-15
Add EFCore.Cosmos.Sql to shipping package list

smitpatel:
Reaction to https://github.com/aspnet/EntityFrameworkCore/pull/12843

@natemcmaster - Anymore changes required anywhere?

@ajcvickers @divega - Currently this package is not included in App/All Metapackage. We should decide on that.


natemcmaster:
>  Currently this package is not included in App/All Metapackage.

It should not go into .App. Adding to .All doesn't buy you much because we're deprecating .All in 3.0 anyways but it doesn't matter to me if you want to add it anyways. https://github.com/aspnet/Announcements/issues/314

natemcmaster:
Once https://github.com/aspnet/EntityFrameworkCore/pull/12843 is merged, the change here looks fine, by the way. :shipit:

PR #1322, opened 2018-08-16
Fix whitespace in task parameters

natemcmaster:
In MSBuild < 15.8, new line characters are removed. In 15.8, they are not. This removes newlines from arguments which are meant to run on bash as a single line.


natemcmaster:
I'm going to merge before EOD to unblock CI.

PR #1338, opened 2018-08-27
OSS Detection and Registration

anurse:
Ok, so here's the first pass at this. Super simple.

The first quirk I've noticed is that it doesn't have a way to distinguish between NuGet packages we build ourselves vs. packages we depend on. So it wants to register `Microsoft.AspNetCore.*` as OSS. Now, we could just let that happen and say it's Internal-only to avoid things, or I can investigate ways to suppress them from the output (which may need changes in the tool itself). Thoughts @natemcmaster @Eilon ?

TODO:
* [x] Decide on how to handle our packages
* [x] Change `scanType` to `Submit` to actually produce the registrations
* [x] Change `governanceProduct` to the real one we plan to use


natemcmaster:
> So it wants to register Microsoft.AspNetCore.* as OSS. 

Can we give this feedback to the authors of this plugin? I'm afraid this is going to be unbearably noisy, to the point that I'm not sure using their automation is the right fit for the job.

anurse:
> to the point that I'm not sure using their automation is the right fit for the job.

I'm not convinced it goes quite that far. You're not the one who has to manually register our NPM packages ;P

anurse:
Another option would be to turn this on for NPM packages only. We can configure the specific detectors we want to run.

anurse:
Sounds like this behavior may be by-design and by-policy now.

Eilon:
@anurse are we good to merge?

anurse:
I'll rebase and then yes, we should be good.

PR #1320, opened 2018-08-16
Add package descriptions for Microsoft.AspNetCore.{App, All}

natemcmaster:
Resolves https://github.com/aspnet/Universe/issues/1321

Replaces the placeholder with real description, including a link to the runtime installers which are required to use these packages.

PR #1343, opened 2018-08-29
Minor tweaks to meta-package descriptions

natemcmaster:
Per feedback on #1320 which came in post-merge.

cc @Eilon


natemcmaster:
I'm taking the liberty of marking as 'servicing approved' since I believe this is just a continuation of #1320

Eilon:
Yeah I'd say this is just as approved as its predecessor.

PR #1345, opened 2018-08-29
Add new options package

HaoK:
@natemcmaster


natemcmaster:
Build is still broken. To add new packages, you also need to update universe's submodule. The PR checks on this repo are correct, by the way. It they are red, the build will fail.

https://dotnet.visualstudio.com/public/_build/results?buildId=16464&view=logs

PR #1348, opened 2018-08-29
Update options module

HaoK:
Does this look correct @natemcmaster


natemcmaster:
Yes, thank you.

HaoK:
Can we put this in the readme for Universe?  I had to search my email inbox to find the instructions on how to update universe...

```
2.	aspnet/Universe uses git submodules to pin to exact commit shas of sub-repos. You should checkout the release/2.0.0 branch of Universe and also update the git submodule to point to the new commit you’ve added. You can do this with these steps

cd Universe
git checkout release/2.0.0
git pull
git submodule update --init –-remote modules/MyRepo
git add modules/MyRepo
git commit -m “Updated MyRepo to include changes from aspnet/MyRepo#123”

```

natemcmaster:
So, something like this? https://github.com/aspnet/Universe/blob/master/docs/Submodules.md#saving-an-update-to-a-submodule

HaoK:
Yeah maybe just a link then :)  I searched my email for submodule instead of the repo

PR #1349, opened 2018-08-29
Add Microsoft.AspNetCore.DataProtection.EntityFrameworkCore

natemcmaster:
Marking this as ship, but not adding to metapackages or shared frameworks.

React to https://github.com/aspnet/DataProtection/pull/323


natemcmaster:
Nope.

PR #1351, opened 2018-08-29
Functional tests for .NET Core CLI

mikeharder:
Current behavior which we may want to change:
- Uses first `dotnet` executable in path
- Uses env var `NUGET_PACKAGE_SOURCE` to specify private feed(s)


mikeharder:
Closing to replace with new PR which preserves history

PR #1360, opened 2018-08-30
Add a script for tagging TeamCity builds

ryanbrandenburg:
Part of the fix for aspnet/Coherence-Signed#975. In order to avoid trying to run Update Repos on ProdCon builds we have to tag every non-prodcon build of Publish so we can use that in the Update Repos dependency definition.


scripts/Tag-TeamCityBuild.ps1
```diff
@@ -0,0 +1,37 @@
...
+)
+
+$ErrorActionPreference = 'Stop'
+Set-StrictMode -Version 2
+
+$authInfo = "${UserName}:$Password"
+$authEncoded = [System.Convert]::ToBase64String([System.Text.Encoding]::ASCII.GetBytes($authInfo))
+$basicAuthValue = "Basic $authEncoded"
+
+$headers = New-Object "System.Collections.Generic.Dictionary[[String],[String]]"
```
natemcmaster: Use a powershell map:
```ps1
$headers = @{
    Authorization  = "Basic $authEncoded"
	"Content-Type" = "text/plain"
}
```

scripts/Tag-TeamCityBuild.ps1
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env pwsh
```
natemcmaster: Does this file have `+x` permissions?

scripts/Tag-TeamCityBuild.ps1
natemcmaster: Huh, it works for me, but I'm using PowerShell 6. What are you using?

The quotes are necessary because dash is an invalid character in PS identifier syntax, so we need to make the content-type entry a string key instead.

ryanbrandenburg: I was imprecise in my wording, I meant that when I did `Content-Type` without the quotes it failed. It didn't give me a very descriptive error message so I didn't know the `-` was the problem. Thanks!

PR #1365, opened 2018-08-30
Start shipping templates to NuGet

ryanbrandenburg:
Note that I marked the ItemTemplates as `ship` as well.


ryanbrandenburg:
Filed #721 about the description.

PR #1376, opened 2018-09-04
Install MicroBuild code signing plugin and set SignType = real

natemcmaster:
Part of https://github.com/aspnet/Coherence-Signed/issues/694

This makes available the infrastructure to run code signing. Additional work is necessary to plumb aspnet bits into the signing tools.


natemcmaster:
@HaoK can you take a look?

.vsts/builds/ci-official.yml
```diff
@@ -18,11 +18,19 @@ phases:
   variables:
     CI: true
     DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+    TeamName: AspNetCore
+    _SignType: real
```
HaoK: is this just to signify its a private var used in this yml only?  just curious what the pattern is

HaoK: Also what are the other sign types, test? (maybe worth a comment for posterity)

natemcmaster: SignType = { test, real }

I actually used the underscore for a different reason. VSTS variables automatically become environment variables, and this one was causing issues in MSBuild projects which use the `$(SignType)` msbuild prop.

HaoK: Ok cool, good to know :)

natemcmaster: I added this to the yml as comments for future reference.

PR #1393, opened 2018-09-07
Use built in DownloadFile task

ryanbrandenburg:
The KoreBuild task was deleted, so we have to switch to the built in one.


ryanbrandenburg:
Merging this to fix the build.

PR #1411, opened 2018-09-20
Remove duplicate CompilerServicesUnsafe

ryanbrandenburg:
A bad merge gave us two copies of `SystemRuntimeCompilerServicesUnsafePackageVersion` I'm removing this one in favor of the pinned one for 4.5.1 as suggested by @pakrym [here](https://github.com/aspnet/Coherence-Signed/issues/1069#issuecomment-423224840). I'm going to merge this in directly to get the build flowing again, but will respond to any feedback left after the fact.


pakrym:
It's not what I suggested. I was talking about System.Memory.

natemcmaster:
Don't we want the latest version of this package to match what's in .NET Core 3.0? Why are we pinning to 4.5.1.?

natemcmaster:
Nvm, I see you resolved already.

PR #1414, opened 2018-09-23
Adding win-arm shared framework

Petermarcu:


build/SharedFx.props
```diff
@@ -70,6 +70,7 @@
 
   <ItemGroup>
     <WindowsSharedFxRIDs Include="win-x64;win-x86"/>
+    <WindowsSharedFxRIDs Include="win-arm" CrossgenSymbols="false" />
```
natemcmaster: Do we have a way to crossgen win-arm builds on CI? I'm not familiar with what's required to emulate or run win-arm executables on x64 machines. If not, this should be changed to `CrossGen="false"`.

JunTaoLuo:
Temporarily closing this since we are doing some repository updates. I'll reopen this as part of https://github.com/aspnet/AspNetCore-Internal/issues/1128.

PR #1415, opened 2018-09-24
 Functional tests for .NET Core CLI 2.1.X

mikeharder:


.vsts/builds/e2e-tests.yml
```diff
@@ -0,0 +1,61 @@
...
+        Test.RuntimeIdentifier: win-x64
+      SelfContainedLinux:
+        Test.RuntimeIdentifier: linux-x64
+      SelfContainedMacOs:
+        Test.RuntimeIdentifier: osx-x64
+  steps:
+  - task: NodeTool@0
+    displayName: Install Node 10.x
+    inputs:
+      versionSpec: 10.x
```
natemcmaster: Should this be 8.x? (per issues in templates that we won't fix)

test/Cli.FunctionalTests/run-tests.ps1
```diff
@@ -0,0 +1,162 @@
...
+param(
+    [switch]$ci,
+    $AssetRootUrl = $env:PB_ASSETROOTURL,
+    $AccessTokenSuffix = $env:PB_ACCESSTOKENSUFFIX,
+    $RestoreSources = $env:PB_RESTORESOURCES,
+    [ValidateSet('none', 'osx-x64', 'linux-x64', 'win-x64')]
+    $TestRuntimeIdentifier,
+    $HostRid,
+    $ProdConManifestUrl,
+    $ProcConChannel = 'release/2.2'
```
natemcmaster: "release/2.1" would be a better default here.

natemcmaster: Pro**d**ConChannel would be the right name to use here.

natemcmaster:
FYI @mikeharder I cherry-picked a fix from 2.2 onto this PR to fix one of the mismatched prodcon variables.

.vsts/builds/e2e-tests.yml
mikeharder: For now we can just test Node 10 on Windows.  I will create another PR to test both Node 8 and 10 on Windows.  Our 2.1 templates are only compatible with Node 8 on Mac.

PR #1418, opened 2018-09-24
Don't install 2.0 on linux-musl

ryanbrandenburg:
Fixes aspnet/Coherence-Signed#1057.


natemcmaster:
```xml
<!-- There is no linux-musl in 2.0. This was new in 2.1. -->
```

PR #1420, opened 2018-09-25
Revert dependencies.props to last known working set

natemcmaster:
Reverts part of https://github.com/aspnet/Universe/commit/48e66d3d101de3b34344d735340118061a5e6949. Something is busted in RAR and the latest builds of System.Runtime.CompilerServices.Unsafe

Resolves https://github.com/aspnet/Home/issues/3544

cc @dsplaisted @dougbu @ryanbrandenburg


natemcmaster:
I think there is a difference between saying you should not and you don't need to. The way I see it, tests were working fine before https://github.com/aspnet/Universe/commit/48e66d3d101de3b34344d735340118061a5e6949, so we should revert https://github.com/aspnet/Universe/commit/48e66d3d101de3b34344d735340118061a5e6949 to get back to a working state.

natemcmaster:
After more investigation, we think the problem was caused by merging release/2.2 to master incorrectly. This PR now puts build/dependencies.props back to the state it was in when build and test were working.

natemcmaster:
It seems to. This set of dependencies isn't perfect -- e.g. it's referencing stale verisons of System.Memory and System.ValueTuple, but it seems the latest corefx bits don't work if we try to mix stable RTM versions with nightly bits. So IMO, let's stick with this set until they fix their packaging and binding redirect generation.

PR #1450, opened 2018-10-03
React to aspnet/Logging#884

halter73:
Only take this if we take aspnet/Logging#884


Eilon:
What's the conclusion to https://github.com/aspnet/Logging/pull/884#issuecomment-426005845 ?

halter73:
If no one objects, I'll merge today.

Eilon:
Can you confirm this is all used in only non-shipping stuff? If so, go ahead.

BrennanConroy:
Looks like Serilog is only used by test projects in our org

halter73:
Thanks @BrennanConroy

PR #1452, opened 2018-10-04
Support using Internal.AspNetCore.Sdk as an MSBuild SDK

natemcmaster:
* Add Microsoft.AspNetCore.BuildTools.ApiCheck as an external dependency
* Move the global.json files temporarily to avoid loading the wrong version of Internal.AspNetCore.Sdk

Resolves https://github.com/aspnet/Coherence-Signed/issues/1117


natemcmaster:
Merging to unblock the build.

PR #1453, opened 2018-10-04
Run real-code signing on individual repo builds

natemcmaster:
This flows the code-signing setting into individual repo builds. Effectively, this will throw the switch from Coherence-Signed to individual repo signed builds.


natemcmaster:
:watch: this has downstream effects, like https://github.com/aspnet/Coherence-Signed/issues/1121. Let's resolve those first.

natemcmaster:
Downstream effects should be resolved. Let's try this now.

PR #1458, opened 2018-10-08
Update System.IO.Pipelines to 4.5.2

pakrym:
as per https://github.com/dotnet/corefx/issues/32163


pakrym:
CI failures seem unrelated. Fine to merge?

natemcmaster:
Yes, it appears unrelated. Looks like the failure is related to https://github.com/aspnet/BuildTools/issues/787. Unfortunately this means the normal checks for external packages are not actually executing.

PR #1459, opened 2018-10-08
Add Redis 2.0

BrennanConroy:


natemcmaster:
CI failure appears unrelated. Looks like the failure is related to https://github.com/aspnet/BuildTools/issues/787. Unfortunately this means the normal checks for external packages are not actually executing.

BrennanConroy:
@natemcmaster So should I wait or just merge?

PR #1463, opened 2018-10-09
Update Roslyn version.

NTaylorMullen:
aspnet/Razor#2634


NTaylorMullen:
Merging this because the error in the build seems completely unrelated and more like a vsts git hiccup

PR #1467, opened 2018-10-09
Rename Cosmos.Sql to Cosmos

ajcvickers:
After sub-module update.


ajcvickers:
@natemcmaster @ryanbrandenburg As far as I know the submodule is now updated, but this PR is still failing. I'm not able to tell from the failure details what is causing it.

ryanbrandenburg:
We're currently operating under the assumption that the problem failing this PR is https://github.com/Microsoft/msbuild/issues/2914 and might be fixed by https://github.com/aspnet/Universe/pull/1468. If that works we'll re-build and then get this merged in.

PR #1468, opened 2018-10-09
wip: Workaround issues in the NuGet SDK resolver

natemcmaster:
Attempt to resolve this error:

>  The SDK 'Internal.AspNetCore.Sdk' specified could not be found.

Marking as WIP because I can't reproduce locally, but I have a hunch.


natemcmaster:
(sad trombone sound). Doesn't fix it. Can't reproduce it locally. I'm assuming that means its an environment variable or networking issue.

natemcmaster:
I traced this back to a bug in MSBuild itself. The code that throws is attempting to do something out of the ordinary with MSBuild's API. This code isn't going to be necessary as we start merging repos, so I'm going to recommend we remove the buggy code altogether. I'll send a separate PR for that.

PR #1469, opened 2018-10-09
Remove tasks for analyzing the repo graph

natemcmaster:
As a part of converging repos, we no longer need this code. It is currently buggy and slow, and we can get away without by hard-coding the repo graph (which rarely changes)

Resolves https://github.com/aspnet/AspNetCore-Internal/issues/1135


build/repo.targets
```diff
@@ -226,12 +223,6 @@
   </Target>
 
   <Target Name="ComputeGraph" DependsOnTargets="ResolveRepoInfo;GeneratePropsFiles">
```
ryanbrandenburg: Technically this is more like "LoadSubmoduleGraph" now but it's probably not worth changing the name.

ryanbrandenburg:
How does it resolve that internal issue though? Did I miss one of the failures?

natemcmaster:
The error "The SDK 'Internal.AspNetCore.Sdk' specified could not be found." was being thrown from the ComputeRepoGraph task. Something about MSBuild's programmatic APIs is non-deterministic when resolving custom SDKs.

ryanbrandenburg:
Spooky. :shipit:

PR #1476, opened 2018-10-09
Use Moq v4.10

rynowak:


rynowak:
Earlier update had the wrong version. 4.10 is the right version

PR #1486, opened 2018-10-11
Prevent contention

ryanbrandenburg:


build/buildorder.props
```diff
@@ -1,6 +1,7 @@
 <Project>
   <ItemGroup>
-    <RepositoryBuildOrder Include="Common" Order="1" />
+    <!-- Common and Microsoft.Data.Sqlite can hypothetically be built together, but they tend to contend in installing tooling. -->
+    <RepositoryBuildOrder Include="Common" Order="0" />
```
Eilon: Is there a reason we don't pre-install tooling so that we can avoid this change? (Not that I'm particularly opposed to this change, but it's different from what we did with dotnet-install.)

natemcmaster: Pre-install should happen when universe runs its `build.sh` script. Normally, when Universe invokes `build.sh` on each repo, the pre-install noops because it has already happened. I believe this broke when we added `-ci` in the TeamCity config since `-ci` changes the default paths where tooling is installed. An alternative solution is to have Universe flow the `-ci` flag into sub-repo builds.

ryanbrandenburg:
:up::date:

PR #1487, opened 2018-10-11
Skip code-signing EFCore with repo signing while we workout issues with signtool

natemcmaster:
Code-signing will still happen. It will fallback to legacy signing which happens after Universe builds. This skips in-line signing for now until https://github.com/aspnet/BuildTools/pull/788 is resolved.


ryanbrandenburg:
The code looks fine, but I'm not sure I understand what is lost by not doing the signing here. Is it just that we're supposed to be deprecating legacy signing?

natemcmaster:
What is lost here is a step backwards in terms of getting Coherence-Signed decoupled from Universe (which is a pre-requisite to building in VSTS). It should be temporary though.

Issue #1496, opened 2018-10-12
Cleanup ANCM installers


jkotalik:
- [x] Move all props/targets files to a central location
- [x] Max path issues with nested folders
- [x] Add comments to explain how version variables should be used

jkotalik:
@natemcmaster feel free to add anything I missed.

natemcmaster:
LGTM, thanks.

natemcmaster:
Prior to https://github.com/aspnet/Universe/pull/1456, Version variables always ended in 0.
https://github.com/aspnet/Universe/blob/949c29c59f3ea0a9acb3cbbb0d23a830db99d814/src/Installers/Windows/Wix.props#L4

With https://github.com/aspnet/Universe/pull/1456, we changed this to include a time-based number. 
https://github.com/aspnet/Universe/blob/c37a332a6a220d36ce510569a3a6ab9f13906d17/src/Installers/Windows/Wix.props#L5-L8

@joeloff is this change the right thing to do?

Asking because I know there are funky rules in MSI around file versions, and want to make sure we're doing this right. This variable is used in 4 different ways. Is it okay to re-use for all of them?

1. On `<Product>` elements in .wxs https://github.com/aspnet/Universe/blob/da48de7d7580ce316784dd3e250786ffa7017c25/src/Installers/Windows/SharedFramework/Product.wxs#L3
2. On `<Bundle>` elements in .wxs https://github.com/aspnet/Universe/blob/da48de7d7580ce316784dd3e250786ffa7017c25/src/Installers/Windows/SharedFrameworkBundle/Bundle.wxs#L3
3. As inputs to GUIDs: https://github.com/aspnet/Universe/blob/949c29c59f3ea0a9acb3cbbb0d23a830db99d814/src/Installers/Windows/Wix.targets#L9-L10
4. And now used in ANCM in `<UpgradeVersion`> (`MsiVersion = "1$(Version)"`) https://github.com/aspnet/Universe/blob/c37a332a6a220d36ce510569a3a6ab9f13906d17/src/Installers/Windows/AspNetCoreModule-Setup/Installers/ANCMIISExpressV1/ancm_iis_express.wxs#L75-L78

joeloff:
[...]

PR #1499, opened 2018-10-15
Cleanup ANCM installers

jkotalik:
For https://github.com/aspnet/Universe/issues/1496

@natemcmaster I still need to investigate signing issues with the aspnetcoreca.dll.


src/Installers/Windows/tasks/GenerateGuid.cs
```diff
@@ -13,19 +13,40 @@ public class GenerateGuid : Task
         [Output]
         public string Guid { get; private set; }
 
-        [Required]
         public string NamespaceGuid { get; set; }
 
-        [Required]
         public ITaskItem[] Values { get; set; }
 
+        // public override bool Execute()
```
joeloff: Is this going to be deleted?

jkotalik: Whoops.

src/Installers/Windows/tasks/GenerateGuid.cs
```diff
@@ -13,19 +13,25 @@ public class GenerateGuid : Task
...
+                {
+                    var value = string.Join(",", Values.Select(o => o.ItemSpec).ToArray()).ToLowerInvariant();
 
-                Guid = Uuid.Create(new Guid(NamespaceGuid), value).ToString();
+                    Guid = Uuid.Create(new Guid(NamespaceGuid), value).ToString();
+                }
+                else
+                {
+                    // Providing a guid of * as the product code 
+                    Guid = "*";
```
natemcmaster: I'm concerned this may cause us to unintentionally break ourselves. I'd rather have a property in the .wixproj file which specifies whether not guids should be generated. So for example, ANCM .wixproj files could set a property `<DisableGuidGeneration>true</DisableGuidGeneration>`. Then, update this target to only run on `Condition=" '$(DisableGuidGeneration)' != 'true' "`
 
https://github.com/aspnet/Universe/blob/da48de7d7580ce316784dd3e250786ffa7017c25/src/Installers/Windows/Wix.targets#L15

jkotalik: Disagree with `DisableGuidGeneration`. We are still generating a guid, we just it to be random. Maybe we can add a bool to the GenerateGuid Task called `GenerateRandomGuid`?

natemcmaster: What's the point of invoking an MSBuild task to set the GUID to `*`?

jkotalik:
@natemcmaster  🆙 📅

